PR #11, opened 2016-02-28
replace sequential fread() calls in Reader by a single read per row

scoder:
changes SparseArrayC to a packed struct of platform independent size to match exactly the file format (minus byte order)
speeds up loading by about 2x

### DIFF
--- a/thinc/linear/serialize.pyx
+++ b/thinc/linear/serialize.pyx
@@ -72,7 +72,7 @@ cdef class Reader:
         self._fp = fopen(<char*>bytes_loc, 'rb')
         if not self._fp:
             PyErr_SetFromErrno(IOError)
-        tatus = fseek(self._fp, 0, 0)
+        status = fseek(self._fp, 0, 0)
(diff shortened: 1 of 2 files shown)


scoder:
I ran the benchmark in https://github.com/spacy-io/thinc/commit/c0b6c74ca417ea985d688e9392d92f1a387d70ff#commitcomment-16371306 and it shows no difference for the text parsing. Only the initialisation overhead is lower. So this seems a safe change, at least on Linux 64bits.

honnibal:
Great! I guess we could also change the file format, so that we read the data into three temporary buffers (keys, lengths and the row contents). You could try that if you think it'll help a lot?

Otherwise, if you still have time, you might have a look at the speed bottle-neck of spaCy, the sparse dot product here: https://github.com/spacy-io/thinc/blob/master/thinc/linear/avgtron.pyx#L115

The comment says the bottle-neck might be fetching the weights from memory, which I guess I must have meant fetching from the hash table. I'm not at all sure about this though.

PR #14, opened 2016-04-16
new import-include proposal

henningpeters:
Please check new idea for getting rid of shipping C headers of dependencies. As far as I can tell with this technique the only drawback is that on first install from a source distribution pip's default wheel building/caching process causes a non-fatal error. This seems better than potentially shipping incompatible headers.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,3 @@
-recursive-include include *.h
 include buildbot.json
 include LICENSE
 include README.rst
--- a/setup.py
+++ b/setup.py
@@ -10,8 +10,11 @@ from distutils.sysconfig import get_python_inc
 
 try:
     from setuptools import Extension, setup
+    from pkg_resources import resource_filename
 except ImportError:
     from distutils.core import Extension, setup
+    def resource_filename(package, _):
+        return import_include(package).get_include()
 
 
 PACKAGES = [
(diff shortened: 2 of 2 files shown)


syllog1sm:
Okay so, the idea is that the setup.py gets executed in one of four ways:

1. Compiling cython code to build the egg. Here we can assume that all dependencies are present at compile time.
2. Installing from egg, dependencies are present (i.e. murmurhash, numpy are there). Here we import the  module, and ask it to give us a path to its copy of the headers.
3. Installing from egg, dependencies aren't present. Here we fall back to vendorised headers.
4. Installing from binary wheel. Here the headers aren't relevant, right?

Not sure this is all correct, but thought going through these cases seemed like a good way through.

henningpeters:
Good idea to list all possible cases. I have following scenarios is mind:

1. Developer build: `pip install -r requirements.txt && pip install -e .`.

2. Packaging build: `pip install -r requirements.txt && python setup.py sdist bdist_wheel`

3. Install from wheel, either manually prepared (see 2) or using the official distribution or via pip's auto-cache mechanism that pip creates after installing from source.

4. Install from official source distribution (dependencies present)

5. Install from official source distribution (dependencies not present)

Eggs should behave the same as wheels (can be built via `pip install -r requirements.txt && python setup.py sdist bdist_egg`). We don't provide an official distribution though and I didn't test my proposal properly with it, yet.

All those cases should be covered. The tricky case is the 5th, but except a potentially confusing error message it also works.

syllog1sm:
What's the error message?

henningpeters:
```
  #include "murmurhash/MurmurHash3.h"
           ^
  1 error generated.
  error: command 'cc' failed with exit status 1
  
  ----------------------------------------
  Failed building wheel for thinc
Failed to build thinc
Installing collected packages: murmurhash
Successfully installed murmurhash-0.26.3
```

syllog1sm:
This seems bad. What's the benefit of the change again? Just the reduced risk of vendorising the wrong headers?

henningpeters:
It's conceptionally cleaner (less moving parts) and can prevent problems, e.g., when numpy makes an incompatible change. Hence, I expect less trouble with it in the future. By vendorizing the headers you blindly trust that the headers we ship are compatible with every numpy version out there.

If we want to make the error message go away/support wheel cache we could retain a fallback with vendorized headers.

syllog1sm:
I see the appeal, but 5) is pretty much the basic case, and if I got that error I'd stop what I was doing and assume I had to fix it. This will happen on many spaCy installations.

henningpeters:
Agree, I expect that we'll see the default case becoming 3 in the near future though: once we have wheels on Linux. Just keep the PR around until then...

PR #18, opened 2017-01-17
Update examples/spacy_tagger.py

kootenpv:
But still does not work, it crashes with:

```python
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/thinc/extra/_vendorized/keras_generic_utils.py", line 114, in update
    numdigits = int(np.floor(np.log10(self.target))) + 1
ValueError: cannot convert float NaN to integer
```


honnibal:
Hey,

This is all under very active development at the moment, so there will continue to be these rough spots. Thanks for the fixes. I hope to have the code updated for this soon, to get the example working again.

kootenpv:
I'm very curious to the examples :) Naturally I'd like to get a taste of the extra value of using thinc over spacy...

PR #19, opened 2017-02-08
Fix off-by-one error in sparse array clone method

almostimplemented:
Currently the result from clone won't have the `-2` terminal key.


almostimplemented:
Fixed in e4ea40bdfcb796be6f54fa36b2c8571190ae48dc .

PR #24, opened 2017-03-20
Fix integer check on array shapes.

rolando:
This mainly address Python 2.7's Windows interpreter where `numpy`'s
`shape` have `long` values and hence `isinstance(shape[i], int)`
returns `False`.

--

Related: https://github.com/numpy/numpy/issues/5809

### DIFF
--- a/thinc/check.py
+++ b/thinc/check.py
@@ -3,6 +3,7 @@ import inspect
 import wrapt
 from cytoolz import curry
 from numpy import ndarray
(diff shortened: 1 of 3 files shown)


honnibal:
Interesting -- I didn't know that, thanks.

rolando:
@honnibal yeah, kinda backwards. Good thing that `long` is gone in python 3.x.

PR #25, opened 2017-05-03
Fix bug: gradient array in feed-forward model is not C-Contiguous

bakarov:
I tried to execute the script `thinc/examples/text-pair/glove_mwe_multipool_predict.py`, and I have encountered with an error:

```
/usr/lib/python3.6/site-packages/thinc/neural/_classes/maxout.py in finish_update(dX__bo, sgd)
     54 
     55         def finish_update(dX__bo, sgd=None):
---> 56             dX__bop = self.ops.backprop_maxout(dX__bo, which__bo, self.nP)
     57             self.d_b += dX__bop.sum(axis=0)

/usr/lib/python3.6/site-packages/thinc/neural/ops.pyx in thinc.neural.ops.NumpyOps.backprop_maxout (thinc/neural/ops.cpp:12005)()

/usr/lib/python3.6/site-packages/thinc/neural/ops.cpython-36m-x86_64-linux-gnu.so in View.MemoryView.memoryview_cwrapper (thinc/neural/ops.cpp:30499)()

/usr/lib/python3.6/site-packages/thinc/neural/ops.cpython-36m-x86_64-linux-gnu.so in View.MemoryView.memoryview.__cinit__ (thinc/neural/ops.cpp:26751)()

ValueError: ndarray is not C-contiguous
```
I have fixed this error by making the array given to `finish_update` C-contiguous in the callback function, and it worked for me. Perhaps some folks may also face this problem.

### DIFF
--- a/thinc/neural/_classes/feed_forward.py
+++ b/thinc/neural/_classes/feed_forward.py
@@ -1,6 +1,6 @@
 from .model import Model
 from ... import describe
-
+from numpy import ascontiguousarray
 
 def _run_child_hooks(model, X, y):
     for layer in model._layers:
@@ -39,6 +39,6 @@ class FeedForward(Model):
             for callback in reversed(callbacks):
                 if gradient is None or callback == None:
                     break
-                gradient = callback(gradient, sgd)
+                gradient = callback(ascontiguousarray(gradient), sgd)
             return gradient
         return X, continue_update


honnibal:
Hey,

This stuff is all pretty messy at the moment -- there's some updated state in `develop`, that will allow the similar script `glove_mwe_multipool_siamese` on the GPU. Things may break unexpectedly.

In general to fix stuff like the error you experienced, we want to be fixing the data when it's output from the layer --- so we wouldn't want to change the hook like this. Also, we want to be using the `self.ops.xp` to access numpy or cupy, so that the code works the same on both CPU and GPU.

Thanks!
Matt

honnibal:
Closing this -- it should be fixed now.

msamogh:
@bakarov How did you manage to recompile thinc and not break anything? I got a `ValueError: thinc.extra.search.MaxViolation has the wrong size, try recompiling` while trying to do that.

Also, why doesn't adding `ascontiguousarray` in my application code fix it?

PR #29, opened 2017-06-06
Make the README example executable

tammoippen:
- The function `backprop_relu` needs the `optimizer` parameter, otherwise the `chain` function does not work.
- Parameter `b` needs to be a 'matrix', so that you can do `Y = W @ X + b`
- The `outer` function (considering it is the numpy version) only works, if `X` and `dY` consist of one training example. With more, i.e. batch learning, `einsum` does column-wise outer product and the sum over these matrices.
- The error for the next step is `dX` in the `backward` function from `create_linear_layer`.
- Finally, some dimensions are wrong and optimizer parameters are fixed

Now you can copy-paste the code examples into a notebook, add a optimizer function, e.g. the `sgd` from `explicit_sgd_update` and run the example.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -80,7 +80,7 @@ To make this less abstract, here's a ReLu activation, following this signature:
 
     def relu(inputs):
         mask = inputs > 0
-        def backprop_relu(d_outputs):
+        def backprop_relu(d_outputs, optimizer):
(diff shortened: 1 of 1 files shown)


honnibal:
Thanks!

PR #40, opened 2017-10-28
Update setup.py

RaananHadar:
'-arch=sm_20' is no longer a supported architecture in cuda 9.0


honnibal:
Okay but does this break CUDA 8.0?

RaananHadar:
You are correct. Now that i've seen your code, i've created a proper pull request that is less of a hack :) Also corrected a small bug on the way.

PR #41, opened 2017-10-28
Update setup.py to work with CUDA 9.0

RaananHadar:
Setup.py didn't run on cuda 9, fixed the proper compile options. This fix should also work on CUDA 8


honnibal:
Great, thanks!

PR #57, opened 2018-03-20
Ship our own copy of select openblas functions, to remove system dependency

honnibal:
The previous solution of having an install-time switch to control the BLAS linkge didn't work, as the switch needs to be flipped at *cythonize* time. Instead we switch to a higher effort but better solution: build the bits of OpenBLAS we need with distutils, and incorporate it into the library.

This should allow us to remove dependencies on the system state, and avoid conflicting with the user's other openblas settings that control threads etc. We also get nogil, no overhead calls.


honnibal:
This is still pretty flakey, but current state of master is basically broken and this is definitely the future, so merging.

PR #61, opened 2018-03-27
Wip/wrap pytorch rnns

dvsrepo:
This a very rough attempt and WIP for adding support for RNNs. 

@honnibal, if you can take a look and comment I can evolve this into a proper PR :-)

It would be nice to add some tests as well, output shapes of RNNs vary greatly with directions and num_layers. 

PS: It also adds some minor fixes

### DIFF
--- a/thinc/extra/wrappers.py
+++ b/thinc/extra/wrappers.py
@@ -20,7 +20,7 @@ class PyTorchWrapper(Model):
         self._model = model
         self._optimizer = None
 
-    def begin_update(self, x_data, drop=0., sgd=None):
+    def begin_update(self, x_data, drop=0.):
         '''Return the output of the wrapped PyTorch model for the given input,
         along with a callback to handle the backward pass.
         '''
@@ -97,19 +97,23 @@ class PyTorchWrapper(Model):
 class PyTorchWrapperRNN(PyTorchWrapper):
     '''Wrap a PyTorch RNN model
     '''
-    def begin_update(self, x_data, h_0=None, drop=0., sgd=None):
+    def begin_update(self, x_data, h_0=None, drop=0.):
         '''Return the output of the wrapped PyTorch model for the given input,
         along with a callback to handle the backward pass.
         '''
         x_var = torch.autograd.Variable(torch.Tensor(x_data),
                                         requires_grad=True)
+        
+        
+        
         # Make prediction
         out, h_n = self._model(x_var, h_0)
         # Shapes will be:
         # out = seq_len, batch, hidden_size * num_directions
         # h_n = num_layers * num_directions, batch, hidden_size
 
-        def backward_pytorch(dy_data, sgd=None):
+        def backward_pytorch_rnn(d_data, sgd=None):
+            dy_data, _ = d_data
             dout = torch.autograd.Variable(torch.Tensor(dy_data))
             torch.autograd.backward((out,), grad_variables=(dout,))
             dX = self.ops.asarray(x_var.grad.data)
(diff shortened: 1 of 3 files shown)


thinc/extra/wrappers.py
```diff
@@ -20,13 +20,14 @@ def __init__(self, model):
         self._model = model
         self._optimizer = None
 
-    def begin_update(self, x_data, drop=0.):
+    def begin_update(self, x_data, drop=0., sgd=None):
```
honnibal: Ah, this isn't quite correct: we pass the `sgd` callable in during the backward pass --- `begin_update` is the forward pass.

thinc/extra/wrappers.py
```diff
@@ -93,3 +94,48 @@ def use_params(self, params): # pragma: no cover
...
+            dout = torch.autograd.Variable(torch.Tensor(dy_data))
+            torch.autograd.backward((out,), grad_variables=(dout,))
+            dX = self.ops.asarray(x_var.grad.data)
+            if sgd is not None:
+                if self._optimizer is None:
+                    self._optimizer = self._create_optimizer(sgd)
+                self._optimizer.step()
+                self._optimizer.zero_grad()
+            return dX
+        # TODO: We should return also h_n e.g. for seq2seq models
```
honnibal: Sure, it's probably a good thing to return a tuple. But then we should also change the gradient argument to `backward_pytorch_rnn` as well: the output from the forward pass and the argument to the backward pass should always line up (as should the argument to the forward pass and the output of the backward pass).

thinc/extra/wrappers.py
```diff
@@ -93,3 +94,48 @@ def use_params(self, params): # pragma: no cover
...
+        '''
+        x_var = torch.autograd.Variable(torch.Tensor(x_data),
+                                        requires_grad=True)
+        # Make prediction
+        out, h_n = self._model(x_var, h_0)
+        # Shapes will be:
+        # out = seq_len, batch, hidden_size * num_directions
+        # h_n = num_layers * num_directions, batch, hidden_size
+
+        def backward_pytorch(dy_data, sgd=None):
```
honnibal: I've learned from painful experience that these closures need specific names, or the tracebacks get pretty ugly. So we want this to be called `backward_pytorch_rnn`

thinc/extra/wrappers.py
dvsrepo: Sorry! Indeed! It is the call to the method in examples/wrap_pytorch. py which was wrong then. I will fix this.

thinc/extra/wrappers.py
honnibal: Returning a tuple isn't so inconvenient, because the model can be wrapped in the `with_getitem` adapter from `thinc.api`. For instance, if you want the first element of the tuple (and want to send back a gradient of that element), you'd do `with_getitem(0, PytorchWrapperRNN())`

dvsrepo:
Hi @honnibal, I have added some changes trying to reflect your comments. Now fwd returns a tuple and backward has a tuple as argument. Currently, we just ignore the second argument (h_n), I am sure you will have some ideas for refining this. 

I have also included some silly tests in the ``examples/wrap_pytorch_rnn.py`` just to verify the current state of affairs, based on your existing test for pytorch_wrapper (learning a zeroed output).

I hope I didn't screw up too much :-)

thinc/extra/wrappers.py
```diff
@@ -119,8 +123,7 @@ def backward_pytorch(dy_data, sgd=None):
                 self._optimizer.step()
                 self._optimizer.zero_grad()
             return dX
-        # TODO: We should return also h_n e.g. for seq2seq models
-        return self.ops.asarray(out.data), backward_pytorch
+        return (self.ops.asarray(out.data), h_n) , backward_pytorch_rnn
```
dvsrepo: Some pending issues: 
- h_n should not be PyTorch specific when send outside this function
- For LSTMs h_n will contain a tuple with h_n and c_n (the cell state for t=seq_len)

honnibal:
@dvsrepo Sorry about the delay getting this reviewed. Looks good! Going to merge so I can more easily test it :)

PR #66, opened 2018-06-23
Improve link command generation for THINC_BLAS

nehaljwani:
This patch allows one to specify BLAS library on macOS and Windows too
```
export THINC_BLAS=/path/to/libmkl_rt.dylib
```
```
set THINC_BLAS=\path\to\mkl_rt.lib
```


honnibal:
Thanks, this looks good!

PR #77, opened 2018-10-23
Depend on cython-blis, instead of shipping OpenBLAS kernel

honnibal:
This patch replaces the vendorized OpenBLAS kernel previously shipped in v6.11 with a dependency on our Blis package: https://github.com/explosion/cython-blis

Multiplying matrices efficiently continues to be a massive struggle :(. The initial strategy of doing this via numpy causes thread safety problems on OSX, and can cause poor performance. I therefore tried to replace the numpy dependency with our own OpenBLAS kernel. However, this made building on Windows too difficult, and prevented us from shipping wheels, as I was relying on detecting the architecture during the setup.py.

This patch instead delegates the matrix multiplications to the Blis linear algebra library, provided by our new `blis` package. We now have wheels up for OSX and Windows for 64-bit versions of Python 3. I hope to have wheels for 32-bit Python 3 up soon. Windows Python 2.7 would be nice, but seems difficult. Linux wheels are unfortunately quite tough: I don't think it can be done with manylinux1, as we need gcc 6.


bradjonesca:
Just an FYI for an alternative to Travis [https://github.com/fastify/fastify/pull/1226#issuecomment-432325195](url)

PR #96, opened 2019-03-25
Comment out assert statement for empty length X in unflatten()

mwakaba2:
This fix is related to the spacy issue [3456](https://github.com/explosion/spaCy/issues/3456)
The unflatten function fails with a two-dimensional list that contains empty lists.

I added a test function `test_nested_unflatten_length_zero` to test this specific example.

### DIFF
--- a/thinc/neural/ops.pyx
+++ b/thinc/neural/ops.pyx
@@ -135,7 +135,7 @@ class Ops(object):
             X = X[length:]
         if pad >= 1 and length != 0:
             X = X[pad:]
-        assert len(X) == 0
+        # assert len(X) == 0 can cause error if 0 exists in lengths
         assert len(unflat) == len(lengths)
         return unflat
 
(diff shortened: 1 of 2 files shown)


honnibal:
Thanks!

I'm still a little confused about where my assumptions in that function went wrong. If `X` has trailing data here, isn't that bad? How come we don't need those rows?

mwakaba2:
Yes I’m also confused to the trailing data for empty inputs. I’m not familiar with how the thinc tagger outputs for empty strings, so I’m hoping someone whose more familiar with the codebase can propose a better solution. 
Commenting out an assertion isn’t an ideal fix. 

FYI, I only tested if the nlp.pipe doesn’t break when I feed in an empty phrase.  I probably want to double check what the tagger outputs look like for those.

honnibal:
Now fixed by #104 :tada: . Thanks for your help on this.

PR #102, opened 2019-06-22
Pass seed to init

chssch:
- Bugfix: Recognize passed seed parameter
- Basis to fix https://github.com/explosion/spaCy/issues/3870

### DIFF
--- a/thinc/neural/_classes/hash_embed.py
+++ b/thinc/neural/_classes/hash_embed.py
@@ -39,7 +39,11 @@ class HashEmbed(Model):
         self.column = kwargs.get("column", 0)
         self.nO = nO
         self.nV = nV
-        self.seed = self.id
+        
+        if seed is not None:
+            self.seed = seed
+        else:
+            self.seed = self.id
 
     def predict(self, ids):
         if ids.ndim >= 2:


chssch:
Should the default actually be the id? Is there any reason that it's not a constant? This would also fix https://github.com/explosion/spaCy/issues/3870

honnibal:
@chssch If we made it constant, then we might get confusing results if you create two hash embeddings and add them, as you'd expect them to distribute differently.

Does your patch alter the trained models? I think not, right -- the trained models load the seed from the cfg file?

chssch:
> @chssch If we made it constant, then we might get confusing results if you create two hash embeddings and add them, as you'd expect them to distribute differently.

Okay, thanks, got it.

> Does your patch alter the trained models? I think not, right -- the trained models load the seed from the cfg file?

No, it shouldn't alter the trained modes. It's just cosmetic preparation to fix https://github.com/explosion/spaCy/issues/3870, but no blocker, since we can also set the seed directly on the object. 

I guess the main problem is, how can we generally inject a constant seed from top level before a training run to get reproducible results. (but thats maybe for  the discussion in https://github.com/explosion/spaCy/issues/3870. )

Issue #108, opened 2019-08-09
DeprecationWarnings for collections ABCs in python3.7


hervenicol:
My project's tests with pytest are issuing the following warnings:
```
/usr/local/lib/python3.7/site-packages/thinc/neural/train.py:7
  /usr/local/lib/python3.7/site-packages/thinc/neural/train.py:7: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from .optimizers import Adam, linear_decay

/usr/local/lib/python3.7/site-packages/thinc/check.py:4
/usr/local/lib/python3.7/site-packages/thinc/check.py:4
/usr/local/lib/python3.7/site-packages/thinc/check.py:4
  /usr/local/lib/python3.7/site-packages/thinc/check.py:4: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Sequence, Sized, Iterable, Callable
```

According to Python's doccumentation, in Python version 3.3, Collections Abstract Base Classes have been moved to the collections.abc module.
For backwards compatibility, they continue to be visible in this module through Python 3.7. But they will be removed entirely in Python 3.8.
(source: https://docs.python.org/3/library/collections.html)

PR #109, opened 2019-08-09
Fix Collections Abstract Base Classes imports for Python 3.8

hervenicol:
Since Python version 3.3, Collections Abstract Base Classes have been moved to the collections.abc module. For backwards compatibility, they continue to be visible in this module through Python 3.7. Subsequently, they will be removed entirely.

(source: https://docs.python.org/3/library/collections.html)

Related to issue #108

### DIFF
--- a/thinc/check.py
+++ b/thinc/check.py
@@ -1,4 +1,9 @@
-from collections import defaultdict, Sequence, Sized, Iterable, Callable
+try:
+    # Python >= 3.3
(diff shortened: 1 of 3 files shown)


honnibal:
Thanks!

hervenicol:
@honnibal , it seems that master branch is more up-to-date than develop.
Maybe I should rather base my work on master ?

honnibal:
@hervenicol Sorry I missed that about this as well: yes, work off master.

PR #115, opened 2019-10-15
Update ufuncobject.h

KoichiYasuoka:
Recent cygwin-devel-3.0.6 (64bit) has `fenv.h` instead of `fenv/fenv.c`. (cf. https://github.com/explosion/spaCy/issues/2411 )


honnibal:
Thanks, I'm pretty sure this just needs to be updated from master --- the test failures look unrelated.

KoichiYasuoka:
Thank you, @honnibal. I'm looking forward to [pypi](https://pypi.org/project/thinc/) update.

PR #117, opened 2019-10-18
Ditch thinc_gpu_ops for simpler GPU install

### LINKED ISSUE #51: Thinc GPU capability on Windows - fails to compile
It looks like Thinc has some hard time making GPU computation available if used on Windows.

Similar issues were raised in spaCy and thinc:
https://github.com/explosion/spaCy/issues/1745
https://github.com/explosion/thinc/issues/50

It seems related in how the build is set-up in setup.py:
https://github.com/explosion/thinc/blob/master/setup.py#L56

Based on quick search, others are struggling with this as well:
https://github.com/rbgirshick/py-faster-rcnn/issues/11

Some more info (they came to a working solution there):
https://github.com/cudamat/cudamat/issues/46#issuecomment-73803650

Will research a bit more later.

honnibal:
spaCy and Thinc have had one last pesky non-wheeled module: `thinc_gpu_ops`, which needed to be compiled with a GPU card to get the GPU support working. This was bad news on Windows, as it forced Windows users to set up a compiler toolchain.

This patch replaces the `thinc_gpu_ops`, instead bringing the kernels into Thinc and using cupy's `RawKernel` to compile them at runtime.

Many Windows users had previously found it difficult to use the GPU support. This should fix the problem.

I've also improved the kernels somewhat, although they weren't a bottleneck, so I doubt it'll make much performance difference. If we do hit a new bug in the GPU code though, that will be why.

I've preferred to keep separate `.cu` source files, instead of embedding the strings into the Python source. Having code as strings feels pretty nasty. Instead I just read in the source file, and segment it with some regexes.

Closes #51, explosion/spaCy#4088


svlandeg:
Windows users thank you 😇

PR #122, opened 2019-11-19
Fix cupy to cuda 10.1 version

mmaybeno:
I think this is a typo but wanted to make sure. `cupy` has a different version of CUDA 10.1 and therefore unable to run `thinc[cuda101]`.


honnibal:
Thanks! Could've sworn I merged this fix before...

PR #124, opened 2019-11-20
Make Model operators and device ops thread-local

adrianeboyd:
Make old/saved Model operators thread-local, too. Doesn't seem like a particularly great solution, but seems to least temporarily fix https://github.com/explosion/spaCy/issues/4349. I think better solutions would require a fairly major restructuring of `Model`.

### DIFF
--- a/thinc/check.py
+++ b/thinc/check.py
@@ -130,8 +130,8 @@ def operator_is_defined(op):
             instance = args[0]
         if instance is None:
             raise ExpectedTypeError(instance, ["Model"])
-        if op not in instance._operators:
-            raise UndefinedOperatorError(op, instance, args[0], instance._operators)
+        if op not in instance.THREAD_LOCAL.operators:
+            raise UndefinedOperatorError(op, instance, args[0], instance.THREAD_LOCAL.operators)
         else:
             return wrapped(*args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


adrianeboyd:
I suspect `use_device()` has similar problems.

adrianeboyd:
If I try the same test outside pytest it does throw an `UndefinedOperatorError`:

```
from thinc.neural._classes.model import Model

op = "+"
m1 = Model(name="a")
m2 = Model(name="b")
with Model.define_operators({op: lambda a, b: a.name + b.name}):
    if op == "+":
        value = m1 + m2
    else:
        value = m1 + m2
    if op == "-":
        value = m1 - m2
    else:
        value = m1 - m2 # raises UndefinedOperatorError
```

Hmm, it looks like the operator `*` is hanging around from `test_overload_operators_in_subthread()`, which seems weird, although potentially related, of course. I'm not really sure what's going on?

adrianeboyd:
I initially thought this setup was doomed to fail with `_operators` as a class variable, but I think I can get it to work by reworking the thread-local definitions a bit. It's still kind of strange overall.

adrianeboyd:
With CupyOps this test fails: `test_backprop_sum_pool()`, but that seems unrelated to these changes.

adrianeboyd:
Is `with Model.use_device()` ever nested like `with Model.define_operators()`?

honnibal:
@adrianeboyd No we don't nest `with Mode.use_device()`. I think that design should probably be revised anyway.

Agree it's a bit ugly, but I think it's the best we can do.

PR #125, opened 2019-11-27
CuPy package for CUDA 10.1 is `cupy-cuda101`

kmaehashi:
Currently `cupy-cuda110` does not exist.
https://docs-cupy.chainer.org/en/latest/install.html#install-cupy
https://pypi.org/project/cupy-cuda101/


honnibal:
Oops, thanks!

PR #126, opened 2019-12-07
Add Config class

honnibal:
With the move to use [catalogue](https://github.com/explosion/catalogue) for extended configuration in spaCy, we need a solution for configuration management. It makes sense for the solution to live in Thinc, because Thinc's optimizers, layers etc should all work with the config management.

I looked closely at all of the options for config management. The config files will get pretty big, so we shouldn't just use JSON. YAML is terrible. XML would work but it'll involve a fair bit of code. TOML and HCL don't have first-part support in Python, and they're a bit less common.

I'm proposing a solution based on Python's configparser module, with a couple of small conventions. 

Configparser's format is an ini-like format that maps into a two-level dict:

```
[outer key 1]
inner key 1a = 1
inner key 1b = 2

[outer key 2]
inner key 2a = 3
inner key 2b = 4
```

Produces:

```
{
  "outer key 1": {
    "inner key 1a": "1",
    "inner key 1b": "2"
  },
  "outer key 2": {
    "outer key 2a": "3",
    "outer key 2b": "4"
  }
}
```

Our configs need nested values. I propose we do these via convention, transforming sections like `parser.model` so that they become `config["parser"]["model"]`. This actually works out pretty nicely: the file is just a long list of leaf-data, with the structure handled in the titles. It's also pretty easy to read and write the data this way.

I've also added some simple value interpretation logic, so that floats, ints and booleans come back as the correct values. The configparser supports unquoted strings in the values, which I think is pretty sloppy...I think we probably want to raise on those. That would also allow us to add some extra syntax for paths.

The other bit of convention I've introduced is using `@` to denote that a section refers to a catalogue entry. The configparser format supports both `:` and `=` as delimiters. I'm proposing using `:` only for the `@` key.

Finally, configparser also supports variable interpolation. You can refer to a different value in the config by writing `${other_section:other_key}`. If you're referring to a value within the same section, you don't need to write the section name.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -1,0 +1,115 @@
+# Configuration files
+
+You can describe Thinc models and experiments using a configuration format
+based on Python's built-in `configparser` module. We add a few additional
+conventions to the built-in format, to provide support for non-string values,
+support nested objects, and to integrate with Thinc's *registry system*.
+
+## Example config
+
+
+```
+[some_section]
+key1 = "string value"
+another_key = 1.0
+# Comments, naturally
+third_key = ["values", "are parsed with", "json.loads()"]
+some_other_key =  
+    {
+        "multiline values?": true
+    }
+
+[another_section]
+more_values = "yes!"
+null_values = null
+interpolation = ${some_section:third_key}
+
+# Describe nested sections with a dot notation in the section names.
+[some_section.subsection]
+# This will be moved, producing:
+#   config["some_section"]["subsection"] = {"hi": true, "bye": false}
+hi = true
+bye = false
+```
+
+The config format has two has two main differences from the built-in `configparser`
+module's behaviour:
+
+* JSON-formatted values. Thinc passes all values through `json.loads()` to
+  interpret them. You can use atomic values like strings, floats, integers,
+  or booleans, or you can use complex objects such as lists or maps.
+
+* Structured sections. Thinc uses a dot notation to build nested sections. If
+  you have a section named `[outer_section.subsection]`, Thinc will parse that
+  into a nested structure, placing `subsection` within `outer_section`
+
+## Registry integration
+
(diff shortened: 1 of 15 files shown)


svlandeg:
This looks pretty neat!

> You can refer to a different value in the config by writing `${other_section:other_key}`. 

I don't suppose you could also do basic numeric operations like  
`width = ${other_section1:width1} + ${other_section2:width2}` ?

If not - wouldn't there be cases where we need that kind of functionality?

honnibal:
> I don't suppose you could also do basic numeric operations like
> `width = ${other_section1:width1} + ${other_section2:width2}` ?
> 
> If not - wouldn't there be cases where we need that kind of functionality?

Yeah I've been thinking about that. I do think we'll want something like that. Not sure what the best way to go about it is.

docs/config.md
```diff
@@ -0,0 +1,115 @@
...
+
+# Describe nested sections with a dot notation in the section names.
+[some_section.subsection]
+# This will be moved, producing:
+#   config["some_section"]["subsection"] = {"hi": true, "bye": false}
+hi = true
+bye = false
+```
+
+The config format has two has two main differences from the built-in `configparser`
```
svlandeg: typo "has two has"

docs/config.md
```diff
@@ -0,0 +1,115 @@
...
+        "multiline values?": true
+    }
+
+[another_section]
+more_values = "yes!"
+null_values = null
+interpolation = ${some_section:third_key}
+
+# Describe nested sections with a dot notation in the section names.
+[some_section.subsection]
```
svlandeg: Nitpicking: do we want a convention where subsections are in the config in order, right below their respective supersection? Because in this example case we first have `some_section`, then `another_section` and then a subsection of `some_section`. We could rewrite the example to be `[another_section.subsection]` so that the order seems more natural ?

PR #153, opened 2020-01-03
Small type fixes

svlandeg:
- adding type for `operators` in `Model.define_operators`
- I think `Config` should get as `data` a `dict` object or a `ConfigParser` ? Or is there also a use-case where a `Config` parser itself is given as argument? (and would this not be caught already because `Config` inherits from `dict` ?) Either way a change is needed because `Config.from_str` calls `interpret_config` with a `ConfigParser`.

### DIFF
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -12,14 +12,14 @@ class Config(dict):
     under the hood.
     """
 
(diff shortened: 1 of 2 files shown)


ines:
Thanks! The Config/Configparser thing was definitely a mistake. (I vaguely remember changing this because I thought it was a typo, but... nope 😅)

PR #154, opened 2020-01-03
Small refactor of config and registry parsing methods

ines:
- [x] make registry methods return `Config` object instead of dict (so you can directly call `to_disk` etc. on the result, which makes sense, since you also typically pass in a config, and the config otherwise behaves like a dict)
- [x] only use one `_fill` helper for recursive validation and filling, and add two separate classmethods to expose the result: `make_from_config` (returns filled config with resolved functions from registry) and `fill_config` (only fills in default values from schema and/or function signature and returns updated config)
- [x] allow toggling validation for both modes
- [x] improve handling of unset required arguments and how errors are raised for it (also skip it if validation is disabled instead of raising the error)
- [x] skip `[DEFAULT]` default section in interpreted configparser result for now – not sure if we want to use this at all, since it introduces another layer of defaults? If users want it, they can add any custom config section and use that for interpolation? The `DEFAULT` is kinda unintuitive and messes up the validation, because all top-level schemas would now require it.
- [x] add more tests and docstrings

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -110,3 +110,4 @@ exclude_lines =
 
 [mypy]
 ignore_missing_imports = True
+plugins = pydantic.mypy
--- a/thinc/_registry.py
+++ b/thinc/_registry.py
@@ -1,4 +1,4 @@
-from typing import Callable, Dict, Any, Tuple, List, Optional, Union
+from typing import Callable, Dict, Any, Tuple, List, Union, Type
 from types import GeneratorType
 import catalogue
 import inspect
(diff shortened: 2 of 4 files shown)


honnibal:
> skip [DEFAULT] default section in interpreted configparser result for now – not sure if we want to use this at all, since it introduces another layer of defaults? If users want it, they can add any custom config section and use that for interpolation? The DEFAULT is kinda unintuitive and messes up the validation, because all top-level schemas would now require it.

Yeah that `[DEFAULT]` section is jank. It's a weird hold-over from the default configparser format that we really have no use for: the main significance of it is you can reference into that section even if you don't have extended interpolation. But we always set extended interpolation, so it doesn't do anything. I definitely agree we should kill it.

I see what you mean about returning a `Config` object, I think. Looks good!

PR #157, opened 2020-01-04
Add Ragged and Padded data types, rename transform layers

honnibal:
We have a couple of types-by-convention currently that would be better as proper types.

* `thinc.data.Ragged`: This is the ragged array format where the data is concatenated, and there's an array with the lengths to represent the sequence information. We can just have this as `ragged.data` and `ragged.lengths`
* `thinc.data.Padded`: The LSTM PR introduces another way of representing sequences, where we pad the sequences instead of concatenating them, and pass through an array indicating where the sequences end.

We can use these types to clear up the `with_flatten`, `flatten_add_lengths`, etc transforms, giving them better names and making them easier to explain. I think we should name them `x2y` and `with_x2y`, e.g. `list2ragged`, `padded2list`, `padded2ragged`, `with_ragged2list` etc. The advantage of the 2 is it gives the `with_` a clearer scoping. I think `with_list2ragged` is easier to read than `with_list_to_ragged`, as the latter looks like `with list, to ragged` rather than `with list_to_ragged`.


setup.cfg
```diff
@@ -45,6 +45,7 @@ install_requires =
     srsly>=2.0.0,<3.0.0
     catalogue>=0.0.7,<1.1.0
     typer>=0.0.7,<1.1.0
+    dataclasses>=0.6,<0.7; python_version < "3.7"
```
ines: This should be moved to the third-party dependencies section

PR #159, opened 2020-01-04
Render Model Graphs

justindujardin:
Inspired by the Keras model rendering utilities, use pydot to graph and connect thinc model layers.

Add an example that renders a model architecture:

```python
model = chain(
    ExtractWindow(3),
    ReLu(n_hidden, dropout=dropout, normalize=True),
    Maxout(n_hidden * 4),
    Affine(n_hidden * 2),
    ReLu(n_hidden, dropout=dropout, normalize=True),
    Affine(n_hidden),
    ReLu(n_hidden, dropout=dropout),
    Softmax(),
)
```

model


honnibal:
Hm. I agree with the spirit of supporting a `name` argument to the `Model` class, but it does complicate the `chain` implementation a bit. Can we drop that part of the PR for now, until we come up with an implementation that looks a bit nicer and will work across the different layers?

ines:
I already reverted this part in #161, which is only adding the visualizers.

PR #162, opened 2020-01-04
Add generic input/output types to Model class

justindujardin:
Screen Shot 2020-01-04 at 2 26 37 PM


honnibal:
Awesome! Will have to work through the layers and add try out typing. I'll get my array-stub PR merged as well.

PR #169, opened 2020-01-05
Experiment with representing dimension values in the type system

honnibal:
We can't expect the type checking to provide full validation over literal types, as we won't be able to express something like "concatenate returns output with dimensionality equal to the sum of its layers' dimensionality. However, even if the checking doesn't really work properly, it's worth discussing what this could look like, and what we can and can't do with it.

I'm not 100% sure I'm doing this right, but I've tried to make the `Floats1d`, `Floats2d` etc arrays into generics, giving us space to pass type variables in to represent the dimensions. I've taken care to use distinct variables for each class. I'm not sure this is actually necessary? I didn't want mypy to think the dimensions across the classes matched the same values or something.

Arguably we could avoid having the `Floats1d`, `Floats2d` etc classes, and make these generic with respect to number of dimensions. I think that's probably worse: I think it's worth being able to express `Floats2d` concisely, and to me it seems to make things simpler and easier to work with. If it turns out making this a parametric type with a literal is better, sure we can do that.

The main thing that I think is neat is having a way to document the expected dimensionalities in the code.

I have no idea what conventions we should use to distinguish the dimension type variables from real variables, though. We want the type variables to have concise names, but we still want concise names available within the code. I think we need a better variable convention in general around this stuff. For instance, maybe weights parameters should be named like `wW`, `wB` by convention? Then their gradients could still be `dW`.

### DIFF
--- a/thinc/layers/relu.py
+++ b/thinc/layers/relu.py
@@ -1,5 +1,4 @@
 from typing import Tuple, Callable, Optional, TypeVar, cast
-from typing_extensions import Literal
 
 from ..model import Model, create_init
 from ..initializers import xavier_uniform_init, zero_init
@@ -7,7 +6,6 @@ from ..types import Array, Floats1d, Floats2d
 from .chain import chain
 from .layernorm import LayerNorm
 from .dropout import Dropout
-
 
 B = TypeVar("B", bound=int)
 O = TypeVar("O", bound=int)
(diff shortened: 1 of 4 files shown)


honnibal:
Hm, Python 3.8 fails due to method resolution order? Seems hard to fix =/

thinc/layers/relu.py
```diff
@@ -17,7 +21,7 @@ def ReLu(
     dropout: Optional[float] = None,
     normalize: bool = False,
 ) -> Model:
-    model = Model[Floats2d, Floats2d](
+    model = Model[Floats2d[B, O], Floats2d[B, O]](
```
svlandeg: should the first one be `[B, I]` instead of `[B, O]` ?

svlandeg:
It seems to work when you put `Array` first like you did for `Floats1d` and `Floats2d` ?

honnibal:
Closing this: I think the experiment really leads to the conclusion that the dimension-based typing isn't a good approach, certainly for now, and possibly in general. I still find the documentation aspect pretty appealing, but the complexity just doesn't work, and it messes up the type variables we want to use for too many methods.

PR #171, opened 2020-01-06
Add/update for spaCy

honnibal:
Add some stuff spaCy needs


thinc/layers/__init__.py
```diff
@@ -22,8 +23,14 @@
 from .noop import noop
 from .recurrent import recurrent
 from .uniqued import uniqued
+from .siamese import siamese
```
ines:
```suggestion
from .siamese import Siamese
```

honnibal: I want that layer to be `siamese`, as it's a weightless combinator.

PR #172, opened 2020-01-06
Add a thinc.api module that provides a flat namespace

honnibal:
While porting spaCy to the new Thinc, I started having trouble remembering what was where, for a couple of functions.

I suggest we add a flat namespace file, `thinc.api`, that imports everything in the library. The library itself should never import this module, to avoid circular imports. This also means we can't import this module during the tests, except perhaps in the body of a test file.

Users will be free to optionally just work directly with `thinc.api`. If they choose to do that, they'll also gain some protection against breaking changes that result from us moving things around. I think it's okay to still have the actual locations of files as part of the interface, but if we wanted to go one step further, we could make everything but `thinc.api` an underscored module, so that the only interface into the library is the flat `thinc.api` namespace. I think that's worse though --- I'd rather have `thinc.api` as just an option.

I've also added `__all__` to the various modules.

It's kind of nice to see the API all in one place though! It's pretty small.


thinc/api.py
```diff
@@ -0,0 +1,83 @@
+# flake8: noqa
```
ines: We should just add this file to the excludes in the `setup.cfg` (at least that's how I've been doing it for the other files – or maybe this solution is cleaner, not sure.)

thinc/api.py
```diff
@@ -0,0 +1,83 @@
...
+from .loss import categorical_crossentropy, L1_distance, cosine_distance
+from .config import Config
+from .initializers import normal_init, uniform_init, xavier_uniform_init, zero_init
+from .loss import categorical_crossentropy, L1_distance, cosine_distance
+from .model import create_init, Model
+from .optimizers import Adam, RAdam, SGD, Optimizer, ADAM_DEFAULTS, SGD_DEFAULTS
+from .schedules import (
```
ines: Maybe a personal preference but I typically use 2 lines for imports if they break (just takes up too much space otherwise and code becomes hard to read). I can fix this in this PR if you want.

svlandeg:
I think this is a great idea, and actually a nice overview of everything that's available in `thinc` :-) I wouldn't make the use of this mandatory though. I can imagine some users wanting to use the imports of the original locations to be able to find the original source code faster, for instance. But others may enjoy the additional abstraction that makes their code more maintainable !

thinc/api.py
```diff
@@ -0,0 +1,83 @@
...
+    siamese,
+    MaxPool,
+    MeanPool,
+    SumPool,
+    list2ragged,
+    ragged2list,
+    with_list2array,
+    with_list2padded,
+    with_reshape,
+)
```
svlandeg: The order of these seem a little arbitrary. Part of it is alphabetical and part of it is ordered by type, it's kind of confusing to find anything back..

PR #174, opened 2020-01-06
Fix CI build for Python 3.8

justindujardin:
Python can't resolve the order of methods when a class extends the same base twice.

In this case, we have `class Array(Generic[t])` and then the Floats/Ints types which extend `Array(Generic[t])` and `Generic[t,...]`

Looking at the Array type it appears the argument it wants is the number of dimensions, so fill Floats/Ints types with Literal[ndim] in the implementation.

I'm not sure, but I suspect that specifying the type for our Array usage disambiguates the base `Generic` type, allowing python to derive the correct call order. Probably because the Array generic no longer has any types that have to be filled in.


thinc/types.py
```diff
@@ -443,11 +467,11 @@ def __get_validators__(cls):
             yield validator
 
 
-f2d1 = TypeVar("f2d1", bound=int)
```
justindujardin: I removed the binding here because it doesn't appear like typing will deal with multiple types, and it says that bindings shouldn't be Union types. Without removing the binding you can't use Floats2d to represent an unknown dimension. I guess if we really wanted to bind to int we could use something like -1 to represent a None dimension?

thinc/types.py
```diff
@@ -1,19 +1,41 @@
...
+            from typing_extensions import Literal
+        except ImportError:
+            msg.fail(
+                title="Incompatible System",
+                text="Either Python 3.8 must be used or the "
+                "`typing_extensions` module must be installed",
+                exits=1,
+            )
+else:
+    from typing import Literal
```
ines: If we do this, we should just make `typing_extensions` a dependency for < 3.8 (just like we currently do with `dataclasses` for < 3.7). And then we can assume that the user always has `Literal`.

honnibal:
Aaah, I see what you did there, to get around that `MRO` problem. Interesting!

It's still an open question whether this is indeed better: I can see advantages and disadvantages. I feel like the error messages from the generic types get pretty ugly, and it's a lot of complexity for probably limited amounts of actual error checking?

thinc/types.py
justindujardin: reverted this for now since we're not doing shape checking

thinc/types.py
justindujardin: Yeah, I did that and forgot to remove it here. It's cleaned up now

justindujardin:
Cool, I removed the example with dimension type checks for now. 

I think the complexity could be brought down in a bunch of ways from that example if we revisit it later

PR #175, opened 2020-01-06
Feature/array ops memory types

justindujardin:
Add generic types for Array, Ops, and Memory classes

This probably needs a decent review on the ops function return types. They were all plain Array, and I made them infer the ArrayT from the passed in variables. I think this is right in most cases, but that may not be true.

This builds on top of https://github.com/explosion/thinc/pull/174


justindujardin:
I didn't have specific examples to verify the types line up as expected, so I had to follow the errors from mypy as a guide for whether the types were right or not.

The general idea is that the type has to be carried either by:
 - A Class `self` variable like https://github.com/explosion/thinc/pull/175/files#diff-bbf863a9d29a482c9040df4c4aca7837R113
 - A function/method argument like https://github.com/explosion/thinc/pull/175/files#diff-d76e2c413ca47ce6c5567ca627156b53R224

PR #177, opened 2020-01-07
Add generic types to Array and Ops.allocate

justindujardin:
- for array generics work fine to give back the type of the item passed in. We could further expand these types to all the self arguments in the Array class.
 - add overloads for Ops.allocate that support up to 4d data with return types of Floats(n)d
 - the ND case is tricky and doesn't play nice with other overloads (because it catches all of them) so separate it out into `allocate_nd`


justindujardin:
There are a few mypy errors to fix from the addition of new (more strict) types. I'll clean them up in the morning.

honnibal:
If we need to differentiate the unlimited dimension vs fixed dimension case anyway, maybe we really should go the other way and do `alloc_f1d`, `alloc_f2d` etc?

justindujardin:
Sure, I can change it to be like that if you prefer. I did it this way to avoid a bunch of new methods, and this looked like a decent in-between where you mostly use one function and resort to the _nd function for things with unknown/dynamic shapes.

The other thing I was thinking is that you lose some ease-of-use style cases with specific functions for each dim, e.g., if you already have the shape expressed as a list, you can pass directly as an argument rather than picking them off:
```python
shape = [1,2]
# This is okay now
a:Floats2d = ops.allocate(shape)
# And is usable with Tensorflow as well
a_tf: tf.Tensor = tf.ones(shape=shape)
# Here you have to pick out the dimensions
b: Floats2d = ops.alloc_f2d(shape[0], shape[1])
# But here it's more natural and you save []
c: Floats2d = ops.alloc_f2d(1, 2)
```
I can do it either way, let me know which way you prefer and I'll update the code. If we do the `_f2d` we'll also want the `_i2d` version, yeah?

honnibal:
I do think it's better to have the individual functions: to me it's nicer to have the general name cover the most general case, and then the more specific methods with more specific names. I guess another way of saying it is, if you saw `allocate()` and `allocate_nd` as method names, what would you guess the difference was between them? I guess I'd think `allocate` could only give me a flat vector, so I'd probably try to call it with an integer. I doubt I'd guess that actually the two methods are identical at runtime, and the `allocate_nd` is only necessary due to a technicality of the type-checking.

Btw I think you shouldn't have to pick out the dimensions:

```python
b: Floats2d = ops.alloc_f2d(*shape)
```

You'd also get a runtime error there if you're wrong about what you think `shape` is, which is kind of nice.

One question is, if it will be `alloc_f2d` etc, I guess it should be `alloc`?

justindujardin:
Cool, I'll start with the shortened names like you suggest, and if you don't like them we can switch back

thinc/backends/numpy_ops.pyx
```diff
@@ -55,7 +55,7 @@ class NumpyOps(Ops):
         else:
             return self.xp.array(data)
 
-    def allocate(self, shape, dtype='float32'):
+    def alloc(self, shape, dtype='float32'):
         if isinstance(shape, int):
```
honnibal: Let's kill this `isinstance` condition? I think it was a hold-over from oldthinc.

thinc/layers/embed.py
```diff
@@ -40,10 +40,10 @@ def forward(model: Model[InT, OutT], ids: InT, is_train: bool) -> Tuple[OutT, Ca
     output = vectors[ids]
 
     def backprop(d_output: OutT) -> InT:
-        d_vectors = model.ops.allocate(vectors.shape)
+        d_vectors: Floats2d = model.ops.alloc(vectors.shape)
```
honnibal:
```suggestion
        d_vectors = model.ops.alloc_f2d(*vectors.shape)
```

thinc/layers/embed.py
```diff
@@ -40,10 +40,10 @@ def forward(model: Model[InT, OutT], ids: InT, is_train: bool) -> Tuple[OutT, Ca
...
         model.ops.scatter_add(d_vectors, ids, d_output)
         model.inc_grad("vectors", d_vectors)
-        dX = cast(Ints2d, model.ops.allocate(ids.shape, dtype=ids.dtype))
+        dX: Ints2d = model.ops.alloc(ids.shape, dtype=ids.dtype)
```
honnibal:
```suggestion
        dX model.ops.alloc(*ids.shape)
```

thinc/layers/embed.py
```diff
@@ -56,7 +56,7 @@ def init(
         if Y is not None:
             model.set_dim("nO", get_width(Y))
         shape = (model.get_dim("nV"), model.get_dim("nO"))
-        vectors = initializer(model.ops.allocate(shape))
+        vectors = initializer(model.ops.alloc(shape))
```
honnibal:
```suggestion
        vectors = initializer(model.ops.alloc_f2d(*shape))
```

thinc/layers/hashembed.py
```diff
@@ -45,20 +45,20 @@ def forward(model: Model[InT, OutT], ids: InT, is_train: bool) -> Tuple[OutT, Ca
 
     def backprop(d_output: OutT) -> InT:
         keys = model.ops.hash(ids, seed) % nV
-        d_vectors = model.ops.allocate(vectors.shape)
+        d_vectors = model.ops.alloc(vectors.shape)
```
honnibal:
```suggestion
        d_vectors = model.ops.alloc_f2d(*vectors.shape)
```

thinc/layers/hashembed.py
```diff
@@ -45,20 +45,20 @@ def forward(model: Model[InT, OutT], ids: InT, is_train: bool) -> Tuple[OutT, Ca
...
         keys = model.ops.xp.ascontiguousarray(keys.T, dtype="i")
         for i in range(keys.shape[0]):
             model.ops.scatter_add(d_vectors, keys[i], d_output)
         model.inc_grad("vectors", d_vectors)
-        dX: InT = cast(Ints2d, model.ops.allocate(ids.shape, dtype=ids.dtype))
+        dX: InT = model.ops.alloc(ids.shape, dtype=ids.dtype)
```
honnibal:
```suggestion
        dX = model.ops.alloc_i2d(*ids.shape)
```

thinc/layers/hashembed.py
```diff
@@ -45,20 +45,20 @@ def forward(model: Model[InT, OutT], ids: InT, is_train: bool) -> Tuple[OutT, Ca
...
         return dX
 
     return output, backprop
 
 
 def create_init(initializer: Callable) -> Callable:
     def init(model: Model, X: Optional[InT] = None, Y: Optional[OutT] = None) -> Model:
-        vectors = model.ops.allocate((model.get_dim("nV"), model.get_dim("nO")))
+        vectors = model.ops.alloc((model.get_dim("nV"), model.get_dim("nO")))
```
honnibal:
```suggestion
        vectors = model.ops.alloc_f2d(model.get_dim("nV"), model.get_dim("nO"))
```

thinc/layers/layernorm.py
```diff
@@ -49,8 +49,8 @@ def init(
         model.set_dim("nI", Y_width)
         model.set_dim("nO", Y_width)
     nO = model.get_dim("nO")
-    model.set_param("G", model.ops.allocate((nO,)))
-    model.set_param("b", model.ops.allocate((nO,)))
+    model.set_param("G", model.ops.alloc((nO,)))
```
honnibal:
```suggestion
    model.set_param("G", model.ops.alloc_f1d(nO))
```

thinc/layers/layernorm.py
```diff
@@ -49,8 +49,8 @@ def init(
...
+    model.set_param("b", model.ops.alloc((nO,)))
```
honnibal:
```suggestion
    model.set_param("b", model.ops.alloc_f1d(nO))
```

thinc/layers/lstm.py
```diff
@@ -132,12 +132,12 @@ def backprop_gates(
     ) -> Tuple[Floats2d, Floats2d]:
         d_cells = d_cells[:size]
         d_hiddens = d_hiddens[:size]
-        d_acts = ops.allocate(acts.shape)
-        d_prevcells = ops.allocate(prev_cells.shape)
+        d_acts: Floats2d = ops.alloc(acts.shape)
```
honnibal:
```suggestion
        d_acts: Floats2d = ops.alloc_f2d(*acts.shape)
```

thinc/layers/maxout.py
```diff
@@ -73,10 +73,10 @@ def init(
             model.set_dim("nI", get_width(X))
         if Y is not None:
             model.set_dim("nO", get_width(Y))
-        W = model.ops.allocate(
+        W = model.ops.alloc(
```
honnibal:
```suggestion
        W = model.ops.alloc_f3d(
```

thinc/layers/maxout.py
```diff
@@ -73,10 +73,10 @@ def init(
...
             (model.get_dim("nO"), model.get_dim("nP"), model.get_dim("nI"))
```
honnibal:
```suggestion
            model.get_dim("nO"), model.get_dim("nP"), model.get_dim("nI")
```

thinc/layers/maxout.py
```diff
@@ -73,10 +73,10 @@ def init(
...
         )
-        b = model.ops.allocate((model.get_dim("nO"), model.get_dim("nP")))
+        b = model.ops.alloc((model.get_dim("nO"), model.get_dim("nP")))
```
honnibal:
```suggestion
        b = model.ops.alloc_f2d(model.get_dim("nO"), model.get_dim("nP"))
```

honnibal:
Looking good! I only got part of the way through switching to the more specific allocation functions. I think the result looks pretty good, and it definitely makes the types easy.

thinc/layers/embed.py
justindujardin: This needed to use alloc_i2d instead, so I applied it separately

honnibal:
🎉

PR #178, opened 2020-01-07
Track origin of ConfigValidationError

svlandeg:
When running a config file through validation, the current code would show a tabular format of errors from Pydantic, but without specifying where exactly in the config they came from:

```
Config validation error

args                 field required
kwargs               field required
depth                extra fields not permitted
embed_size           extra fields not permitted
maxout_pieces        extra fields not permitted
pretrained_vectors   extra fields not permitted
width                extra fields not permitted
window_size          extra fields not permitted
```

This PR changes that into: 

```
Config validation error

nlp.pipeline.tok2vec.model   args                 field required
nlp.pipeline.tok2vec.model   kwargs               field required
nlp.pipeline.tok2vec.model   depth                extra fields not permitted
nlp.pipeline.tok2vec.model   embed_size           extra fields not permitted
nlp.pipeline.tok2vec.model   maxout_pieces        extra fields not permitted
nlp.pipeline.tok2vec.model   pretrained_vectors   extra fields not permitted
nlp.pipeline.tok2vec.model   width                extra fields not permitted
nlp.pipeline.tok2vec.model   window_size          extra fields not permitted
```

By keeping track of the keys we're running through as we're parsing and validating the config.

### DIFF
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -104,7 +104,7 @@ class ConfigValidationError(ValueError):
         for error in errors:
             err_loc = " -> ".join([str(p) for p in error.get("loc", [])])
             # appending element at the end so it doesn't matter if it's empty
-            data.append((err_loc, error.get("msg", element)))
+            data.append((err_loc, error.get("msg"), element))
         result = [message, table(data), f"{config}"]
         ValueError.__init__(self, "\n\n" + "\n".join(result))
 
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -103,10 +103,8 @@ class ConfigValidationError(ValueError):
         data = []
         for error in errors:
             err_loc = " -> ".join([str(p) for p in error.get("loc", [])])
-            if element:
-                data.append((element, err_loc, error.get("msg")))
-            else:
-                data.append((err_loc, error.get("msg")))
+            # appending element at the end so it doesn't matter if it's empty
+            data.append((err_loc, error.get("msg", element)))
         result = [message, table(data), f"{config}"]
         ValueError.__init__(self, "\n\n" + "\n".join(result))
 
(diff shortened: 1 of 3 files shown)


thinc/config.py
```diff
@@ -201,9 +203,10 @@ def _fill(
         filled: Dict[str, Any] = {}
         validation: Dict[str, Any] = {}
         for key, value in config.items():
+            key_parent = str(parent + "." + key).strip(".")
```
ines:
```suggestion
            key_parent = f"{parent}.{key}".strip(".")
```

thinc/config.py
```diff
@@ -97,12 +97,13 @@ def __init__(
...
         errors: List[Dict[str, Any]],
         message: str = "Config validation error",
+        element: str = "",
     ) -> None:
         """Custom error for validating configs."""
         data = []
         for error in errors:
             err_loc = " -> ".join([str(p) for p in error.get("loc", [])])
-            data.append((err_loc, error.get("msg")))
+            data.append((element, err_loc, error.get("msg")))
```
ines: We should probably do an `if element:` / `else:` here so we don't add an empty string column (plus column spacing) if the error is raised with `element` unset.

svlandeg: mypy is not happy about this solution ...

svlandeg: We can probably get that sorted somehow, or the other option is to add the origin to the last column of the table instead of the first. Then the emptyness at the end wouldn't matter in cases when `element` is empty ?

ines: Ah okay, nevermind. It's not that big of a deal. I'll go ahead and merge this, and see if maybe I can find a way to solve this.

PR #183, opened 2020-01-08
fix input type for concatenate to support "a | b"

svlandeg:
Changed the signature for `concatenate()` to `*layers: Model` instead of using a `List`, to allow using the operator in cases like `prefix | suffix`. The current version gave the error 

> concatenate() takes 1 positional argument but 2 were given


honnibal:
Thanks! I typed this wrongly. I think we can improve this later actually: we should use type variables so that it's `Model[InT, OutT]`: the concatenate operator only works on inputs that are the same type.

PR #187, opened 2020-01-08
Fix python3.6 support for TensorflowWrapper

justindujardin:
`tfdlpack` only has a pip package for python 3.7 right now. This PR removes tfdlpack and adds more tests for the Tensorflow Wrapper

Changes:
- [x] remove tfdlpack
- [x] add more tests for TensorflowWrapper
- [x] install Tensorflow during CI build so wrapper tests are executed


thinc/tests/layers/test_tensorflow_wrapper.py
```diff
@@ -1,19 +1,137 @@
...
+    with pytest.raises(ValueError):
+        model_bytes = model.to_bytes()
+    # After predicting, the model is built
+    model.predict(X)
+    # And can be serialized
+    model_bytes = model.to_bytes()
+    assert model_bytes is not None
+
+
+@pytest.mark.skip("keras load fails with weights mismatch")
```
justindujardin: FYI: The TF wrapper fails to `from_bytes` properly with a Keras error. I don't know how to fix it immediately.

setup.cfg
```diff
@@ -75,7 +75,7 @@ torch =
     torch>=1.0.0,<1.4.0
 tensorflow =
     tensorflow>=2.0.0,<2.1.0
-    tfdlpack>=0.1.1,<0.2.0
+    tfdlpack>=0.1.1,<0.2.0; python_version == "3.7"
```
honnibal: Maybe we should abandon the tfdlpack thing altogether?

Like, it only installs on Linux on particular Python versions, it requires tricky pip installation, it's unclear whether it'll be maintained well in the future, and it's not even obvious it will make a big performance difference.

thinc/util.py
```diff
@@ -279,7 +277,7 @@ def xp2tensorflow(
 def tensorflow2xp(tensorflow_tensor: "tf.Tensor") -> Array:  # pragma: no cover
     """Convert a Tensorflow tensor to numpy or cupy tensor."""
     assert_tensorflow_installed()
-    if "GPU" in tensorflow_tensor.device:
+    if "GPU" in tensorflow_tensor.device and has_tfdlpack:
         return cupy.fromDlpack(tfdlpack.to_dlpack(tensorflow_tensor))
```
honnibal: I think we should abandon DLPack and just find a different way to copy the tensor over to cupy.

justindujardin: Okay, I'm not sure what a replacement would look like, do you want me to remove it for now and someone else can add a replacement at a later point? Or do you want to prescribe a set of changes and a replacement library for this PR?

honnibal: Let's just remove it. Since it's a small change and the time difference makes this a bit async, I'll just merge and adjust :)

PR #188, opened 2020-01-08
WIP: Trouble with recursive objects in config

svlandeg:
I'm pretty certain there's a bug in parsing recursive objects in thinc's `_fill` function on the current `develop` branch. I ran into it while testing from spaCy, so I tried creating a minimal example for testing here in `test_config.py`, avoiding generators and iterators etc for simplicity.

It looks like an object within an object is not returned properly. I tried debugging with this example. The `my_cool_repetitive_schedule.v1` creates the correct `learn_rate` object `[0.001, 0.001, 0.001, 0.001]` which is correctly [stored](https://github.com/explosion/thinc/blob/develop/thinc/config.py#L226) in `validation[key]`. But then when the recursive loop goes on to fill in `my_cool_optimizer.v1`, it doesn't have access to that object anymore, and fills in the original json `{'@schedules': 'my_cool_repetitive_schedule.v1', 'base_rate': 0.001, 'repeat': 4}` as `learn_rate` parameter instead.

You'll see this when the unit test crashes on `assert`. While the `optimizer` object is correct and has a proper field `b1` (from `beta1`), its `learn_rate` field is still a json string instead of the `List`.

I've been trying to fix this but no luck so far. @honnibal: could you have a look ?

PS: also fixed `EXAMPLE_CONFIG` which had some invalid arguments for `warmup_linear.v1`

### DIFF
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -214,14 +214,14 @@ class registry(object):
             key_parent = f"{parent}.{key}".strip(".")
             if cls.is_promise(value):
                 promise_schema = cls.make_promise_schema(value)
-                filled[key], _ = cls._fill(
+                filled[key], validation[key] = cls._fill(
                     value, promise_schema, validate, parent=key_parent
                 )
                 # Call the function and populate the field value. We can't just
                 # create an instance of the type here, since this wouldn't work
                 # for generics / more complex custom types
-                getter = cls.get_constructor(filled[key])
-                args, kwargs = cls.parse_args(filled[key])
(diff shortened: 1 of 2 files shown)


ines:
Thanks for the test case – I'll take a look!

ines:
I think I got it! Took a bit of trial and error but I think the problem here was that we weren't generating the arguments that are passed into the function from the already generated objects. And it looks like we were also not passing the tree of generated objects down correctly.

PR #191, opened 2020-01-09
Friendly debugging errors

svlandeg:
1. Model/layer names in error msg helps debugging
2. Ensure HashEmbed has a default `None` `nI` dimension
3. Throw an error when a none-`None` dimension is set to a different value (this shouldn't really happen, right ?)

### DIFF
--- a/thinc/layers/maxout.py
+++ b/thinc/layers/maxout.py
@@ -18,7 +18,7 @@ OutT = Floats2d
 def Maxout(
     nO: Optional[int] = None,
     nI: Optional[int] = None,
-    nP: int = 3,
+    nP: Optional[int] = 3,
     *,
     init_W: Callable = xavier_uniform_init,
     init_b: Callable = zero_init,
--- a/thinc/layers/concatenate.py
+++ b/thinc/layers/concatenate.py
@@ -44,7 +44,12 @@ def init(
         X_width = get_width(X)
         model.set_dim("nI", X_width)
         for layer in model.layers:
-            layer.set_dim("nI", X_width)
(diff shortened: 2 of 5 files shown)


svlandeg:
(Maybe hold off on merge, will probably be adding few more smaller edits to this one in the next few hours)

svlandeg:
Ready for review. I'm not sure about throwing the error for dimension changes, but it's been really helpful for me for debugging so far. Maybe it should be an optional error or a warning instead ?

thinc/layers/maxout.py
```diff
@@ -18,7 +18,7 @@
 def Maxout(
     nO: Optional[int] = None,
     nI: Optional[int] = None,
-    nP: int = 3,
+    nP: Optional[int] = 3,
```
honnibal: This doesn't need to be `Optional`, because it should always be an int I think? I don't think we need to set it to `None`.

honnibal:
Looks good! Definitely agree about the error messages.

I think we should have the `set_dim` error for now, because currently this will result in broken behaviour. In future we can look at supporting resizable layers. Maybe it should always be a create-and-replace though? Resize in-place will probably cause lots of problems.

thinc/layers/maxout.py
svlandeg: Yea, so we need this if we want to keep the `set_dim` error. Otherwise, during serialization, `nP` gets first set to the default value 3 and then overwritten to its actual value from disk, which throws an error. That's why now it reads `nP=None` in `test_serialize`, that test crashes otherwise.

PR #194, opened 2020-01-09
Add mypy tests to test suite WIP

justindujardin:
It's come up that types are not always clear, and knowing that we're returning `Model[Floats2d, Floats2d]` instead of `Any` is important.

Mypy has existing test machinery that is used in other plugins, so I copied it. It's pretty gross but does the job of asserting about types.

The tests exist in `thinc/tests/test-data/unit` which is hardcoded in the mypy test runner, so we can't rename or simplify the hierarchy. :sob:

They are plaintext files ending in `.test` extension, and have a syntax similar to ini files with bracketed sections:
```
[case test_mypy_control_flow_type_discrimination]
from typing import Optional

def my_foo(op: Optional[int]):
   reveal_type(op) # N: Revealed type is 'Union[builtins.int, None]'
   assert op is not None
   reveal_type(op) # N: Revealed type is 'builtins.int'
[out]
```

The comment after `reveal_type` is used by the test suite to make that an assertion. You can use wildcards in the asserted type name if you want to be less specific, e.g. `builtins.*`

This also cleans up the return types for model forward functions, and adds a test:
```python
from thinc.model import Model, ModelFwd
from thinc.layers import Linear
from thinc.types import Floats2d, Array
from thinc.backends import get_current_ops, Ops

ops: Ops = get_current_ops()
example = ops.alloc_f2d(1, 12)
label = ops.alloc_f2d(1, 1)
model = Linear(12)

def forward(model: Model, X: Floats2d, is_train: bool) -> ModelFwd[Floats2d]:
    def backprop(dY: Floats2d) -> Floats2d:
        pass

    return label, backprop


reveal_type(forward(model, example, False))
# Revealed type is 'Tuple[thinc.types.Floats2d, def (thinc.types.Floats2d) -> thinc.types.Floats2d]'

```


conftest.py
```diff
@@ -0,0 +1,8 @@
+# NOTE: This is required for the MyPy test suite machinery to pick up
```
ines: Does this file have to be at the root? Or can we move this to `thinc/tests/conftest.py`?

justindujardin: Yeah, sadly it has to be here in order to register the mypy test runner; if it's in a subfolder pytest complains. 

I'm thinking we could refactor this away and use our own mypy test runner that's less complicated and hardcoded. I skipped it for now to keep things simple until we verify it's useful

justindujardin: I think I found a decent solution. I inlined the test runner bits we need from mypy and removed the various ENV and conftest.py hacks from before

justindujardin:
I talked with @tiangolo and the test setup from pydantic looks nicer than this because it offers a better developer experience using `.py` files that IDEs will highlight. Closing this in favor of that.

PR #195, opened 2020-01-09
WIP: Fix small misconfigs in example, initial pass

tiangolo:
:bug: Fix small misconfigurations in example script.

Just an initial pass that fixes some initial problems detected while using my local branch of Pydantic (with support for generics).

It depends on https://github.com/samuelcolvin/pydantic/pull/1159 merged and released.

There's also a small change to `thinc.model.Model`, I haven't committed it yet, and it might be unnecessary, depending on the end result of the PR in Pydantic.

### DIFF
--- a/examples/basic_pos_tagger_from_config.py
+++ b/examples/basic_pos_tagger_from_config.py
@@ -29,32 +29,32 @@ learn_rate = 0.001
 [model.*.with_list2array]
 @layers = "with_list2array.v0"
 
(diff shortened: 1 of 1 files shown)


ines:
I'll already go ahead and merge this so we don't have any actual typos and mistakes in the config example here. Once the update is released with pydantic, we can update `Model` if needed and remove the comments about the generics not working (both in the example and in `config.py`).

PR #197, opened 2020-01-10
Add Decorator type to fix registry functions

honnibal:
Registered functions were becoming type-opaque, because of the registry decorator. Even after adding type annotations in catalogue, there were still problems getting it to work properly, because we assign the decorator to a class variable, which makes things that much more opaque to mypy.

I tried a few approaches, but it's complicated having an interaction between the type variable and the class. I don't really know what's going on but it complained of some sort of ambiguity.

My solution is to use a "protocol" to define a new type, `Decorator`. We can then declare the variable to be of that type when we assign it. This works! I even got a type error in my transformers script.

There's still some tricks to be solved with the `chain` operator. Not sure what's up but I don't always get type errors from incorrect combinations.

### DIFF
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -145,7 +145,9 @@ class registry(object):
         """Create a new custom registry."""
         if hasattr(cls, registry_name):
             raise ValueError(f"Registry '{registry_name}' already exists")
-        reg: Decorator = catalogue.create("thinc", registry_name, entry_points=entry_points)
+        reg: Decorator = catalogue.create(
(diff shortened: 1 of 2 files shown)


thinc/config.py
```diff
@@ -1,5 +1,6 @@
 from typing import Union, Dict, Any, Optional, List, Tuple, Callable, Type, Sequence
 from types import GeneratorType
+from .types import Decorator
```
ines: This should be moved down onto line 14 to separate absolute from  relative imports

PR #199, opened 2020-01-10
Fixing chain ini and maxout backprop

svlandeg:
- The initialization of `chain` would throw an error if the underlying layer didn't not properly handle `predict(None)`, so preventing that here for robustness.
- `maxout` had a bug in its backprop function: the value to increment `b` needs to be reshaped to `(nO, nP)` (right?)
- Raising an informative error when these sort of wrong shapes occur in `inc_grad`

### DIFF
--- a/thinc/layers/chain.py
+++ b/thinc/layers/chain.py
@@ -84,7 +84,8 @@ def init(model: Model, X: Optional[InT] = None, Y: Optional[OutT] = None) -> Non
             break
     for layer in model.layers[:-1]:
         layer.initialize(X=X)
-        X = layer.predict(X)
+        if X is not None:
+            X = layer.predict(X)
     model.layers[-1].initialize(X=X, Y=Y)
     if model.layers[0].has_dim("nI"):
         model.set_dim("nI", model.layers[0].get_dim("nI"))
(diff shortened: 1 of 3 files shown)


thinc/layers/chain.py
```diff
@@ -84,7 +84,8 @@ def init(model: Model, X: Optional[InT] = None, Y: Optional[OutT] = None) -> Non
             break
     for layer in model.layers[:-1]:
         layer.initialize(X=X)
-        X = layer.predict(X)
+        if X is not None:
+            X = layer.predict(X)
```
honnibal: I think we probably want the for loop under the conditional, rather than the conditional on the inside? It should only be the initial `X` that's allowed to be `None`.

thinc/layers/maxout.py
```diff
@@ -57,7 +57,7 @@ def backprop(d_best: OutT) -> InT:
         dY = model.ops.backprop_maxout(d_best, which, nP)
         dY = dY.reshape((dY.shape[0], nO * nP))
         model.inc_grad("W", model.ops.gemm(dY, X, trans1=True).reshape((nO, nP, nI)))
-        model.inc_grad("b", dY.sum(axis=0))
+        model.inc_grad("b", dY.sum(axis=0).reshape((nO, nP)))
```
honnibal: Probably clearer to just move the `inc_grad` call above the `dY.reshape` above.

thinc/layers/chain.py
svlandeg: I actually tried that first because I thought it was more clean, too. But then you have to copy the for loop because the intermediate layers may not get initialized, so you would run them with `layer.initialize()` in the second `for` loop. But that gets more verbose and boils down to the same thing as giving them `X=None`?

honnibal: Hm, I see. Okay!

PR #200, opened 2020-01-10
Improve transforms and Padded type

honnibal:
More work towards transformers examples (which need the Padded type to be working well).

The `Padded` type previously lost information about how to restore the batch to its original order. We now track two additional lists with it. In total this makes the data structure:

* `data`: A 3d array, sorted by decreasing sequence length. The dimensions are timestep, batch item, row data.
* `size_at_t`: An array indicating how the batch can be truncated at different sequence lengths. You can do `data[:, :size_at_t[t]]` to get an unpadded batch. So let's say you have a batch of 4 documents, of lengths `[6, 5, 2, 1]`. The `size_at_t` will be `[4, 3, 3, 3, 2, 1]`.
* `lengths`: A list of ints, indicating the sequence lengths. Applies to the reordered sequences, not the original ordering. So it'll be decreasing length.
* `indices`: Indicates how to put the items back into original order.

I've also killed a redundant padding op, and given the remaining one a more sensible name and behaviour.

I've also added a couple of transforms. I think we have too many of these and we should considate on a more polymorphic set:

* `with_array`
* `with_padded`
* `with_ragged`
* `with_list`

They would each take a type-var of `Union[List[Array], Padded, Ragged]`, producing the same type on output. ~I plan to add that to this PR.~ Done! Made sure to test for 0-length sequences as well -- caught a few bugs. I used a pretty tricky set of pytest fixtures to test all four modules to 100% with one file.


thinc/layers/with_list2padded.py
```diff
@@ -2,7 +2,6 @@
```
ines: Should this file be deleted? This is now also in `with_padded`.

PR #211, opened 2020-01-13
fix model serialization

svlandeg:
Fix model (de)serialization:

- `int` has a `to_bytes(length, ...)` method so we need to prevent calling that one
- `data["refs"]` was not parsed correctly in `from_bytes`


honnibal:
Argh. Bloody duck-typing...@ines is thinking about a new API for this

PR #212, opened 2020-01-13
Serializable attrs

ines:
Fixes serialization problems around the attrs.

All attrs are now (de)serialized with `msgpack` by default. For custom types, users can register their own functions (implemented via Python's [single dispatch generic functions](https://www.python.org/dev/peps/pep-0443/)).

```python
from thinc.api import serialize_attr, deserialize_attr

@serialize_attr.register(MyCustomAttr)
def serialize_my_custom_attr(_, value, name, model):
    return value.to_bytes()

@deserialize_attr.register(MyCustomAttr)
def deserialize_my_custom_attr(_, value, name, model):
    return MyCustomAttr().from_bytes(value)
```

The first argument of the function is always the instance of the attr (passed down from `Model.to_bytes`/`Model.from_bytes`). This is what the singledispatch uses to decide which function to call. In the serialization function, this is technically redundant and the same as `value` – but I still kept it in there for consistency. The functions also receive the attribute name and the `Model` instance, just in case.

Using the value of the attr to decide which serialization function to call means that custom attrs that require special (de)serialization need a default value of its type. For instance, setting `attrs={"custom": None}` won't work, since the `None` doesn't tell us that it should be `MyCustomType`. We can't easily solve this with typed dicts, and the alternative of registering loaders for attribute _names_ is messy and not really viable.

Requiring an instance of the expected value as the default value seems like an okay constraint.


honnibal:
Nice!

Btw I don't know whether this works, but is it possible to use a `Protocol` to have a dispatch that's like `HasToBytes`, and then it just calls that?

PR #220, opened 2020-01-14
Allow to set dropout rate

svlandeg:
- Fix cosine from PR #217
- Add function to change dropout rates across all layers in a model

### DIFF
--- a/thinc/backends/ops.py
+++ b/thinc/backends/ops.py
@@ -280,9 +280,14 @@ class Ops:
             return Y * (1.0 - Y)
 
     def cosine(self, X: Array, Y: ArrayT) -> float:
-        Xnorm = self.get_norm(X)
-        Ynorm = self.get_norm(Y)
-        return self.xp.dot(X, Y.T) / (Xnorm * Ynorm)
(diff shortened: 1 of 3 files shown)


honnibal:
Ah, this should be in `thinc.util`, we don't want a method for it.

There's not really any reason it needs to be a method, because we don't expect subclasses to want to change the behaviour. (Actually if we wanted that we'd have to call the method on every node, which is pretty awkward.)

We also want to keep a clear line between what sort of stuff gets into the `Model` class, and what doesn't. The `Model` should stay really generic and minimal.

svlandeg:
Ok, will fix !

svlandeg:
So I was trying to figure out why my code wouldn't run anymore, and I finally realised it's because there's a circular dependency if I move the function to `util` and set the parameter as type `Model`, because then I have to `from .model import Model` ... :|

honnibal:
Hm, yeah the circular imports can be a real pain. I think we need to break out the utils that `Model` needs into a separate module or something to ease things.

Small style point: I normally use the variable name `node` within `model.walk()`. And I know you've been testing this in your spaCy code, but a test would be good too :).

PR #223, opened 2020-01-15
Add textcat from config example

adrianeboyd:
A textcat from config example that supports two datasets with mutually exclusive classes:

* `imdb`: 2 classes, from ml-datasets
* `dbpedia_ontology`: 14 classes, loader implemented locally for fast.ai dataset on AWS

Requires `syntok` for tokenization.


adrianeboyd:
It still needs a better solution for the partial config issue.

honnibal:
Thanks!

While chaining the model together in the config like that works, it's usually not the best option so it's probably not what we want to show off in the example. What we probably want to do is register a function within the script, and then use it. For instance:

```python
@registry.layers("EmbedPoolTextcat.v0") -> Model[List[Array2d], Array2d]:
def EmbedPoolTextcat(embed: Model[Array2d, Array2d]):
    model = chain(
        list2ragged(),
        with_array(embed),
        MeanPool(),
        Softmax()
    )
    model.set_ref("embed", embed)
    return model
```

You can then reference this model within the config, so you only need to configure two models.

I guess the rule of thumb would be that we only want to be specifying meaningful units in the config, that you would want to swap out. It's sometimes handy to be able to compose in the config, but we'd rather recommend writing larger blocks in Python.

examples/textcat_from_config.py
```diff
@@ -0,0 +1,190 @@
...
+        total += len(Y)
+    return correct / total
+
+
+def tokenize_texts(texts):
+    tok = Tokenizer()
+    return [[token.value for token in tok.tokenize(text)] for text in texts]
+
+
+# Dataset loader for DBPedia Ontology from https://course.fast.ai/datasets
```
ines: I'll add this to `ml_datasets` 👍

adrianeboyd:
Okay, so here's what I started with, which doesn't seem so bad:

```
    # Define model 
    vector_width = len(vocab) 
    model = chain(list2ragged(), 
            with_array(Embed(width, vector_width)), 
            MeanPool(), 
            Softmax(nr_class, width) 
        )
```

`width` could come from a config, but `vector_width` and `nr_class` come from the dataset.

Without a config I could change it to this, which doesn't seem better, really:

```
    embed = Embed(width, len(vocab))
    model = EmbedPoolTextcat(embed)
    for layer in model.walk():
        if layer.name == "softmax":
            layer.set_dim("nI", width)
            layer.set_dim("nO", nr_class)
            layer.initialize()
```

I could modify `EmbedPoolTextcat` to set `width` from the config, but I still have `nr_class` to set for `softmax`, and I haven't figured out a better way to do it...

And I don't think I can use this version of `EmbedPoolTextcat` in a config at all because I don't see a way to set `embed`. If it has a default of `None` the config won't load, and I don't see how you could provide a value for `embed` in the config at all?

honnibal:
The following should work, via the shape inference:

```
model = chain(list2ragged(), 
    with_array(Embed(width, vector_width)), 
     MeanPool(), 
     Softmax() 
)
model.initialize(X=batch_of_inputs, Y=batch_of_outputs)
```

In most cases the `chain` function will be able to use the `Y` argument to figure out the output shape if it's missing. It can also use the `X` argument to figure out missing input shapes in many cases, although that's not relevant here.

> And I don't think I can use this version of EmbedPoolTextcat in a config at all because I don't see a way to set embed. If it has a default of None the config won't load, and I don't see how you could provide a value for embed in the config at all?

Just to check, you're referring to the vocabulary size problem, right? Because if we had the vocab size in the config this would work just fine:

```
[model]
@layers = "EmbedPoolTextcat.v0"

[model.embed]
@layers = "Embed.v0"
nO = 128
nV = 30000
```

adrianeboyd:
That example makes sense! If I make `nV` in `Embed` optional, I can get the following to work:

```
[model]
@layers = "EmbedPoolTextcat.v0"

[model.embed]
@layers = "Embed.v0"
nO = ${hyper_params:width}
```

```
    model = loaded_config["model"]
    model.get_ref("embed").set_dim("nV", len(vocab))
    model.initialize(X=train_X, Y=train_y)
```

That finally seems not-terrible.

honnibal:
@adrianeboyd Great, yes let's make `nV` optional.

honnibal:
Is this ready now? It looks good to me.

adrianeboyd:
Yes, with the Embed PR it should work.

PR #227, opened 2020-01-15
Refactor / update optimizers

ines:
As discussed with @honnibal:

* remove lookahead and related code
* make `Optimizer.__call__` take arguments `key, weights, gradient` (instead of making `key` a required keyword argument)

Also tidy up and add tests.


honnibal:
Nice! I think the Ray example might need the same change, with the optimizer? Maybe do a quick search for `optimizer(` to double check for other cases. Otherwise, LGTM!

PR #229, opened 2020-01-15
WIP: JaxOps

honnibal:
Okay so this is miles away from ready, but opening this up to explain what I've learned.

[Jax](https://jax.readthedocs.io/en/latest/index.html) is kind of the new hotness in ML engines. It features:

* numpy interface
* Autodiff as a function transformation
* JIT compilation
* Support for CPU, GPU and TPU

It's actually a perfect fit for us, and I've had one eye on Jax-compatibility during the whole redesign. The autodiff is really cool because you can write something like this:

```
def some_new_forward_function(model, X, is_train):
    W = model.get_param("W")
    b = model.get_param("b")
    Y, get_gradients = jax.vjp(_do_work, X, W, b)
    def backprop(dY):
        dX, dW, db = get_g(dY)
        model.inc_grad("W", dW)
        model.inc_grad("b", db)
        return dX

    return Y, backprop

def _do_work(X, W, b):
    # This could be pretty much any maths, with some
    # limitations (but not many)
    return X @ W.T + b
```

In other words, it does exactly the thing we need if people don't want to write the backprop for some function. The JIT compilation should be very cool too, because it fits quite well into how we do things.

Unlike Tensorflow and PyTorch, Jax support will be as a backend -- so we'll have a `JaxOps` class. We want to keep Jax as an optional dependency, so we'll need to do the conditional imports etc. I think we probably need a `thinc.compat` module to manage those, as it's getting a bit out of control.

The `JaxOps` module should specialise various `Ops` methods, and call into methods decorated with `@jax.jit`. The `jax.jit` function also takes a second argument, `static_argnums`. I'm not 100% sure of the trade-offs on this but I think it means you get different versions of the function compiled for the different branches of that value, which is nice for stuff like small constants. For instance, it seems good for stuff like window size, padding size, axis args, dtype, etc.

`@jax.jit` functions can only take certain types of arguments, because it needs to be able to understand the object (? not sure of the inner workings). You can register "flatten" functions that let you express some object in terms of a flattened list, sort of like a saner pickle. I've created flatten functions for the `Model` class. Using it requires that the model instance has an attribute `registry_name`, which should be the name of a layers registry function that accepts no arguments, and can be used to create a shell of the model. The flattened data is then used to fill in all of the dims, params, attrs etc.

Another trap is that Jax doesn't let you set sub-arrays directly, e.g. you can't do `arr[i] = something`, although you *can* do `arr += something`. I don't really get what's up with that. Anyway, you have to write `arr = jax.ops.index_update(arr, jax.ops.index[i], something`. It's a completely mechanical transformation: whatever you would put in the index, you move that over to `jax.ops.index`. The first arg will always be the whole array, and you'll always be assigning to the whole array as the return value. If you're doing something like `X[:, n] += Y[:, n]` that would become `X = index_update(X, index[:, n], Y[:, n] + X[:, n]`.

It's possible that with JaxOps we could even be faster than PyTorch or Tensorflow for some networks, without our code actually looking any different. It's extremely hard to predict though -- we just have to try it and see.

* [x] Grok Jax
* [x] Basic `JaxOps`
* [x] Allow `jax.jit` to take `Model` instances
* [x] Make some GPU benchmarks of JaxOps vs NumpyOps with JIT
* [x] Benchmark a small network using JaxOps
* [x] Get ops tests passing for JaxOps
* [x] JIT more ops
* [ ] Benchmark more realistic networks


svlandeg:
> Autodiff as a function transformation

This will be awesome to have! Thanks for the walk-through, excited to see how this will come together :-)

honnibal:
I've got most things working with the `JaxOps`, but it's *really* slow. I don't know whether I'm doing things deeply wrong, or whether I'm just running examples that are too small.

I rebroke everything on the branch trying to get the performance better. I'll refix it later to get it into a state to merge.

I think we should merge this in advance of it actually being useful/performant. This will make it easier for other people to tinker with it later.

I suspect it will be difficult to have code that works well with Jax and with cupy, which is really disappointing --- I'd hoped for a more drop-in experience. Again, I'm not 100% sure: I could just be doing this wrong.

PR #231, opened 2020-01-15
Merge pull request #231 from justindujardin/mathy-thinc

justindujardin:
This adds some experimental changes to support de/serializing subclassed Keras models by decorating `tf.keras.Model` subclasses with enough information to recreate them reliably.

Here's a minimal example:

```python
import tensorflow as tf
import numpy
from thinc.api import keras_subclass

@keras_subclass(
    "TestModel",
    X=numpy.array([0.0, 0.0]),
    Y=numpy.array([0.5]),
    input_shape=(2,)
)
class TestModel(tf.keras.Model):
    def call(self, inputs):
        return inputs

```

_This decorator is not used for Tensorflow Functional/Sequential models which serialize with no special configuration._

- [x] add `keras_subclass` class decorator
- [x] update TensorFlowWrapper and TensorFlowShim to support decorated classes 
- [x] add tests
- [ ] Mathy integration
    - [x] test prediction with mathy models
    - [x] test de/serialization with mathy models
    - [ ] ~~test training with mathy models~~
    - [ ] ~~test restoring optimizer with mathy models~~


thinc/config.py
```diff
@@ -142,6 +142,7 @@ class registry(object):
     layers: Decorator = catalogue.create("thinc", "layers", entry_points=True)
     losses: Decorator = catalogue.create("thinc", "losses", entry_points=True)
     initializers: Decorator = catalogue.create("thinc", "initializers", entry_points=True)
+    keras: Decorator = catalogue.create("thinc", "keras", entry_points=True)
```
ines: This should probably live in the wrapper module file and not in the global `registry`, since it's specific to the wrapper.

thinc/layers/tensorflowwrapper.py
```diff
@@ -1,10 +1,19 @@
...
+import numpy as np
+
+from ..config import registry
 from ..model import Model
 from ..shims import TensorFlowShim
-from ..util import xp2tensorflow, tensorflow2xp, assert_tensorflow_installed
-from ..util import is_tensorflow_array, convert_recursive, is_xp_array
-from ..types import ArgsKwargs
+from ..types import ArgsKwargs, Array
+from ..util import (
```
ines: Nitpick, but we try to avoid this by splitting the imports over multiple lines and repeating the `from` etc. It's not ideal either but it's less verbose.

thinc/layers/tensorflowwrapper.py
```diff
@@ -1,10 +1,19 @@
-from typing import Any, Callable, Optional, Tuple, Type, TypeVar
+from typing import Any, Callable, Dict, List, Optional, Tuple, Type, TypeVar
 
+import numpy as np
```
ines: Also nitpick, but I think we don't use the `np` shortcut in the code base and just use `numpy` as `numpy`.

thinc/shims/tensorflow.py
```diff
@@ -1,9 +1,19 @@
-from typing import List, Optional, Dict
 import contextlib
+import copy
 import itertools
+import os
+import shutil
+import tempfile
 from io import BytesIO
+from typing import Any, Callable, Dict, List, Optional, TypeVar
```
ines: Worst nitpick, but I always try to make `from typing` the first import.

Actually, speaking of imports: maybe we should adpot [`isort`](https://github.com/timothycrosley/isort) (just realised this is by the author of Hug btw) and have the import sorting taken care of automatically? I found this config which is apparently compatible with `black`: https://github.com/psf/black/issues/333#issuecomment-516088368

thinc/tests/layers/test_tensorflow_wrapper.py
```diff
@@ -1,12 +1,24 @@
+from typing import Any, Callable, List, Optional, TypeVar
+
 import numpy
 import pytest
-
-from thinc.api import TensorFlowWrapper, tensorflow2xp, xp2tensorflow, Linear
-from thinc.api import Ops, get_current_ops, Model, Adam, ArgsKwargs
+from thinc.api import (
```
ines: Same here, I think it's better to not exceed the line length

thinc/layers/tensorflowwrapper.py
```diff
@@ -32,18 +86,40 @@ def TensorFlowWrapper(
...
+        tensorflow_model.to_json()
+    except NotImplementedError:
+        is_subclass = True
+
+    if is_subclass:
+        for prop_name in ["catalogue_name", "eg_x", "eg_y", "eg_shape"]:
+            if not hasattr(tensorflow_model, prop_name):
+                raise ValueError(
+                    "Keras subclassed models are not whole-model serializable by "
+                    "Tensorflow. To work around this, you must decorate your keras "
```
ines: Ok, I take that "worst nitpick" comment back, this one is it 😂

```suggestion
                    "TensorFlow. To work around this, you must decorate your keras "
```

thinc/layers/tensorflowwrapper.py
justindujardin: Sorry, I use the "organize imports" functionality in VSCode and then run black on top of it. I'll check out isort

thinc/layers/tensorflowwrapper.py
justindujardin: Ahh right! I think I was got used to capitalizing it for the wrapper class names. Should I rename those?

ines: I think the class/function names are all fine? We try to use spelling consistent with the "official" spelling, so `TensorFlow` is correct.

justindujardin: Oh, I had it wrong! 😅

thinc/layers/tensorflowwrapper.py
ines: Just tested isort and it's pretty cool, but there's no way to make it work without wrapping imports like this 😞 There's an open issue on the tracker about a mode that preserves multiple lines importing from the same source (which black leaves untouched, too), but it's not supported yet. So I'm not sure if we want to adpot it... I ran it over the whole code base and it does make things a lot more verbose.

justindujardin: I'll keep an eye out for import wrapping and break the lines manually for now

thinc/layers/tensorflowwrapper.py
justindujardin: I can't accept your suggestion because I don't have write privs to thinc, so I applied your suggestion manually

thinc/layers/tensorflowwrapper.py
justindujardin: I'll keep an eye out for import wrapping and break the lines manually for now

honnibal:
I quite like this! I think the decorator really makes the best of the situation, I think it'll be very usable.

ines:
@justindujardin Is this ready to merge? 🙂

justindujardin:
Sure, I have some more changes, but we can follow up with another PR later

PR #233, opened 2020-01-16
Have concatenate support lists

svlandeg:
Add in support for concatenating lists in the `concatenate` layer.

This works, as tested with spaCy. Before we had to do 
`concatenate_lists(CharacterEmbed(....), FeatureExtracter(...)`
now just `concatenate` works (without messing up the normal `concatenate` cases)


honnibal:
It's a bit of a pain, but I think it'll be best if we follow the style from the `with_array`, with_ragged` etc layers and have different forward functions for the different data-types. Otherwise the conditional logic gets really messy.

Could you change it over to that style?

svlandeg:
Both ways kind of suck from a code-maintenance perspective, but your suggestion is probably more readable :-)

PR #234, opened 2020-01-16
Update examples, add notebooks

ines:
- [x] add installation instructions for dependencies etc. to the top of all example scripts
- [x] create `requirements.txt` with dependencies for all examples
- [x] remove example-only dependencies (`typer`, `tqdm`) from setup
- [x] remove `howto` directory and add notebooks (see below)
- [x] add `datasets` registry (came up in example and seems like something users would want)

## New content

I want to create more notebooks, since they make it really easy to illustrate the concepts and let users try things live. We can then add them to the README with "Launch on Colab" buttons. This also works for the GPU examples, which is pretty nice.

- [x] "Thinc for beginners" notebook that walks through MNIST example and explains config system with runnable examples
- [x] Intro to `Model` class and methods: this is mostly a merged and notebook-ified version of the 2 howto scripts, still needs some more text and also has some errors at the moment
- [x] Transformers tagger example


examples/basic_pos_tagger.py
```diff
@@ -24,8 +69,38 @@ def main(
             >> ReLu(width, width)
             >> Softmax(17, width)
         )
```
svlandeg: I guess it's a matter of personal preference, but I find it more easily understandable if you add `nO=` and `nI=` bits to the layer definitions. Once you know what these abbreviations mean, I find it helps understanding the flow better.

examples/notebooks/intro_model_definition_methods.ipynb
```diff
@@ -0,0 +1,477 @@
...
+    "assert model.get_dim(\"nI\") == 16\n",
+    "assert model.get_dim(\"nO\") == 10\n",
+    "assert model.layers[0].get_dim(\"nO\") == n_hidden"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We call functions like `chain` **combinators**. Combinators one or more models as arguments, and return another model instance, without introducing any new weight parameters. Another useful combinator is `concatenate`:"
```
svlandeg: missing word "take" in second sentence

examples/notebooks/intro_model_definition_methods.ipynb
```diff
@@ -0,0 +1,477 @@
+{
```
svlandeg: I love this idea of adding intro notebooks!

examples/notebooks/intro_model_definition_methods.ipynb
```diff
@@ -0,0 +1,477 @@
...
+   "source": [
+    "dX = backprop(dY)\n",
+    "dX"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The `backprop() `callback only increments the parameter gradients, it doesn't actually change the weights. To increment the weights, call `model.finish_update()`, passing it an optimizer:"
```
svlandeg: there's a space in the backticks (nitpicking!)

examples/notebooks/transformers_tagger.ipynb
```diff
@@ -0,0 +1,454 @@
...
+    "        return_tensors=\"pt\",\n",
+    "    )\n",
+    "    return TokensPlus(**token_data), lambda d_tokens: d_tokens"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The forward pass takes the model and a list of lists of strings and outputs the `TokensPlus` dataclass and a callback to use during the backwards (which does nothing in this case)."
```
svlandeg: duplicate "list of list of"

examples/basic_pos_tagger.py
ines: Ah, so you mean `Softmax(nO=17, nI=width)`? If we do this, I think we should adopt this consistently across the examples. @honnibal What do you think?

examples/notebooks/intro_model_definition_methods.ipynb
ines: Yeah, I think the notebook format actually works especially well for this kind of how-to style!

svlandeg: Agreed!

examples/notebooks/intro_model_definition_methods.ipynb
ines: Haha omg yes, that's exactly my level of nitpick ❤️😅

examples/notebooks/transformers_tagger.ipynb
ines: I think it is actually a list of lists, i.e. `List[List[str]`, but it's kinda awkward to get across 🤔

examples/basic_pos_tagger.py
honnibal: Yeah it's probably a bit important: our layers have the args swapped compared to PyTorch, they go `(input_size, output_size)`. We don't want to do that because we more often allow the input size to be omitted, but it does mean the strictly positional args are probably not great in initial examples :(.

examples/notebooks/transformers_tagger.ipynb
honnibal: Maybe `list-of-lists of strings`?

PR #245, opened 2020-01-20
Fixing ops

svlandeg:
Trying to get spaCy's `train_from_config` working with GPU. Found some cases in `initializers.py` that were using `numpy` instead of `ops.xp`.

Also note that `ops.xp.prod(shape)` does not work for `cupy`, you have to make sure the argument is an array first.


svlandeg:
Update: added `to_numpy` from the `feature/jax2` branch. Should be an easy-to-resolve conflict because I copy-pasted.

PR #247, opened 2020-01-20
Make mypy tests run without requiring pip install -e ./

tiangolo:
:bug: Make `mypy` tests run without requiring `pip install -e ./`


tiangolo:
Superseded by #293

PR #249, opened 2020-01-20
restore some edits from concatenate

svlandeg:
In spaCy, `test_tok2vec_configs` was failing on calling `backprop(vectors)` because of the recent changes to `concatenate`. I reverted part of the edits from [here](https://github.com/explosion/thinc/commit/4d32ac32e165f425fb30c69281c4dd0dda142c72#diff-5be155357e784dde8a496aa6aad6771fL74), and that seems to fix it. I'm not sure why this got removed in the first place though.


honnibal:
Ah, sorry! I had trouble merging the concatenate changes, because Jax was fussy about it too.

PR #251, opened 2020-01-21
Fix and test losses

svlandeg:
* Removed `L1Distance` which wasn't being used and had kind of a complex signature
* Created simple `L2Distance` for use in spacy's [`pretrain`](https://github.com/svlandeg/spaCy/blob/feature/config/spacy/cli/pretrain.py#L256)
* Implemented all `get_loss` functions
* `CategoricalCrossentropy` is kind of similar to `L2Distance`, though the latter can work outside the [0,1] interval and the former can't (on purpose). I thought this might be useful for debugging.
* Some usages of these functions assumed normalization, i.e. division by the number of examples given. For instance `CategoricalCrossentropy` was dividing by `guesses.shape[0]` while for `CosineDistance` this was done by the calling code. To make this more homogeneous I added a parameter `normalize` to all functions which is `True` by default.
* I added tests that ensure the calculations are correct for all these functions.


thinc/loss.py
```diff
@@ -32,92 +32,131 @@ def get_loss(self, guesses: GuessT, truths: TruthT) -> LossT:
...
+        self.cc = CategoricalCrossentropy(normalize=normalize)
 
     def __call__(
         self, guesses: List[Array2d], truths: List[Array]
     ) -> Tuple[List[Array2d], List[float]]:
         return self.get_grad(guesses, truths), self.get_loss(guesses, truths)
 
     def get_grad(self, guesses: List[Array2d], truths: List[Array]) -> List[Array2d]:
-        if not guesses:
+        if guesses is None:
```
honnibal: I think we don't have to handle this sort of thing: our stance in Thinc is that the function behaves respectably if you give it respectable types. So we only need conditions for non-type invariants.

honnibal:
Looks good!

PR #252, opened 2020-01-21
Fix errors using TensorFlowWrapper for multithreaded training

justindujardin:
- add `get_thread_state` helper for backends. When called it will get the thread local for the current thread, or create it from the global state attributes.
 - After some searching I found a blog post that talked about a subtle detail that meant you cannot do mutable thread local data without subclassing threading.local. Update model operators thread local to use a subclass. This prevents errors in the overloaded operators test.

Source: http://slinkp.com/python-thread-locals-20171201.html

### DIFF
--- a/thinc/model.py
+++ b/thinc/model.py
@@ -34,6 +34,7 @@ class Model(Generic[InT, OutT]):
     """Class for implementing Thinc models and layers."""
 
     global_id: int = 0
+    global_id_lock: threading.Lock = threading.Lock()
     _thread_local = create_thread_local({"operators": {}}, ModelThreadState)
 
     name: str
@@ -95,7 +96,8 @@ class Model(Generic[InT, OutT]):
         self._shims = list(shims)
         # Take care to increment the base class here! It needs to be unique
         # across all models.
-        Model.global_id += 1
+        with Model.global_id_lock:
+            Model.global_id += 1
         self.id = Model.global_id
         self._has_params = {}
         for name, value in params.items():
--- a/thinc/model.py
+++ b/thinc/model.py
@@ -5,6 +5,7 @@ import srsly
 from pathlib import Path
 import copy
 import functools
+import threading
 
 from .backends import ParamServer, Ops, NumpyOps, CupyOps, get_current_ops
 from .optimizers import Optimizer  # noqa: F401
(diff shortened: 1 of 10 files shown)


thinc/model.py
```diff
@@ -22,11 +23,18 @@ def empty_init(model: "Model", *args, **kwargs) -> "Model":
...
+    def __init__(self):
+        self.operators = {}
+
+
 class Model(Generic[InT, OutT]):
     """Class for implementing Thinc models and layers."""
 
     global_id: int = 0
-    _thread_local = create_thread_local({"operators": {}})
+    _thread_local = create_thread_local({"operators": {}}, ModelThreadState)
```
honnibal: Do we have to attach this to the `Model` class? We had it there before but it's always added to the vague this-code-is-haunted vibe I had from the threading stuff.

thinc/model.py
```diff
@@ -22,11 +23,18 @@ def empty_init(model: "Model", *args, **kwargs) -> "Model":
...
+    operators: Dict[str, Callable[["Model", "Model"], "Model"]]
+
+    def __init__(self):
+        self.operators = {}
+
+
 class Model(Generic[InT, OutT]):
     """Class for implementing Thinc models and layers."""
 
     global_id: int = 0
```
honnibal: This ID is maybe a problem...We want this to be *properly* global, it will cause problems if models come back from other threads and IDs aren't unique. Should we use a lock here?

thinc/model.py
justindujardin: We could do it like I did for the backend file with a `get_model_state` helper function that attaches the model local to the thread object itself. It's a little boilerplate code but I think it makes the whole thing less magic because you get a thread local once, and store a reference to it on the thread it's associated with.

thinc/model.py
justindujardin: Sure, I'll add a threading lock and acquire it when incrementing the global value

justindujardin:
I'm getting inconsistent results with this training mathy. Don't merge it yet, I want to do some more testing.

justindujardin:
Okay, I was able to resolve the inconsistencies by excluding the TensorFlow graph reset (i.e. `tf.keras.backend.clear_session()`) when performing `from_bytes` on a subclassed keras model that is only receiving new weights.

PR #255, opened 2020-01-22
Mypy plugin: chain, add, concatenate

tiangolo:
:sparkles: Update Mypy plugin to correctly handle subtypes with `chain`, add specific support for `add` and `concatenate`.


thinc/types.py
```diff
@@ -26,7 +26,8 @@
 OpsNames = Literal["numpy", "cupy", "jax"]
 DeviceTypes = Literal["cpu", "gpu", "tpu"]
 ArrayT = TypeVar("ArrayT", bound="Array")
-Reduced_OutT = TypeVar("Reduced_OutT")
+Chained_OutT = TypeVar("Chained_OutT")
+InToIn_OutToOut_OutT = TypeVar("InToIn_OutToOut_OutT")
```
tiangolo: I don't like this name `InToIn_OutToOut_OutT`, but it's the most expressive I could think of.

Combinators that have a return of `Model[Something, InToIn_OutToOut_OutT]` will be checked by the mypy plugin. Each layer's input (argument) of the combinator will be checked/compared agains the next layer's input. And each layer's output to the next layer's output.

This is used for `add` and `concatenate`.

honnibal: It does make sense. An alternative could be `X2X_Y2Y_OutT`? Exploiting the x is in, y is out convention.

honnibal: Or it could be `XY_XY_OutT`? And chain would be `XY_YZ_OutT`

tiangolo: I like the `X2X_Y2Y_OutT`! Looks a lot simpler. :tada: 

---

I think `XY_YZ_OutT` for chain might be a bit less clear, as the first `Y` doesn't really need to be the same type as the second `Y`, it just needs to be a subtype. Although `Chained` doesn't convey the meaning any better :man_shrugging: 

---

Also, as we are using these types to trigger the plugin, users could actually use them to make mypy do the same checks. I don't think it would be super common, but still, they could end up interacting with these `TypeVar`s.

Anyway, you have the final word :crown:, let me know what you prefer and I'll update it.

honnibal: I think `XY_XY_OutT` might be better than `X2X_Y2Y_OutT`. I agree the partial-match is maybe misleading, but I do think it's close enough, so I'd rather try to get a convention that generalizes a little. Like, we could also have a layer that went `XY_XZ`, if the inputs are the same type but the layers have different outputs. For instance, let's say a combinator that passes the same input to a bunch of layers, and makes a tuple from their output. That would be `XY_XZ` I think.

tiangolo: Get it. Yep, makes sense. :heavy_check_mark:

PR #260, opened 2020-01-23
Fix model.ops type and related issues

ines:
`Model.ops` was incorrectly typed as `Union[NumpyOps, CupyOps]`, which meant that a variety of type issues went unnoticed.

## TODO

- [ ] fix maxout

### DIFF
--- a/thinc/layers/maxout.py
+++ b/thinc/layers/maxout.py
@@ -3,7 +3,7 @@ from typing import Tuple, Callable, Optional, cast
 from ..model import Model
 from ..config import registry
 from ..initializers import glorot_uniform_init, zero_init
-from ..types import Array2d
+from ..types import Array2d, Array3d
 from ..util import get_width, partial
 from .dropout import Dropout
 from .layernorm import LayerNorm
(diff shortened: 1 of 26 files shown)


thinc/layers/maxout.py
```diff
@@ -44,7 +44,7 @@ def forward(model: Model[InT, OutT], X: InT, is_train: bool) -> Tuple[OutT, Call
     nP = model.get_dim("nP")
     nI = model.get_dim("nI")
     b = model.get_param("b")
-    W = model.get_param("W")
+    W = cast(Array2d, model.get_param("W"))
```
honnibal:
```suggestion
    W = model.get_param("W")
```

thinc/layers/maxout.py
```diff
@@ -44,7 +44,7 @@ def forward(model: Model[InT, OutT], X: InT, is_train: bool) -> Tuple[OutT, Call
...
     W = W.reshape((nO * nP, nI))
```
honnibal:
```suggestion
    W = cast(Array3d, W.reshape((nO * nP, nI)))
```

PR #263, opened 2020-01-24
Replace thread locals with contextvars

tiangolo:
:arrow_up: Replace thread locals with contextvars.

This would allow using Thinc in scenarios that are not purely thread-based, like `asyncio` (e.g. in FastAPI apps). While still working the same for thread-based scenarios.

From PEP 567: https://www.python.org/dev/peps/pep-0567/#backwards-compatibility

> This proposal preserves 100% backwards compatibility.
> 
> Libraries that use threading.local() to store context-related values, currently work correctly only for synchronous code. Switching them to use the proposed API will keep their behavior for synchronous code unmodified, but will automatically enable support for asynchronous code.

---

Python docs: https://docs.python.org/3/library/contextvars.html
PEP 567 "Converting code that uses threading.local()": https://www.python.org/dev/peps/pep-0567/#converting-code-that-uses-threading-local

### DIFF
--- a/thinc/model.py
+++ b/thinc/model.py
@@ -1,6 +1,7 @@
 from typing import Dict, List, Callable, Optional, Any, Union, Iterable, Set
 from typing import Generic, Sequence, Tuple, TypeVar
 import contextlib
+from contextvars import ContextVar
 import srsly
 from pathlib import Path
 import copy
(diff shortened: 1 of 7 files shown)


thinc/model.py
```diff
@@ -35,7 +37,7 @@ class Model(Generic[InT, OutT]):
 
     global_id: int = 0
     global_id_lock: threading.Lock = threading.Lock()
-    _thread_local = create_thread_local({"operators": {}}, ModelThreadState)
```
justindujardin: I had to add `ModelThreadState` as a workaround for thread-locals not being mutable. If that's not required here can you remove that class?

PR #149, opened 2019-12-30
Fix murmur hash for GPU/Windows

svlandeg:
Fixes #128 : [`test_hash_gives_distinct_keys[CupyOps]`](https://github.com/explosion/thinc/blob/master/thinc/tests/unit/test_ops.py#L30) stops failing

Fixes https://github.com/explosion/spaCy/issues/4758: no more weird POS & dep tags

Probably fixes https://github.com/explosion/spaCy/issues/4816 (should be tested after merge)

## Description

The murmur hash calculation for Windows (when using GPU/CUDA) was not working properly.

I finally figured out why: `int64_t` was being defined as `long int`. This is correct on LINUX, but not on [Windows](https://docs.microsoft.com/en-us/cpp/c-runtime-library/standard-types?redirectedfrom=MSDN&view=vs-2019).

So I added a bit of code to do some byte-size checking upfront and only then define the correct types. It's a little ugly, but I couldn't get this implemented in a more clean fashion because
- I couldn't find a way to check for platform (`#if defined(_MSC_VER)` e.g. didn't work)
- I couldn't find a way to include the platform-specific header files (which is what you see in most implementations of the murmur hash because of this type incompatibility)
- You're not allowed to redefine a `typedef`

Nonetheless, I think this works :-)

## Example output

This has been the source of many quirks reported with GPU & Windows. I ran some tests myself and this was the output:

#### en_core_web_sm

```
tok.text ['The', 'decrease', 'in', '2008', 'primarily', 'relates', 'to', 'the', 'decrease', 'in', 'cash', 'and', 'cash', 'equivalents', '1', '.', '\n']
tok.pos_ ['PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'PUNCT', 'SPACE']
tok.dep_ ['ROOT', 'pobj', 'neg', 'neg', 'neg', 'neg', 'neg', 'neg', 'neg', 'neg', 'neg', 'neg', 'neg', 'neg', 'neg', 'neg', '']
```

#### en_core_web_md

```
tok.text ['The', 'decrease', 'in', '2008', 'primarily', 'relates', 'to', 'the', 'decrease', 'in', 'cash', 'and', 'cash', 'equivalents', '1', '.', '\n']
[...]

PR #265, opened 2020-01-24
Running MNIST example with GPU

svlandeg:
I was trying to run the MNIST example from the 0'th notebook and ran into trouble on the GPU. I added a unit test that has `prefer_gpu()` - I'm not sure we want that, but at least for now it helps us diagnose things. I fixed a few errors trying to run this, until I got completely stuck at:

`python -m pytest model\test_model.py`

```
model\test_model.py:385:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
..\model.py:323: in finish_update
    param, grad = optimizer((node.id, name), param, grad)
..\optimizers.py:248: in __call__
    weights, gradient = self._adam(xp, weights, gradient, lr_scale, key, nr_upd)
..\optimizers.py:351: in _adam
    weights_1D, gradient_1D, mom1, mom2, b1, b2, eps, lr * lr_scale
..\backends\cupy_ops.py:145: in adam
    )(gradient, learn_rate, 1 - beta1, 1 - beta2, eps, weights, mom1, mom2)
cupy\core\_kernel.pyx:591: in cupy.core._kernel.ElementwiseKernel.__call__
    ???
cupy\core\_kernel.pyx:371: in cupy.core._kernel._broadcast
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   ValueError: operands could not be broadcast together with shapes (10, 320), (), (), (), (), (10, 320), (10, 32), (10, 32)

cupy\core\_routines_manipulation.pyx:73: ValueError
```

@honnibal @ines Any pointers on where to look to address this ? `cupy_ops.adam` looks scary.

### DIFF
--- a/thinc/backends/cupy_ops.py
+++ b/thinc/backends/cupy_ops.py
@@ -38,8 +38,7 @@ class CupyOps(Ops):
     def gemm(self, x, y, out=None, trans1=False, trans2=False):
         if type(x) == numpy.ndarray or type(y) == numpy.ndarray:
             raise ValueError("Encountered a numpy array when processing with cupy. "
-                             "This is likely an error in thinc, so feel free to open "
-                             "an issue on the GitHub tracker.")
+                             "Did you call model.ops.asarray on your data?")
         if trans1:
             x = x.T
         if trans2:
--- a/thinc/backends/cupy_ops.py
+++ b/thinc/backends/cupy_ops.py
@@ -36,6 +36,10 @@ class CupyOps(Ops):
             return data.get()
 
     def gemm(self, x, y, out=None, trans1=False, trans2=False):
+        if type(x) == numpy.ndarray or type(y) == numpy.ndarray:
+            raise ValueError("Encountered a numpy array when processing with cupy. "
+                             "This is likely an error in thinc, so feel free to open "
+                             "an issue on the GitHub tracker.")
         if trans1:
             x = x.T
         if trans2:
--- a/thinc/backends/cupy_ops.py
+++ b/thinc/backends/cupy_ops.py
@@ -36,7 +36,7 @@ class CupyOps(Ops):
             return data.get()
 
     def gemm(self, x, y, out=None, trans1=False, trans2=False):
-        if type(x) == numpy.ndarray or type(y) == numpy.ndarray:
+        if isinstance(x, numpy.ndarray) or isinstance(y, numpy.ndarray):
             raise ValueError("Encountered a numpy array when processing with cupy. "
                              "Did you call model.ops.asarray on your data?")
         if trans1:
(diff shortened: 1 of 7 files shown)


thinc/backends/cupy_ops.py
```diff
@@ -36,6 +36,10 @@ def to_numpy(self, data):
             return data.get()
 
     def gemm(self, x, y, out=None, trans1=False, trans2=False):
+        if type(x) == numpy.ndarray or type(y) == numpy.ndarray:
```
ines:
```suggestion
        if isinstance(x, numpy.ndarray) or isinstance(y, numpy.ndarray):
```

thinc/backends/cupy_ops.py
```diff
@@ -36,6 +36,10 @@ def to_numpy(self, data):
...
+            raise ValueError("Encountered a numpy array when processing with cupy. "
```
ines: We should probably rephrase this error, because there's nothing that keeps the user from calling `CupyOps.gemm` with a numpy array. If it happens in the internals, it's likely our fault – but if the user does it, we wouldn't consider that a bug and wouldn't want them to open an issue about it 😛

svlandeg: Better?

honnibal:
I think the error's probably associated with the array shaping. I'll take a look

honnibal:
Yes in `Optimizer` we're setting `mom1` and `mom2` to the original weights shape.

The problem here is that we actually want to avoid this reshaping if possible, and do it a bit later. We should change it back to not reshaping in the `optimizer.py`, and instead only do it within the `NumpyOps.adam` method.

The `JaxOps` backend needs to make a copy to reshape, so I was trying to avoid unnecessary reshapings. It's also a bit of an unintuitive way to structure the code, as it doesn't make that much sense in itself.

svlandeg:
UPDATE: also ported the fix from https://github.com/explosion/thinc/pull/149 so that I can run the unit tests locally on GPU

PR #281, opened 2020-01-27
Remove type from fwd in FeatureExtractor

svlandeg:
I'm not sure this is the "fix" we want, but the current `docs` type in `FeatureExtractor.forward()` does not match with the `begin_training(X=...)` arguments we're giving in spaCy, making the unit tests fail if we type it as `List[Doc]`. 

I don't think I know the historic trail here, and why we're also having this check of `if hasattr(doc, "to_array")` in this layer. Could be that the type should be as currently stated, and we need to refactor elsewhere? I'm not sure.

### DIFF
--- a/thinc/layers/featureextractor.py
+++ b/thinc/layers/featureextractor.py
@@ -15,7 +15,7 @@ def FeatureExtractor(columns: List[Union[int, str]]) -> Model[InT, OutT]:
 
 
 def forward(
-    model: Model[InT, OutT], docs: InT, is_train: bool
+    model: Model[InT, OutT], docs, is_train: bool
 ) -> Tuple[OutT, Callable]:
     columns = model.attrs["columns"]
     features: OutT = []


honnibal:
Hm, fair enough. What's the data type we're sending in? Is it a generator?

Anyway, sure, we can remove it for now.

svlandeg:
That's the weird part, we are actually passing a `List` of `Doc`'s: https://github.com/explosion/spaCy/blob/feature/config/spacy/pipeline/pipes.pyx#L1373

```
docs = [Doc(Vocab(), words=["hello"])]
self.model.initialize(X=docs)
```
The tests crash at `validate_fwd_input_output` called by `Model.initialize()`:

```
    def validate_fwd_input_output(
        name: str, func: Callable[[Any, Any, bool], Any], X: Any, Y: Any
    ) -> None:
        """Validate the input and output of a forward function against the type
        annotations, if available. Used in Model.initialize with the input and
        output samples as they pass through the network.
        """
        sig = inspect.signature(func)
        empty = inspect.Signature.empty
        params = list(sig.parameters.values())
        if len(params) != 3:
            bad_params = f"{len(params)} ({', '.join([p.name for p in params])})"
            err = f"Invalid forward function. Expected 3 arguments (model, X , is_train), got {bad_params}"
            raise DataValidationError(name, X, Y, [{"msg": err}])
        annot_x = params[1].annotation
        annot_y = sig.return_annotation
        sig_args: Dict[str, Any] = {"__config__": _ArgModelConfig}
        args = {}
        if X is not None and annot_x != empty:
            sig_args["X"] = (annot_x, ...)
            args["X"] = X
        if Y is not None and annot_y != empty:
            sig_args["Y"] = (annot_y, ...)
            args["Y"] = (Y, lambda x: x)
        ArgModel = create_model("ArgModel", **sig_args)
        try:
>           ArgModel.parse_obj(args)

..\thinc\thinc\util.py:367: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cls = <class 'ArgModel'>, obj = {'X': [hello ]}

    @classmethod
    def parse_obj(cls: Type['Model'], obj: Any) -> 'Model':
        if cls.__custom_root_type__ and (
            not (isinstance(obj, dict) and obj.keys() == {ROOT_KEY}) or cls.__fields__[ROOT_KEY].shape == SHAPE_MAPPING
        ):
            obj = {ROOT_KEY: obj}
        elif not isinstance(obj, dict):
            try:
                obj = dict(obj)
            except (TypeError, ValueError) as e:
                exc = TypeError(f'{cls.__name__} expected dict not {type(obj).__name__}')
                raise ValidationError([ErrorWrapper(exc, loc=ROOT_KEY)], cls) from e
>       return cls(**obj)

..\..\..\Anaconda3\envs\myexplosion\lib\site-packages\pydantic\main.py:402: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

__pydantic_self__ = ArgModel(), data = {'X': [hello ]}, values = {}
fields_set = {'X'}
validation_error = ValidationError(model='ArgModel', errors=[{'loc': ('X', 0), 'msg': 'instance of Doc expected', 'type': 'type_error.arbitrary_type', 'ctx': {'expected_arbitrary_type': 'Doc'}}])

    def __init__(__pydantic_self__, **data: Any) -> None:
        # Uses something other than `self` the first arg to allow "self" as a settable attribute
        if TYPE_CHECKING:
            __pydantic_self__.__dict__: Dict[str, Any] = {}
            __pydantic_self__.__fields_set__: 'SetStr' = set()
        values, fields_set, validation_error = validate_model(__pydantic_self__.__class__, data)
        if validation_error:
>           raise validation_error
E           pydantic.error_wrappers.ValidationError: 1 validation error for ArgModel
E           X -> 0
E             instance of Doc expected (type=type_error.arbitrary_type; expected_arbitrary_type=Doc)

..\..\..\Anaconda3\envs\myexplosion\lib\site-packages\pydantic\main.py:283: ValidationError
```

PR #298, opened 2020-01-30
fix issue where GPU model couldn't be serialized

justindujardin:
Oops, we were converting the model dict to numpy, then passing `self.to_dict()` to `convert_recursive` instead of the converted numpy array 😅


justindujardin:
Possibly related to: https://github.com/explosion/spacy-transformers/issues/99

PR #300, opened 2020-02-03
Rename Unserializable 'data' field

svlandeg:
This was a cute little bug:
* `Unserializable` has a `data` field
* `StaticVectors` expects a `vectors` argument (e.g. spaCy's `Vectors` object) that has a `data` field
* So you'd need to do `model.attrs["vectors"].data.data` to get to the actual data, but I renamed it instead to avoid future confusion

### DIFF
--- a/thinc/layers/concatenate.py
+++ b/thinc/layers/concatenate.py
@@ -36,7 +36,7 @@ def concatenate(*layers: Model) -> Model[InT, XY_XY_OutT]:
 
 
 def forward(model: Model[InT, OutT], X: InT, is_train: bool) -> Tuple[OutT, Callable]:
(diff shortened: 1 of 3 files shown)


svlandeg:
Looking at the failing tests, I guess the original assumption was that the `vectors` object is directly compatible with `Floats2d`, so from spaCy we should create this layer with `vectors=vectors.data` instead. Updated the PR accordingly. Still would like to keep the renamed field in `Unserializable` which caused this confusion :-)

honnibal:
Agree that `data` is poor. Maybe `obj` instead of `wrapped_object` though? The codebase normally prefers shorter names, so `wrapped_object` is a bit inconsistent.

PR #303, opened 2020-02-04
Fixing StaticVectors

svlandeg:
`StaticVectors` didn't work for me, it got the wrong output dimensions. I shuffled the code around, more or less as it used to be. This seems to at least fix the training step.

A review would be good because I'm not entirely confident.

Also, there's still quite a bit to do on the current spaCy `develop` branch with respect to `StaticVectors`, `cfg` passing, IO etc, so I haven't yet been able to test this end-to-end yet, which is why I'm suggesting to have this as WIP (see also https://github.com/svlandeg/spaCy/tree/feature/static-vectors).


svlandeg:
Ok, after some more testing this seems to run OK at least. Ready for review :-)

svlandeg:
Update: also allowed to specify the `init_W` and set default to `glorot_uniform_init`. Oddly, that doesn't seem to make much difference in spaCy's model training.

Issue #301, opened 2020-02-04
Using list2padded for sequences


naveenjafer:
I have referred to https://github.com/explosion/thinc/blob/master/examples/03_pos_tagger_basic_cnn.ipynb to get a better understanding of the thinc layers. The model is as follows in the example  
```
model = strings2arrays() >> with_array(
        HashEmbed(nO=width, nV=vector_width, column=0)
        >> expand_window(window_size=1)
        >> ReLu(nO=width, nI=width * 3)
        >> ReLu(nO=width, nI=width)
        >> Softmax(nO=nr_classes, nI=width)
    )
```
Can someone please explain what a string2arrays does? The documentation says that it takes a sequence of sequence of string and produces a list[int2D].  

The input X_train is something like [["this","is","awesome"],["thinc","is","cool"]]. What does the strings2arrays transform this example to? I am unable to wrap my head around what exactly strings2arrays does and how it transforms the input from a List[List] (2D lists) to List[Int2D] technically (3D lists/sequence).

honnibal:
You're right that this is a bit confusing. I just had a look at the source, and the `strings2arrays` is inserting an extra dimension: https://github.com/explosion/thinc/blob/master/thinc/layers/strings2arrays.py

During development we had the `Embed` and `HashEmbed` layers expecting only 2d inputs, which didn't make that much sense. I coded the `strings2array` accordingly, making it return a 1-column array instead of the more obvious `ints1d`.

I think this should be changed: `strings2arrays` should return only `ints1d`. We'll have to update the examples accordingly.

naveenjafer:
Oh that makes sense! I do have another continuing query. LSTM's need a padded sequence as input. In order to satisfy that, I assume the list2padded() layer should follow the string2array layer( Takes as input List[Array2D])? But when I do that, the execution of the model simply gets stuck without any progress. And since you are planning on reverting strings2array to a list[ints1d] would'nt the list2padded also need a change ( Assuming that these 2 layers need to be chained to feed it to with_array and subsequently the LSTM layer inside it. 

I was working on something like this, but it only gets stuck.
model = strings2arrays() >> list2padded() >> with_array(
        HashEmbed(nO=width, nV=vector_width, column=0)
        >> LSTM(nO=100, nI=width)
        >> Softmax(nO=nr_classes, nI=width)
    )

honnibal:
[...]

PR #304, opened 2020-02-04
Faster CPU BiLSTM

honnibal:
Thinc should have a decent BiLSTM implementation, so we can use it in spaCy. The current one is too slow, so we need to use PyTorch.

For GPU, we can call into CUDNN via cupy, and get fast speeds (although it's not so flexible). It was pretty hard to figure out the API, but I've pretty much worked all of that out (although I'm not ready to push the work on it yet).

The missing piece was a CPU implementation that was compatible with CuDNN (so that you can use a network trained on GPU on CPU). I don't know about full compatibility yet, but the speed is now pretty good.

## Benchmark

I ran a quick benchmark with synthetic data against PyTorch on my laptop (at `examples/benchmark/lstm_tagger.py`). I used three timings each, and picked the best one. Results reported in words per second (higher is better).

|         | Fwd | Fwd+Bwd |
| ----- | --------- | ------ |
| Thinc | 12.9k    | 4.0k |
| PyTorch | 17.6k | 4.1k |

The network size (depth 2, 300d, bidirectional) is about the size we'd be looking to run in spaCy. I set both libraries to single-threaded. I normally get worse performance multi-threading on CPU.

I definitely consider anything better than 50% as fast as PyTorch a victory, especially since this was only a few days work. The backward speed is maybe a bit too good to be true --- possibly I'm making a slight mistake somewhere that could cost accuracy?

## Threading, nogil, and blis

The inference loop is almost nogil, so we *could* make it close to 2x faster by using `cython.parallel.prange` and using 2 threads. This is sort of a pain in the ass though: it means using openmp, which introduces all sorts of cross-platform headaches. Not worth it.

I used `blis.cy.gemm` to get matrix multiplication from Cython. This isn't strictly necessary --- we could use numpy instead. The thing is though, to do the implementation efficiently we have to be working on subarrays using inplace operations. That's really messy with numpy --- it's actually a lot clearer and easier with raw pointers. So I would prefer to just add back the blis dependency. We can keep the option of `use_blis` on `NumpyOps`, but set it to `True`.

## Variations

There are many variations of LSTMs we can implement as well. For instance, layer norm LSTM is apparently popular now. It's nice to have a decent Cython implementation, so that we know we can do these variants relatively efficiently. However, we need to make sure we have an implementation that remains compatible with CuDNN. Other libraries often end up with two implementations, one CuDNN-compatible, the other not. We might end up with that too.

## TODO

* [x] Implement BiLSTM forward pass for `NumpyOps`
* [x] Implement BiLSTM backward pass for `NumpyOps` 
* [x] Check accuracy against PyTorch using tagger script
* [x] Check speed against PyTorch
* [x] Implement dropout for `NumpyOps`
* [ ] Fix `Ops` reference implementation
* [ ] Add CuDNN implementation for `CupyOps`
* [ ] Verify GPU->serialize->CPU works accurately
* [ ] Verify CPU->serialize->GPU works accurately
* [ ] Refactor/tidy up


naveenjafer:
Hi @honnibal any idea as to when this LSTM fix will be merged to master?

naveenjafer:
Hi @honnibal I pulled this branch feature/lstm and merged it with the latest master and tried evaluating for LSTM, but looks like there are some issues with other layers. With the current master, PyTorchLSTM is working as expected, but native LSTM doesn't. Once I merged Feature/LSTM to it, both he LSTMs are not working as expected. I believe this happened a few months back too, but a fix that you had put in for list2padded fixed it

> Pretty sure I've also found a bug in the `list2padded` method. Should have a fix shortly.
> 
> _Originally posted by @honnibal in https://github.com/explosion/thinc/issues/301#issuecomment-582126887_
>

honnibal:
The CuDNN path I went down here is tough, and I need to spend more time to get it working. I'll close this in preference of getting a simpler non-cudnn approach into the library.

Issue #306, opened 2020-02-06
Issue in Embed


naveenjafer:
@honnibal 
I have a model as below using HashEmbed, which trains without any issues.
```
with Model.define_operators({">>": chain}):
    model = (
        strings2arrays() 
        >> list2padded() 
        >> with_array(HashEmbed(nO=width, nV=10, column=0))
        >> LSTM(nO=100, nI=width)        
        >> with_array(Softmax(nO=n_tags+1, nI=100))
        >> padded2list()
    )
```

When I replace HashEmbed with Embed, I run into some error like "IndexError: index 94100425 is out of bounds for axis 0 with size 11". The error is thrown in  
`---> 55     output = vectors[ids]` of /thinc/layers/embed.py  

I tried comparing the code of HashEmbed to Embed to understand what might be causing this issue, but I had no luck. I even updated HashEmbed to calculate output as  
`output = vectors[ids]  `
instead of 
`output = vectors[keys].sum(axis=1)`
to try and reproduce the error in HashEmbed, but it works just the same.

I also have a doubt regarding the nV input. I believe nO is the embedding dimension, but then what is nV? In the case of Embed it defaults to 1 and not in the case of HashEmbed.

naveenjafer:
I think I figured the issue, it seems to be a bug with how Embed is implemented. In the forward method of embed.py modding ids by nV ensures that they are all in range. This is similar to something that is done with the hashembed.py method. I will put in a pull req for this if that's okay.

```
ids = ids % nV
output = vectors[ids]
```

@honnibal And as for my previous question, I guess nV technically should be set to the number of unique words in the vocab?  

I also have a doubt regards this 'ids' array. This is a sample of the numbers in it  
```
[            94100425 18446744072411621407           1558403699
 18446744073360372835            603615019            131900689
 18446744072992786556 18446744072281243677 18446744072757647121
           1299665196 18446744072576802658 18446744073308518867
 18446744073252088270            589538811 18446744072576802658
            979103863 18446744072576802658 18446744071958240053
 18446744071931561673 18446744073272641596           2021799277
 18446744073253355074           1832516214 18446744072608696245
 18446744073259610218 18446744072459575127           1299665196
           1299665196 18446744071958240053 18446744071646439289
            893251777 18446744072576802658           1009084850
``` 
I guess these numbers are created during the strings2array step? But why is the range of numbers so large?. I would have expected the numbers in ids to be <= number of words in my vocab.

honnibal:
[...]

PR #309, opened 2020-02-13
fixed breaking bug in Embed

naveenjafer:
**What was the Bug?**
Embed was breaking in the previous version due to indexes out of bounds error. #306 

**Reproduction**
Replace HashEmbed in Example 3 pos_tagger_basic_cnn.ipynb of thinc documentation with Embed. 

**How was it fixed?**
A modulo operation with the vocabulary size was needed to keep it within range (Similar to the implementation in hashEmbed)

**Additional Comments**
Documentation for Embed needs to be updated and further discussion is needed on the input Parameter nV. It does not make sense to default it to 1. nV typically needs to be the size of the vocabulary of the corpus to reduce collisions. 

**Why does HashEmbed work with small nV values then?** 
Given the way the hashing has been implemented in HashEmbed, it technically generates nV^4 different combinations for the word to embedding mapping. Hence, even an nV of 16 was sufficient to cover the entire vocabulary. The same does not happen in Embed, hence the need to set nV to size of the vocab. Even in the case of HashEmbed, if someone sets nV too small such that nV^4 << vocab size then the collisions would be high and model accuracy takes a big hit.

**Future Discussions**
Why is ids parameter generating such large and sparse numbers in the first place? Why dont we have a continuous mapping of the numbers there? It would do away with the need for a modulo operation. In fact the modulo operation is not entirely reliable in handling collisions.

### DIFF
--- a/thinc/layers/embed.py
+++ b/thinc/layers/embed.py
@@ -50,7 +50,9 @@ def forward(
     vectors = cast(Floats2d, model.get_param("E"))
     nO = vectors.shape[1]
     nN = ids.shape[0]
+    nV = vectors.shape[0]
     dropout: Optional[float] = model.attrs.get("dropout_rate")
+    ids = ids % nV 
     output = vectors[ids]
     drop_mask = cast(Floats1d, model.ops.get_dropout_mask((nO,), dropout))
     output *= drop_mask


adrianeboyd:
Hmm, `Embed` and `HashEmbed` aren't really intended to be interchangeable. The docs could definitely be improved here, because it's not clear unless you read the code that `strings2array` uses hashing rather than some other kind of vocab indexing and as a result, it's only intended as input for a layer like `HashEmbed`. (The background is that a lot of these layers were developed with spacy in mind, which represents all strings as `uint64` hashes internally.)

`Embed` is intended for things like vocab IDs in a small range like `0..len(vocab)` and stores a unique vector for each ID, so it seems acceptable to have errors when `nV` is too small. It's certainly possible to write a custom embedding layer that uses `%` to handle IDs outside the range `0..nV`, but I think it's probably not a good idea to do this in the default `Embed` layer. Since `nV` is a parameter that needs to be provided, having a default of `nV=1` as the smallest possible size seems fine.

You can see an example with `Embed` in https://github.com/explosion/thinc/blob/master/examples/03_textcat_basic_neural_bow.ipynb.

naveenjafer:
@adrianeboyd The separation of HashEmbed and Embed makes sense, but I am trying to understand a use case where a default value of nV=1 would be helpful or even work for that matter. For any input vocab size,  it would simply throw an array index error, right? Typically as developers, default value if available in a parameter is still expected to run the code without errors, and if and when needed, that parameter is tweaked to improve performance/results.

honnibal:
@naveenjafer While it's true that you normally want to create objects in a working state, it's not always possible to do everything at once. This is especially true with the config system, where we want to let the objects get created during the `create_from_config` pass. In this circumstance, we need the object to get created, and we want the members to have their appropriate types. It's not really wonderful, but it's the least bad solution.

PR #311, opened 2020-02-14
Set Embed shape to (nV, nO)

adrianeboyd:
Use `nV` rather than `nV + 1` to remove all whiffs of OOV handling from `Embed`.


adrianeboyd:
It is slightly annoying that `nV` has to be `(max(col) + 1)`, but I think the alternative is worse?

PR #315, opened 2020-02-27
init instead of _init

svlandeg:
As we may need to copy/set the `Model`'s `init` function in downstream applications, renaming from `_init` to `init`.

Also expanded / structured the `gitignore` file, similar as how we're doing it in spaCy. Nothing was removed from the original list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,44 +1,137 @@
-*.egg-info/
-dist/
-*.dat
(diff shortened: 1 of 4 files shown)


svlandeg:
The failing CI tests will be fixed by https://github.com/explosion/thinc/pull/316

PR #319, opened 2020-03-07
Support for additional initializers from keras

naveenjafer:
Includes Lecun_Unfiorm, He_Uniform and Glorot_Normal initializers - Reference - https://keras.io/initializers/ 

Completes - 
TODO: Harmonize naming with Keras, and fill in missing entries https://keras.io/initializers/ We should also have He normal/uniform and probably lecun normal/uniform.

Will add Lecun and He Normal once Glorot normal is approved

### DIFF
--- a/thinc/initializers.py
+++ b/thinc/initializers.py
@@ -15,9 +15,31 @@ from .util import partial
 # It's especially helpful for JAX, which has a pretty intrincate PRNG scheme I
 # haven't figured out yet.
 
+
+def lecun_normal_init(ops: Ops, shape: Shape) -> FloatsXd:
+    scale = numpy.sqrt(1.0 / shape[1])
+    return ops.asarray_f(numpy.random.normal(0, scale, shape))
+
+
+@registry.initializers("lecun_normal_init.v1")
+def configure_lecun_normal_init() -> Callable[[Shape], FloatsXd]:
+    return partial(lecun_normal_init)
+
+
+def he_normal_init(ops: Ops, shape: Shape) -> FloatsXd:
+    scale = numpy.sqrt(2.0 / shape[1])
+    return ops.asarray_f(numpy.random.normal(0, scale, shape))
+
+
+@registry.initializers("he_normal_init.v1")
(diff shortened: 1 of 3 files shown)


thinc/initializers.py
```diff
@@ -29,7 +54,6 @@ def configure_glorot_uniform_init() -> Callable[[Shape], FloatsXd]:
 def zero_init(ops: Ops, shape: Shape) -> FloatsXd:
     return ops.alloc(shape)
 
-
```
svlandeg: As an FYI, we use the `black` formatting tool in Thinc and in spaCy. I believe in this case it would keep these additional lines: it has 2 empty lines between functions declared at the top level. Could you install `black` in your IDE to make sure we're not introducing unnecessary formatting changes in PR's ? cf https://black.readthedocs.io/en/stable/

honnibal:
Thanks a lot for these PRs that fill out these extra pieces of the library!

I want to emphasise that we really appreciate the help. We're trying to keep the library very consistent, so some of our comments might end up nitpicky and the process of merging the PRs might get a little tedious sometimes. Sometimes it might be easier for us to merge first and then make some changes. I hope you won't find this discouraging --- we just want to make sure we keep maintaining consistency.

naveenjafer:
> Thanks a lot for these PRs that fill out these extra pieces of the library!
> 
> I want to emphasise that we really appreciate the help. We're trying to keep the library very consistent, so some of our comments might end up nitpicky and the process of merging the PRs might get a little tedious sometimes. Sometimes it might be easier for us to merge first and then make some changes. I hope you won't find this discouraging --- we just want to make sure we keep maintaining consistency.

Absolutely, I understand :). The least one can do is be patient when using and contributing to an open source repo.

naveenjafer:
> I think this looks great! You could add the few missing ones in this PR, that keeps it nicely together :-)

Thank you for the update about black :) . I have added the remaining initializers and black checked the syntax.

svlandeg:
Looks good! I think this can be merged if Matt agrees :-)

PR #324, opened 2020-03-10
set nI/nO on chained models

svlandeg:
Ensure that the `nO` and `nI` of a model, that gets chained in its constructor, are set appropriately.

Before this PR, you'd have:
```
width=64
lower = HashEmbed(nO=width, nV=embed_size, column=cols.index(LOWER))
print(lower.get_dim("nO"))
```
returning

> Cannot get dimension 'nO' for model 'ints-getitem>>hashembed': value unset

which is counter-intuitive and unhelpful...

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,7 +15,7 @@ jobs:
         imageName: 'vs2017-win2016'
         python.version: '3.6'
       Python36Mac:
-        imageName: 'macos-10.13'
+        imageName: 'macos-10.15'
         python.version: '3.6'
       Python38Linux:
         imageName: 'ubuntu-16.04'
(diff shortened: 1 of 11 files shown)


honnibal:
It feels like this would be better fixed in `chain`, right?

https://github.com/explosion/thinc/blob/master/thinc/layers/chain.py#L20

We should have logic in the init that avoids defining `nI` and `nO` if the first and last layers (respectively) don't define them, and which uses their values if their values are there. Something like:

```
dims = {}
if layers[0].has_dim("nI") is not False:
    dims["nI"] = layers[0].get_dim("nI")
if layers[-1].has_dim("nO") is not False:
    dims["nO"] = layers[0].get_dim("nO")
```

svlandeg:
Yes, ofcourse! In a sense it's very similar to the `nI` stuff in `concatenate`.

svlandeg:
This turned out to be a little less straightforward then I'd hoped. If I remove `nI`/`nO` entirely from `chain` when its first/last layer doesn't have that dimension, many tests in spaCy start breaking :|

I've made a minimal change for now that at least fixes the present issue, as I'm in the middle of a different upgrade in spaCy and don't want things to get too mixed up...

PR #327, opened 2020-03-18
with_cpu

svlandeg:
To accomodate for the spaCy BOW text categorizer, I transferred the `with_cpu` code from spaCy 2.X to here.

There was trouble with running `finish_update` though, because the optimizer couldn't deal with the Numpy weights from the layers on CPU. Not sure how that used to work. Probably I'm missing an easier/prettier fix, but for now this seems to work.

### DIFF
--- a/thinc/model.py
+++ b/thinc/model.py
@@ -299,13 +299,16 @@ class Model(Generic[InT, OutT]):
         with each parameter and gradient of the model.
         """
         for node in self.walk():
+            orig_ops = node.ops
             for name in node.param_names:
                 if node.has_grad(name):
                     param = node.get_param(name)
                     grad = node.get_grad(name)
-                    param, grad = optimizer((node.id, name), param, grad)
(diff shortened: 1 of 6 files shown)


thinc/model.py
```diff
@@ -299,13 +299,17 @@ def finish_update(self, optimizer: Optimizer) -> None:
...
         """
         for node in self.walk():
+            orig_ops = node.ops
             for name in node.param_names:
                 if node.has_grad(name):
                     param = node.get_param(name)
                     grad = node.get_grad(name)
+                    if hasattr(optimizer,'ops'):
+                        param = optimizer.ops.xp.asarray(param)
+                        optimizer.ops.xp.asarray(grad)
```
honnibal: Why `ops.xp.asarray` and not `os.asarray`?

svlandeg: I think that was wrong ...

PR #329, opened 2020-03-23
multiply by drop_mask in staticvectors.backprop

svlandeg:
Also moved some lines. Not sure this actually ever worked correctly before? cf https://github.com/explosion/thinc/blob/v7.x/thinc/neural/_classes/static_vectors.py#L66 - looks like `mask` was always `None` previously?

Anyway this fix does boost spaCy's parser performance by 0.5pp.


svlandeg:
Looks like the tests fail on the `tensorflow-2.2.0rc1` for Python 3.8.

PR #332, opened 2020-03-26
Fix CPU/GPU IO by checking the layer before setting the param

svlandeg:
If we have a mixed model with CPU & GPU layers, the current IO doesn't work because `model.from_bytes` [converts](https://github.com/explosion/thinc/blob/master/thinc/model.py#L507) the read data with `msg = convert_recursive(is_xp_array, self.ops.asarray, msg)` which results in all-cupy or all-numpy arrays across the model.

We used to have layer-specific conversions though, cf. https://github.com/explosion/thinc/blob/v7.x/thinc/neural/_classes/model.py#L357, but this has gotten a bit more tricky with the current design using `convert_recursive`.

This PR has a quick hack that does the conversion of a `param` once more before actually setting it. It feels a little inefficient, because an array may be converted from `numpy` to `cupy` and back... But this keeps the code change minimal. Happy to discuss other options!

### DIFF
--- a/thinc/about.py
+++ b/thinc/about.py
@@ -1,2 +1,2 @@
-__version__ = "8.0.0a3"
+__version__ = "8.0.0a4"
 __release__ = True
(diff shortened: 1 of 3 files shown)


svlandeg:
Something is not yet quite right. This extra conversion means that the length of bytes is altered when doing `from_bytes.to_bytes` in spaCy's roundtrip tests

PR #333, opened 2020-03-31
Add tests for notebook examples

justindujardin:
I think I figured out how we can test our ipython notebooks in the `examples` folder! The `nbconvert` package provides [a way to execute notebooks](https://nbconvert.readthedocs.io/en/latest/execute_api.html#executing-notebooks-using-the-python-api-interface)

When there's an error it looks like this:
Screen Shot 2020-03-31 at 1 21 48 PM

The downside to these tests is that they can be slow, or break if you don't have the right system configuration (e.g. no GPU for config notebook). I put the notebooks that execute quickly in one test, and the others in one that is wrapped with `@pytest.mark.slow`

Changes
 - run notebooks files in `examples` folder and assert that there are no errors in cell outputs
 - add a slow test for the notebooks that take a long time to run

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,15 +19,15 @@ jobs:
       Python36Mac:
         imageName: 'macos-10.15'
         python.version: '3.6'
-      # Python38Linux:
-      #   imageName: 'ubuntu-16.04'
-      #   python.version: '3.8'
-      # Python38Windows:
-      #   imageName: 'vs2017-win2016'
-      #   python.version: '3.8'
-      # Python38Mac:
-      #   imageName: 'macos-10.15'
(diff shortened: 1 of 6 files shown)


justindujardin:
I wired up the notebook tests to run on Azure (and they pass :tada:), but there are a few bits that should probably be improved:

 - the "slow" notebooks aren't executed during test runs. That means any slow notebooks won't get the benefit of regression checks. We could make the notebooks faster, or make sure to run the "slow" tests once before each release?
 - at least one notebook requires a GPU to execute properly. Are there any systems we can execute GPU tests on? I currently put them in the "slow" category.

If we resolved the two above items I think that we could test all of the example notebooks. For now, I suppose a few of them are better than none.

PR #335, opened 2020-04-13
Remove ops attribute from Optimizer

honnibal:
The `Optimizer` object can't assume that all layers are using the same backend, because we might have mixed CPU/GPU execution. So the optimizer needs to fetch an ops appropriate to the inputs.

We remove the `ops` argument to the optimizer, and add a helper `get_array_ops`.

### DIFF
--- a/thinc/model.py
+++ b/thinc/model.py
@@ -306,9 +306,6 @@ class Model(Generic[InT, OutT]):
                 if node.has_grad(name):
                     param = node.get_param(name)
                     grad = node.get_grad(name)
-                    if hasattr(optimizer, "ops"):
-                        param = optimizer.ops.asarray(param)  # type: ignore
-                        grad = optimizer.ops.asarray(grad)  # type: ignore
                     param, grad = optimizer((node.id, name), param, grad)
                     node.set_param(name, orig_ops.asarray(param))  # type: ignore
                     node.set_grad(name, orig_ops.asarray(grad))  # type: ignore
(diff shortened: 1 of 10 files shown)


thinc/model.py
```diff
@@ -304,12 +306,12 @@ def finish_update(self, optimizer: Optimizer) -> None:
                 if node.has_grad(name):
                     param = node.get_param(name)
                     grad = node.get_grad(name)
-                    if hasattr(optimizer, 'ops'):
-                        param = optimizer.ops.asarray(param)        # type: ignore
-                        grad = optimizer.ops.asarray(grad)          # type: ignore
+                    if hasattr(optimizer, "ops"):
```
svlandeg, line 309: is this check still useful ?

honnibal: It's not, no. We should remove that code

honnibal, line 309:
```suggestion

```

thinc/model.py
```diff
@@ -304,12 +306,12 @@ def finish_update(self, optimizer: Optimizer) -> None:
...
+                        param = optimizer.ops.asarray(param)  # type: ignore
```
honnibal, line 310:
```suggestion
```

thinc/model.py
```diff
@@ -304,12 +306,12 @@ def finish_update(self, optimizer: Optimizer) -> None:
...
+                        grad = optimizer.ops.asarray(grad)  # type: ignore
```
honnibal, line 311:
```suggestion
```

PR #340, opened 2020-05-07
fix(to_categorical): make consistent with tf/torch one_hot helpers

### LINKED ISSUE #339: to_categorical discards input shape
While porting a PyTorch example to thinc, I noticed that `to_categorical` does not produce the same results as TensorFlow or PyTorch. Specifically, thinc discards the shape when dealing with more than 2d inputs.

While looking at other libraries to compare results, I noticed there are differences between how the output shape is calculated based on whether you're using TF/Torch helpers, or `keras.utils.to_categorical` 😅

Here's a snippet demonstrating the differences:
```python
>>> import numpy as np
>>> data = np.random.randint(0,12, size=(2,3,1))

>>> from thinc.api import to_categorical
>>> to_categorical(data, n_classes=12).shape
(6, 12)

>>> import torch
>>> torch.nn.functional.one_hot(torch.from_numpy(data),12).shape
torch.Size([2, 3, 1, 12])

>>> import tensorflow as tf
>>> tf.one_hot(data, depth=12).shape
TensorShape([2, 3, 1, 12])

>>> import keras
>>> keras.utils.to_categorical(data, 12).shape
(2, 3, 12)
```

@honnibal what should be the correct behavior for thinc? Perhaps `[2, 3, 1, 12]` since TensorFlow and PyTorch agree on that output?

justindujardin:
The to_categorical helper outputs different shapes than PyTorch and TensorFlow equivalent helpers. 

This "fixes" it so that Thinc/TF/Torch all agree on the output.

 - resume testing to_categorical helper _(lol at xfail reason)_
 - to_categorical restores the input shape before returning
 - closes #339

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -72,7 +72,10 @@ jobs:
   - script: |
       pip install tensorflow
       pip install "mxnet; sys_platform != 'win32'"
-      pip install torch -f https://download.pytorch.org/whl/torch_stable.html
+      pip install "mxnet; sys_platform != 'win32'"
+      # NOTE: need this to close for 1.5 on windows: https://github.com/pytorch/pytorch/issues/37209
+      pip install "torch==1.4; sys_platform == 'win32'" -f https://download.pytorch.org/whl/torch_stable.html
+      pip install "torch; sys_platform != 'win32'" -f https://download.pytorch.org/whl/torch_stable.html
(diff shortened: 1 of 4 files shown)


honnibal:
Nice, thanks! So is it just the behaviour on higher-dimensional arrays?

PR #342, opened 2020-05-09
fix(xp2torch): don't explode if input is already a tensor

justindujardin:
With the current behavior, you have to be careful to check the input types before calling xp2torch. With this change, xp2torch returns the tensor directly if given one.

Does this fix look alright? If returning a tensor do I need to do any device checking to ensure the tensor is on the current device?


honnibal:
Hmm, I don't know if this is the direction we want to go in. Maybe we can have a `to_torch`, similar to `to_numpy`? I generally want to have an approach where you should know what type you have. xp arrays and torch arrays aren't interchangeable, so your code should generally know which you have.

justindujardin:
Cool, I don't disagree that you should know what kind of array you have. In practice I found that it was often unclear which kind of array to expect when you're mixing Thinc and Torch pieces.

PR #344, opened 2020-05-11
add "PyTorchCrossEntropy.v1" loss

justindujardin:
The PyTorch cross_entropy loss uses log softmax + negative-log-likelihood and it performs better on certain tasks than the thinc CCE loss:
Screen Shot 2020-05-11 at 12 39 21 PM

I'm not sure if this loss is desirable as part of thinc, but I figured it couldn't hurt to have a PR to look at and talk about.

If this is something thinc would like to keep I can add better tests and document the loss class options.

### DIFF
--- a/thinc/loss.py
+++ b/thinc/loss.py
@@ -6,7 +6,7 @@ from .util import get_array_module, to_categorical, xp2torch
 from .config import registry
 
 
-# Needed to allow
+# Needed to keep "torch.Tensor" types from causing mypy errors
 try:  # pragma: no cover
     import torch
 except ImportError:  # pragma: no cover
--- a/thinc/loss.py
+++ b/thinc/loss.py
@@ -2,8 +2,15 @@ from typing import Tuple, List, cast, TypeVar, Generic, Any, Union, Optional
 from typing import Dict
 
 from .types import Floats2d, Ints1d
-from .util import get_array_module, to_categorical
+from .util import get_array_module, to_categorical, xp2torch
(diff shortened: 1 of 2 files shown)


honnibal:
I think it's important to notice that this calculates something different from the CCE we have, but I wouldn't want to merge literally this. Like, we don't want one loss that outputs `torch.Tensor`, and we shouldn't need to call into PyTorch for this. We would need to figure out exactly what it's computing to document it, at which point we'd want to implement it.

Btw I think the point about not returning `torch.Tensor` is worth emphasising. The library will be in pretty bad shape if we start returning a mix of arrays and torch tensors. The two objects serve very similar purposes, but you can't mix them at all. So we should never be accepting or returning torch tensors, you can convert to the numpy/cupy container without a copy and that's what we should always be doing.

justindujardin:
> Like, we don't want one loss that outputs torch.Tensor

Just to be clear, it only uses the `torch.Tensor` internally, both the get_grad and get_loss functions return numpy arrays (which should have been torch2xp). I felt awkward including the torch import for just the type annotation, but.. I can't help it, I :heart: types so much! 

> and we shouldn't need to call into PyTorch for this.

Yeah, I understand that, but it's surprisingly difficult to find information disambiguation the various kinds of cross entropy losses. It's even more difficult when you have to also find the correct derivations to calculate the grads manually. I could attempt this, but it would be days or more of research for me to get it all right. 😓

I'll close this for now 👍

PR #346, opened 2020-05-11
Use thinc optimizer in Shim classes

justindujardin:
Update the shim classes to use the thinc optimizers instead of creating framework-specific ones. 

- [x] Update tensorflow shim
- [x] Update mxnet shim
- [x] Update torch shim
- [x] Verify torch with cuda
- [x] Verify unique parameter ID use in torch/mxnet

### DIFF
--- a/thinc/shims/mxnet.py
+++ b/thinc/shims/mxnet.py
@@ -1,4 +1,4 @@
-from typing import Any
+from typing import Any, cast
 import srsly
 import copy
 
@@ -9,9 +9,9 @@ try:
 except ImportError:  # pragma: no cover
     pass
 
-from ..util import mxnet2xp, convert_recursive, make_tempfile
-from ..backends import get_array_ops
-from ..types import ArgsKwargs
+from ..util import mxnet2xp, convert_recursive, make_tempfile, xp2mxnet
+from ..optimizers import Optimizer
+from ..types import ArgsKwargs, FloatsXd
 from .shim import Shim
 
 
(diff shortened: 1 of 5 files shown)


thinc/shims/mxnet.py
```diff
@@ -56,53 +56,14 @@ def backprop(grads):
...
-                ops.update_averages(sgd.averages[key], xp_param, sgd.nr_update[key])
-            else:
-                sgd.averages[key] = xp_param.copy()
-                sgd.nr_update[key] = init_steps
+    def finish_update(self, optimizer: Optimizer):
+        ctx = mx.current_context()
+        for value in self._model.collect_params().values():
+            grad = cast(FloatsXd, mxnet2xp(value.grad(ctx)))
+            param = cast(FloatsXd, mxnet2xp(value.data(ctx)))
+            param, _ = optimizer((value.name, value.name), param, grad)
```
justindujardin, line 64: I don't think the `(value.name, value.name)` tuples will be unique necessarily, and the "update_averages" code fails with shape errors if you have overlapping keys that don't match. I need to figure out how to get unique parameter IDs out of mxnet and pytorch

thinc/shims/pytorch.py
```diff
@@ -53,39 +54,13 @@ def backprop(grads):
...
-                args["momentum"] = sgd.b1
-        if self._optimizer is None:
-            self._optimizer = cls(self._model.parameters(), **args)
-        else:
-            for param_group in self._optimizer.param_groups:
-                param_group.update(args)
-        return self._optimizer
+    def finish_update(self, optimizer: Optimizer):
+        for name, value in self._model.named_parameters():
+            cpu_data = cast(FloatsXd, torch2xp(value.data))
```
honnibal, line 59: Hm, `cpu_` seems like a misnomer, right? We're not copying between devices, we're just changing the container. We can call it `xp_data`, and maybe make the loop variable `torch_data`? Alternatively I've been using `array` for the numpy/cupy typed data and `tensor` for the pytorch-typed data, but that's maybe confusing.

honnibal:
Looking good! Just that naming tweak and then we can go ahead and merge.

thinc/shims/pytorch.py
justindujardin: Nice catch, you're right! I originally had this as `value.data.cpu().numpy()` before switching to torch2xp. xp_data sounds more appropriate now

PR #359, opened 2020-07-09
Fixing random seed to ensure reproducibility

svlandeg:
WIP

Original post, which was quite wrong, below in collapsible section to avoid future confusion...

<details>

### Background
It looks like we haven't really been supporting `fix_random_seed` well. This function would call `random.seed(value)`, but the problem is that this call needs to be repeated in between generating random numbers:
```
random.seed(666)
print("getting int 0: ", random.randint(0, 100000))
print("getting int 1: ", random.randint(0, 100000))
print("getting int 2: ", random.randint(0, 100000))
```
Gives

> getting int 0:  59784
> getting int 1:  49417
> getting int 2:  56934

But in contrast,
```
random.seed(666)
print("getting int 0: ", random.randint(0, 100000))
random.seed(666)
print("getting int 1: ", random.randint(0, 100000))
random.seed(666)
print("getting int 2: ", random.randint(0, 100000))
```
Gives

> getting int 0:  59784
> getting int 1:  59784
> getting int 2:  59784

Note, btw, that this is also a problem for list comprehensions:
```
random.seed(666)
print("getting int 666: ", [random.randint(0, 100000) for _ in range(10)])
```

> getting int 666:  [59784, 49417, 56934, 37158, 65618, 1864, 71792, 72571, 93346, 43664]

### Fix
We can't expect the user to call `fix_random_seed` multiple times, because internally we may be generating multiple random numbers in between atomic calls for the user. So we need to do this within `Thinc` and we need to manage it carefully so we call `ensure_fixed_seed` EVERY TIME before we generate something random.

Currently this PR solves this with a global variable `INTERNAL_SEED` in `thinc.util`. I'm not too happy about that, and worry about consequences for parallellization. But if multiple threads/workers need to reproduce the same results, there will be no other option then to make sure they are initialized with the same seed.

Added also the functionality to `remove_random_seed`.

Finally, `HashEmbed` needs to also use this seed if it's available and is not overwritten in its constructor. The new tests in `test_hash_embed` show the different possibilities. Many of these would be failing on the current `master` branch.

I was also able to reproduce the original bug with spaCy, and after this fix, those issues seem solved. Will submit the new unit tests for spaCy on `develop` there.
</details>

### DIFF
--- a/examples/benchmarks/lstm_tagger.py
+++ b/examples/benchmarks/lstm_tagger.py
@@ -17,7 +17,7 @@ import numpy.random
 from timeit import default_timer as timer
 from thinc.api import Model, Config, registry, chain, list2padded, with_array
 from thinc.api import to_categorical, set_current_ops, JaxOps
-from thinc.api import NumpyOps, CupyOps, fix_random_seed, require_gpu
+from thinc.api import NumpyOps, CupyOps, fix_random_seed, ensure_fixed_seed, require_gpu
 from thinc.types import Array2d, Padded
 import jax.tree_util
 
@@ -74,10 +74,13 @@ def get_dummy_data(n_samples, n_tags, n_vocab, length_mean, length_variance):
     Xs = []
     Ys = []
     for _ in range(n_samples):
+        # ensure_fixed_seed()
         # length = numpy.random.normal(size=1, scale=length_variance) + length_mean
         length = length_mean
         shape = (max(1, int(length)),)
+        ensure_fixed_seed()
         X = numpy.random.uniform(0, n_vocab - 1, shape)
+        ensure_fixed_seed()
         Y = numpy.random.uniform(0, n_tags - 1, shape)
         assert X.size, length
         assert Y.size, length
--- a/examples/benchmarks/lstm_tagger.py
+++ b/examples/benchmarks/lstm_tagger.py
@@ -17,7 +17,7 @@ import numpy.random
 from timeit import default_timer as timer
 from thinc.api import Model, Config, registry, chain, list2padded, with_array
 from thinc.api import to_categorical, set_current_ops, JaxOps
-from thinc.api import NumpyOps, CupyOps, fix_random_seed, ensure_fixed_seed, require_gpu
+from thinc.api import NumpyOps, CupyOps, fix_random_seed, require_gpu
 from thinc.types import Array2d, Padded
 import jax.tree_util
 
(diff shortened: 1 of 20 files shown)


honnibal:
> It looks like we haven't really been supporting fix_random_seed well. This function would call random.seed(value), but the problem is that this call needs to be repeated in between generating random numbers:

Maybe I'm missing your intention here, but this is what I expected, and I'm not sure the current code is wrong in this respect?

Setting the seed on the generator sets it into a particular state. The next number is determined by the state, so by setting the seed, you control what number will be returned next.

The idea is that when you start the program, the RNG starts in an arbitrary state...and then we seed the RNG to put it in the same state each time. If we then the RNG a deterministic number of times in a deterministic order, we should see the same RNGs between runs.

Note btw that the `HashEmbed` uses its seed in a different way. The seed there is for its hash function, rather than the RNG.

svlandeg:
Ugh, yea you're right ofcourse. Let me poke at this further.

svlandeg:
It's just simpler, right. We do need to fix the `HashEmbed`'s `seed` if we're working with a fixed seed, because otherwise it's `forward` `output` just won't be the same?

svlandeg:
So I do think the variation stems from the `HashEmbed`'s `seed` though?

By calling `fix_random_seed`, `HashEmbed` gets the same fixed `vectors` initialized, but because of the difference in hashing, the `output` result differs, and this will influence the next layers?

honnibal:
Each `HashEmbed` table should get a different seed from each other, so that if you have a short word its `NORM`, `PREFIX` and `SUFFIX` vectors end up different from each other. I achieve this by using the `model.id` as the hash-embed table's seed if none is provided, because that id should be global.

This does have an annoyance though, in that you'll get different results if you run twice within the same process. I actually pushed a change to the `Tok2Vec` recently that simply sets the seeds explicitly, with values 0-4. But the results are still not reproducible after that.

svlandeg:
Ah, that would explain why the spaCy reproducibility tests work for `Tok2vec` but not for the default `textcat` architecture, which uses its own embedded `tok2vec` architecture which doesn't set these seeds specifically (all of that needs to be cleaned up but that's a different matter)?

svlandeg:
Ok, so if I fix the seeds also for the `textcat` architecture, all the tests in that spaCy PR pass. Was there another architecture where you will still seeing irreproducibility?

PR #366, opened 2020-07-15
robustness against empty X

svlandeg:
Various checks in case `X` is empty, e.g. in spaCy running with `doc = Doc(vocab, words=[])`


honnibal:
Nice, thanks!

PR #374, opened 2020-08-04
Allow config overrides on creation (and used with interpolation)

ines:
The `Config.from_str`, `Config.from_bytes` and `Config.from_disk` methods can now take optional `overrides` that are added to the config parser _before_ the config is interpreted. This allows overwriting values that are referenced as variables in other sections, without breaking the config.

### Example

```ini
[section]
a = 2
b = 3

[section.subsection]
c = "hello"

[other_section]
d = ${section:a}
e = ${section.subsection}
```

```python
overrides = {"section.a": 20, "section.subsection.c": "world"}
config = Config.from_str(config_str, overrides=overrides)
```

```python
# Result
{
  "section": {
    "a": 20,
    "b": 3,
    "subsection": {
      "c": "world"
    },
  "other_section": {
    "d": 20,
    "e": {
       "c": "world"
    }
  }
}
```

### DIFF
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -3,6 +3,7 @@ from types import GeneratorType
 from configparser import ConfigParser, ExtendedInterpolation, MAX_INTERPOLATION_DEPTH
 from configparser import InterpolationMissingOptionError, InterpolationSyntaxError
 from configparser import NoSectionError, NoOptionError, InterpolationDepthError
+from configparser import SectionProxy
 from pathlib import Path
 from pydantic import BaseModel, create_model, ValidationError
 from pydantic.main import ModelMetaclass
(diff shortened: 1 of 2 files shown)


thinc/config.py
```diff
@@ -195,10 +196,24 @@ def copy(self) -> "Config":
             raise ValueError(f"Couldn't deep-copy config: {e}")
         return Config(config)
 
-    def from_str(self, text: str) -> "Config":
-        "Load the config from a string."
+    def _set_overrides(self, config: "ConfigParser", overrides: Dict[str, Any]) -> None:
+        """Set overrides in the ConfigParser before config is interpreted."""
+        err_title = "Error parsing config overrides"
+        for key, value in overrides.items():
+            parts = key.split(".")
```
honnibal, line 203: This is not really significant, and maybe you find this code structure worse. But how about:

```
if "." not in key:
    raise ConfigValidationError("", err, message=err_title)
section, field = key.rsplit(".", 1)
if section not in config:
    raise ConfigValidationError("", err, message=err_title) 
config.set(section, field, srsly.json_dumps(value))
```

PR #377, opened 2020-08-12
Support optional variable interpolation and fix config resolution

ines:
* Allow setting `interpolate=False` on `Config.to_str`, `Config.from_str` and other serialization methods. This will preserve variables in the configs, and will make it easier for us to provide "starter configs" that can be filled without destroying the variables.
* Support non-interpolated configs in `registry.resolve`: in that case, we keep a copy of the original config, fill and resolve the interpolated version (because that's the only way it can work) and then merge the original config with its variable references back into the filled config. This makes it possible to auto-fill a config with variables and add missing defaults etc. while keeping the references intact.
* Add `Config.interpolate` method to interpolate an existing config (returns a copy).
* Port config merging logic over from spaCy so it can live in Thinc instead, and add `Config.merge` that takes a configs and merges it with the current config, wtih some special logic to handle variables and promises.
* Make it illegal save or interpolate a config with top-level keys that aren't sections (e.g. `{"foo": 1}`, which can only really happen if you construct a `Conig` from a dict). Otherwise, Python's configparser will add them to a section `[DEFAULT]` and when the serialized config is loaded back in from a string, those values are added to _every section_ (!) which is almost never what you want and causes very confusing results and validation errors.
* Raise better error if a variable references a section within a string, i.e. `"hello ${foo.bar}"`. If the section is a regular dict, this _would_ work in theory, and you'd get the string `"hello {'x': 'y'}"` – but it most likely indicates a mistake. If the section is a reference to a registered function, it can't work, because registered functions are resolved _after_ the config is created and interpolated. 

### Open questions

* Does it make sense to have `interpolate=True` as the default? Realistically, we'd be setting it to `False` a lot in spaCy internally, but from a user's perspective, I think it makes more sense if `Config.from_str` etc. always returns the interpolated final version and doesn't require calling `.interpolate()` manually or setting `interpolate=True`? Or maybe it _is_ actually more logical if it _doesn't_ "destroy" variables by default? `registry.resolve` will always interpolate anyways and can handle variables, so that's no problem. So I might actually be leaning towards `default=False` now 🤔

### DIFF
--- a/thinc/about.py
+++ b/thinc/about.py
@@ -1,2 +1,2 @@
-__version__ = "8.0.0a23"
+__version__ = "8.0.0a24"
 __release__ = True
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -397,8 +397,11 @@ def deep_merge_configs(
             config[key] = value
         elif isinstance(value, str) and re.search(VARIABLE_RE, value):
             # If the original values was a variable or a string containing a
-            # reference to the variable, we always prefer the variable
-            config[key] = value
+            # reference to the variable, we always prefer the variable (unless
+            # the new value is also a variable).
+            orig = config[key]
+            if not isinstance(orig, str) or not re.search(VARIABLE_RE, orig):
+                config[key] = value
     return config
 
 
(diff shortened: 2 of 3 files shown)


honnibal:
I kind of lean towards `interpolate=True` I think. I think the most basic thing is like, you want to get the values. So it makes sense to me to handle the interpolation.

PR #378, opened 2020-08-13
Fix reporting of whether config is interpolated

ines:
This is relevant, because it impacts how a config is resolved and whether it's interpolated before.

### Summary of interpolation / resolving non-interpolated configs

```python
config_str = """
[variables]
width = 123

[model]
@architectures = "my_model"
width = ${variables:width}
"""

@registry.architectures("my_model")
def my_model(width: int, depth: int = 456):
    ...
```

Auto-fill partial config (e.g. as part of `init config` command)

```python
config = Config().from_str(config_str)
filled = registry.fill_config(config)
```

Result:

```ini
[variables]
width = 123

[model]
@architectures = "my_model"
width = 123
depth = 456
```

Problem: reference to `${variables:width}` is gone. A user's CLI script that overwrites `--variables.width` will now produce different and very misleading results. All the convenience of referencing blocks and values is gone, just because the user wanted to fill in a few defaults.

Solution: we now also support loading configs without interpolating and  resolving the variables on load. Config can be interpolated later by calling `config.interpolate()`.

```python
raw_config = Config().from_str(config_str, interpolate=False)
# filled = registry.fill_config(raw_config)
```

Next problem: resolving (create functions, check signatures) and filling (add missing defaults) is always done jointly in `registry.resolve`. In order to do this, we need to know all values so we can create the functions, so we need to interpolate the config.

```python
# internally:
interpolated = raw_config.interpolate()
filled, resolved = registry.resolve(interpolated)
```

Next problem: The filled config now looks like the config above and doesn't include variables anymore. But we still have our original non-interpolated config, so we can create a final config by putting the two configs together:

* the filled config provides the "truth", including additional properties
* the original non-interpolated config provides the overrides for the fields that were originally variables

```python
# internally:
filled = filled.merge(raw_config)
```

Result:

```ini
[variables]
width = 123

[model]
@architectures = "my_model"
width = ${variables:width}
depth = 456
```

### DIFF
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -133,6 +133,7 @@ class Config(dict):
     def __init__(
         self,
         data: Optional[Union[Dict[str, Any], "ConfigParser", "Config"]] = None,
+        *,
         is_interpolated: Optional[bool] = None,
     ) -> None:
         """Initialize a new Config object with optional data."""
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -254,12 +254,14 @@ class Config(dict):
             raise ValueError(f"Couldn't deep-copy config: {e}") from e
         return Config(config, is_interpolated=self.is_interpolated)
 
-    def merge(self, updates: Union[Dict[str, Any], "Config"]) -> "Config":
+    def merge(
+        self, updates: Union[Dict[str, Any], "Config"], *, prefer_vars: bool = False
+    ) -> "Config":
         """Deep merge the config with updates, using current as defaults."""
         defaults = self.copy()
         updates = Config(updates).copy()
         is_interpolated = defaults.is_interpolated and updates.is_interpolated
-        merged = deep_merge_configs(updates, defaults)
+        merged = deep_merge_configs(updates, defaults, prefer_vars=prefer_vars)
         return Config(merged, is_interpolated=is_interpolated)
 
     def _set_overrides(self, config: "ConfigParser", overrides: Dict[str, Any]) -> None:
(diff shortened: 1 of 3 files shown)


thinc/config.py
```diff
@@ -128,8 +128,12 @@ class Config(dict):
     under the hood.
     """
 
+    is_interpolated: bool
+
     def __init__(
-        self, data: Optional[Union[Dict[str, Any], "ConfigParser", "Config"]] = None
+        self,
+        data: Optional[Union[Dict[str, Any], "ConfigParser", "Config"]] = None,
+        is_interpolated: Optional[bool] = None,
```
honnibal, line 136: We probably want this keyword only? Maybe there's a reason to allow positional, but I'm trying to remember to do the KWO args, so I'm trying to remember to suggest it in reviews.

```suggestion
        *,
        is_interpolated: Optional[bool] = None,
```

honnibal: Would it be better to make the argument name `interpolate`? I'm not sure.

To me `is_interpolated` leans a little more towards describing the state of the data being passed in. I'd normally prefer a verb to indicate whether to perform the action.

I would definitely keep the `is_interpolated` name for the boolean, though.

ines: > To me is_interpolated leans a little more towards describing the state of the data being passed in.

Yes, that's the intention. The `is_interpolated` argument here is purely to note the state of the data being passed in, if the `Config` is created from a dict. So it's mostly a shortcut for setting `config.is_interpolated` on a separate line.

PR #379, opened 2020-08-13
Support custom sort in serialization methods

ines:
Allows passing in a `sort_key`, a function that takes `(key, value)` tuples. Only applied to the top-level sections. The default alphabetic order isn't always the best.

## TODO

- [x] fix sorting for subsections
- [x] implement sort order setting
- [x] Does it make more sense to allow passing in a complete or incomplete list of section names instead? This is likely the most common use case: you want to specify an exact order of sections, rather than sorting them in reverse or by value or whatever.
- [x] Does it make more sense to set the sort key on `Config` initialization? This would make it easier to integrate it under the hood because it means you only have to define the sort key once and not every single time you call `to_str` in various places.

### DIFF
--- a/thinc/about.py
+++ b/thinc/about.py
@@ -1,2 +1,2 @@
-__version__ = "8.0.0a25"
+__version__ = "8.0.0a26"
 __release__ = True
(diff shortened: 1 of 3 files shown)


honnibal:
> Does it make more sense to allow passing in a complete or incomplete list of section names instead?

Yeah, I think it does. You always have the list when you go to serialize, so if you had a function it'd be trivial to get the ordered list. And yeah the basic use-case is knowing what order you want.

> Does it make more sense to set the sort key on Config initialization? 

That would be okay, yeah. Could a config set it on `from_str`? So you get the sections back in the order they came in.

PR #386, opened 2020-08-28
Listener flag

svlandeg:
Add `is_listener` flag in `model.attrs`, default `False`. If this flag is `True`, prevent calling `model.predict` directly for initialization/shape inference, as that won't work (the model is not independent)


honnibal:
Ugh, I'm not sure how to solve this but this will be a nightmare :(. If we accept this as the solution, then every layer has to think about listeners. We need to find another way.

PR #389, opened 2020-09-16
check for conflicting promise before merging

svlandeg:
There was an issue when merging a config with a default. If the original config has a promise (`@foo`) but the default has something else that doesn't match, the default values would still get added in the merged version. This was because the `continue` in the code started from the viewpoint of the default config. So I think we need both viewpoints - hence the second condition for another `continue`.

Hopefully the new unit tests make it clear. We need this behaviour to be able to implement different version of `training.corpus` in spaCy.


honnibal:
Looks good to me!

PR #393, opened 2020-09-24
fix model.from_dict

svlandeg:
Fixed required to make spacy's pretraining work
+ bump 8.0.0a35

### DIFF
--- a/thinc/model.py
+++ b/thinc/model.py
@@ -562,7 +562,7 @@ class Model(Generic[InT, OutT]):
             for param_name, value in msg["params"][i].items():
                 if value is not None:
                     value = node.ops.asarray(value)
-                node.set_param(param_name, value)
+                node.set_param(param_name, value.copy())
             for i, shim_bytes in enumerate(msg["shims"][i]):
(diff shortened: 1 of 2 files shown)


thinc/model.py
```diff
@@ -562,7 +562,7 @@ def from_dict(self, msg: Dict) -> "Model":
             for param_name, value in msg["params"][i].items():
                 if value is not None:
                     value = node.ops.asarray(value)
```
honnibal, line 564:
```suggestion
                    value = node.ops.asarray(value).copy()
```

thinc/model.py
```diff
@@ -562,7 +562,7 @@ def from_dict(self, msg: Dict) -> "Model":
...
-                node.set_param(param_name, value)
+                node.set_param(param_name, value.copy())
```
honnibal, line 565:
```suggestion
                node.set_param(param_name, value)
```

svlandeg: hehe yea, we just pushed the same commit ;-)

PR #405, opened 2020-10-05
check dims before calling blis.gemm

svlandeg:
Avoid a more cryptic Cython error downstream


thinc/backends/numpy_ops.pyx
```diff
@@ -75,6 +75,10 @@ class NumpyOps(Ops):
         return self.xp.zeros(shape, dtype=dtype)
 
     def gemm(self, np.ndarray x, np.ndarray y, *, np.ndarray out=None, trans1=False, trans2=False):
+        if x.ndim < 2:
```
honnibal, line 78: I think we might actually need _exactly_ two dimensions?

svlandeg: I thought so too, but wasn't 100% sure :|

PR #431, opened 2020-11-23
avoid initializing with Y if not necessary

svlandeg:
Only initialize the internal layer of `residual` with `Y` if the output dim `nO` is yet to be determined.

This is necessary to avoid wrong shape inferences stemming from the `chain` layer. This is all a bit brittle, and perhaps we want to rethink the whole shape inference mechanism at some point, but for now I think this works.

### DIFF
--- a/thinc/layers/parametricattention.py
+++ b/thinc/layers/parametricattention.py
@@ -53,7 +53,7 @@ def _get_attention(ops, Q, X, lengths):
     return attention, get_attention_bwd
 
 
-def _apply_attention(self, attention, X, lengths):
+def _apply_attention(ops, attention, X, lengths):
     output = X * attention
 
     def apply_attention_bwd(d_output):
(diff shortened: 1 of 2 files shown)


honnibal:
Can you point me to the code where this comes up? Is it in the rel model?

svlandeg:
The original issue that instigated this was https://github.com/explosion/spaCy/issues/6401#issuecomment-729655596. Some more background on the analysis of what goes wrong in that report is here: https://github.com/explosion/spaCy/pull/6431

This is about combining a `textcat` with a `TransformerListener`. The latter doesn't know its `nO` dimension straight away, so the `chain` layer gets into shape inference and tries to propagate the output `Y` through its layers: https://github.com/explosion/thinc/blob/master/thinc/layers/chain.py#L79. In this code, we specifically prevent propagating `Y` if `nO` is already set, because for the intermediate layers there's no guarantee that the given `Y` is actually compatible. But for the `residual` layer, it may be the case that its own `nO` is unset, but its encapsulated `layer` does have a proper `nO`. So in that case we need to prevent the `Y` being used to (re)initialize that encapsulated layer.

honnibal:
Ugh, this is ugly. Okay I'll merge this, but you're right that the policy is quite unclear.

PR #432, opened 2020-11-24
Propagate dims

svlandeg:
spaCy issue https://github.com/explosion/spaCy/issues/6318 pointed towards some issues with inferring dimensions. To better support chaining layers and shape inference, this PR:
- transfers dimensions from the encapsulated layer in the `with_X` type-transformation layers
- sets an appropriate `nO` and `nI` dimension on the `PyTorchLSTM`

### DIFF
--- a/thinc/layers/with_array.py
+++ b/thinc/layers/with_array.py
@@ -6,12 +6,11 @@ from ..types import Array2d, Floats2d, Padded, Ragged, ArrayXd
 from ..types import List2d
 
 
-ValT = TypeVar("ValT", bound=Array2d)
 SeqT = TypeVar("SeqT", bound=Union[Padded, Ragged, List2d, Array2d])
 
 
 @registry.layers("with_array.v1")
-def with_array(layer: Model[ValT, ValT], pad: int = 0) -> Model[SeqT, SeqT]:
+def with_array(layer: Model[Array2d, Array2d], pad: int = 0) -> Model[SeqT, SeqT]:
     """Transform sequence data into a contiguous 2d array on the way into and
     out of a model. Handles a variety of sequence types: lists, padded and ragged.
     If the input is a 2d array, it is passed through unchanged.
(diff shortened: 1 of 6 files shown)


thinc/layers/with_array.py
```diff
@@ -6,12 +6,11 @@
 from ..types import List2d
 
 
-ValT = TypeVar("ValT", bound=Array2d)
 SeqT = TypeVar("SeqT", bound=Union[Padded, Ragged, List2d, Array2d])
 
 
 @registry.layers("with_array.v1")
-def with_array(layer: Model[ValT, ValT], pad: int = 0) -> Model[SeqT, SeqT]:
```
honnibal, line 14: I think this change probably isn't quite right. The types are tricksy =/.

`Array2d` is a shorthand for `Union[Floats2d, Ints2d]`. If you ave `Model[Union[Floats2d, Ints2d], Union[Floats2d, Ints2d]`, that works out to:

```
Union[
  Model[Floats2d, Floats2d],
  Model[Ints2d, Ints2d],
  Model[Floats2d, Ints2d],
  Model[Ints2d, Floats2d]
]
```

That's not the right typing for the function, as only the first two are valid. That's why we use the type variable.

thinc/layers/with_cpu.py
```diff
@@ -10,16 +10,14 @@
...
-        dims["nO"] = layer.get_dim("nO")
-    return Model(f"with_cpu({layer.name})", forward, layers=[layer], ops=ops, init=init, dims=dims)
+    return Model(
+        f"with_cpu({layer.name})",
+        forward,
+        layers=[layer],
+        ops=ops,
+        init=init,
+        dims={name: layer.maybe_get_dim(name) for name in layer.dim_names},
+    )
```
honnibal, line 20: Very nice refactor here.

honnibal (changes requested):
I think that one type change needs to be reverted. Otherwise, looks good!

thinc/layers/with_array.py
svlandeg: Ah, yes, ofcourse! Thanks!

PR #434, opened 2020-11-30
Update setup

adrianeboyd:
* Update cythonize in setup
* Remove python version-related blis pins now that blis is easier to build from source in v0.7.4


adrianeboyd:
I think the updated version of of `preshed` (with https://github.com/explosion/preshed/pull/28) should fix this problem.

PR #436, opened 2020-12-02
Remove upper pin for torch

adrianeboyd:
`dataclasses` 0.8 should fix the python version compatibility issues for `torch` 1.7.0.


adrianeboyd:
Nevermind...

PR #438, opened 2020-12-02
Forbid torch==1.7.0

adrianeboyd:
The next release of torch after 1.7.0 should fix this, so just forbid 1.7.0 specifically.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,6 +19,9 @@ jobs:
       Python36Mac:
         imageName: 'macos-10.15'
         python.version: '3.6'
+      Python37Linux:
+        imageName: 'ubuntu-16.04'
+        python.version: '3.7'
       Python38Linux:
         imageName: 'ubuntu-16.04'
         python.version: '3.8'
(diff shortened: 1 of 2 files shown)


adrianeboyd:
I think the other alternative is pinning dataclasses to "==0.6".

adrianeboyd:
But ugh, I don't really like any of this. Maybe only do this for the affected versions of python?

bratao:
Pinning dataclasses looks more elegant then pinning he actual pytorch version

adrianeboyd:
I mixed up two different problems here.

Pinning `dataclasses` works for the reported install problem where no version of `dataclasses` is found (which I can't manage to replicate, it's some kind of pip + `-f` bug for particular versions of pip, maybe?).

The problem is that having `dataclasses` installed at all for python 3.7+ leads to the "`typing` has no attribute `_ClassVar` error" where you can't run the test suite as we do in the CI. Whether you get the built-in `dataclasses` or the one in `site-packages` seems to depend on exactly how/where things are imported and when only one of them works, this is no good.

It's extremely frustrating that the torch 1.7.0 wheels are broken like this and that they haven't published an update. There's no way for the thinc package to specify that a module be uninstalled or that a module is forbidden. There's no good way to express that the version is compatible but install it with `--no-deps`.

You can always install torch 1.7.0 by hand and uninstall `dataclasses` if you know what you're doing?

Again, argh.

I guess this is possible

```
torch>=1.5.0 ; python_version == '3.6'
torch>=1.5.0,!=1.7.0 ; python_version >= '3.6'
```

The other problem is that installing with `thinc[torch]` probably doesn't get the exact CPU/CUDA version you want anyway, so you're better off installing it separately following their install instructions, and then none of this is checked anyway, so on the one hand who cares about our specifications here, and on the other, why bother add this confusion at all.

honnibal:
Ugh.

Maybe `thinc[torch]` and `spacy[torch]` were a mistake? They're just nowhere near fine-grained enough to be helpful. Should we remove them?

adrianeboyd:
I don't think we have `spacy[torch]`? Probably we should just have `>=1.5.0` here and not care too much since it's just an extra.

The requirement that probably matters more is the one in `spacy-transformers`, since that will be checked on install no matter how you've installed `torch`.

adrianeboyd:
The CI should avoid 1.7.0 and use the `+cpu` versions, though, I can at least do that here.

PR #443, opened 2020-12-04
Update for pydantic 1.7 and python 3.9

adrianeboyd:
* Rename `Ragged._cumsums` to `Ragged.cumsums` so it's not flagged as an invalid field.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ wasabi>=0.4.0,<1.1.0
 catalogue>=0.2.0,<3.0.0
 ml_datasets==0.2.0a0
 # Third-party dependencies
-pydantic>=1.5.0,<1.7.0
+pydantic>=1.7.0
 numpy>=1.7.0
 # Backports of modern Python features
 dataclasses>=0.6,<1.0; python_version < "3.7"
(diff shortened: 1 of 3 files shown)


adrianeboyd:
There is maybe some way to get this to work as a private attribute or having it ignore fields that start with underscore, but I couldn't get it to work.

adrianeboyd:
Maybe this should be 1.7.1 instead because of the dataclasses arbitrary types thing?

honnibal:
What's the deal with this invalid field thing? You can't have private fields on a dataclass?

It's a bit annoying, because I don't want that to be part of the public interface =/

adrianeboyd:
I know, but if this is the only thing holding up python 3.9 support maybe we could at temporarily modify it. But maybe someone more knowledgeable (@tiangolo?) knows how to configure this correctly?

You get this warning from the `is_valid_field` check:

```
RuntimeWarning: fields may not start with an underscore, ignoring "_cumsums"
```

and then the:

```
X -> _cumsums   extra fields not permitted
```

I things like tried adding `underscore_attrs_are_private` to `_ArgModelConfig` and it didn't seem to do what I expected, but I don't understand all the details here.

tiangolo:
I'm sorry for the delay, my GitHub notifications are a mess, I have 1500 unread emails. :see_no_evil:

PR #447, opened 2020-12-15
Make dataclasses work with Pydantic >= 1.7

ines:
Don't use private dataclass fields for now (which conflict with the way Pydantic validates dataclasses in > 1.7).  This should make Thinc and by extension spaCy work with the latest Pydantic and Python 3.9.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,7 @@ wasabi>=0.4.0,<1.1.0
 catalogue>=0.2.0,<3.0.0
 ml_datasets==0.2.0a0
 # Third-party dependencies
-pydantic>=1.5.0,<1.7.0
+pydantic>=1.5.0,<1.8.0
 numpy>=1.7.0
 # Backports of modern Python features
 dataclasses>=0.6,<1.0; python_version < "3.7"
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -8,7 +8,7 @@ wasabi>=0.4.0,<1.1.0
 catalogue>=0.2.0,<3.0.0
 ml_datasets==0.2.0a0
 # Third-party dependencies
-pydantic>=1.5.0,<1.7.0
+pydantic>=1.5.0,<1.8.0
```
adrianeboyd, line 11: We want `>=1.7.0` here or python 3.9 will break.

ines: Okay, fair enough. (I guess technically we could use Python version-specific markers here but not sure if's worth the hassle 🤷‍♀️)

adrianeboyd: Wait, I think we had a discussion about a bug fix on their end and it should be `>=1.7.1`.

ines: Yes, good point, I remember that. I'll update the pin accordingly.

Issue #450, opened 2020-12-16
Error when using radam optimizer


bratao:
Hello, 
I have an error when using radam optimizer with thinc version v8.0.0rc2 and spaCy V3.

I´m just doing a regular NER pipeline training.
Python: 3.8
cuda 11.0
cupy-cuda110==8.2.0

If I use Adam or SGD, works great. 

```
Traceback (most recent call last):
  File "experimental/performance_benchmark.py", line 469, in train_spacy_multiple
    metrics = model.train(
  File "spacy/spacy_tagger.py", line 408, in train
    self.train_loop(
  File "spacy/spacy_tagger.py", line 359, in train_loop
    raise e
  File "spacy/spacy_tagger.py", line 342, in train_loop
    for batch, info, is_best_checkpoint in training_step_iterator:
  File "lib/python3.8/site-packages/spacy/training/loop.py", line 195, in train_while_improving
    proc.finish_update(optimizer)
  File "spacy/pipeline/trainable_pipe.pyx", line 241, in spacy.pipeline.trainable_pipe.TrainablePipe.finish_update
  File "lib/python3.8/site-packages/thinc/model.py", line 324, in finish_update
    param, grad = optimizer(
  File "lib/python3.8/site-packages/thinc/optimizers.py", line 232, in __call__
    weights, gradient = self._radam(
  File "lib/python3.8/site-packages/thinc/optimizers.py", line 279, in _radam
    exp_avg_sq += (1 - beta2) * (grad ** 2)
  File "cupy/core/core.pyx", line 1085, in cupy.core.core.ndarray.__iadd__
  File "cupy/core/_kernel.pyx", line 1059, in cupy.core._kernel.ufunc.__call__
  File "cupy/core/internal.pyx", line 353, in cupy.core.internal._broadcast_core
ValueError: operands could not be broadcast together with shapes (64, 12800), (64, 200), (12800,)
RUNTIME ERROR None
```

PR #452, opened 2020-12-23
Fix radam

bratao:
Radam was broken in latest version.

This PR fixes https://github.com/explosion/thinc/issues/450 by mimicking _adam behavior and to work with a 1d array.

Radam is much more resistant to the learning rate. + Lookahead = Ranger is the best optimizer for me!


honnibal:
Thanks!

PR #466, opened 2021-02-05
avoid initializing with Y if X is set

svlandeg:
Inspired by [this issue](https://github.com/explosion/spaCy/issues/6946). The "trouble" is being caused by the shape inference code in `chain`, but it can't hurt to add some additional robustness into `layernorm` to avoid using `Y` if not necessary.

### DIFF
--- a/thinc/layers/layernorm.py
+++ b/thinc/layers/layernorm.py
@@ -44,7 +44,7 @@ def init(
         X_width = get_width(X)
         model.set_dim("nI", X_width)
         model.set_dim("nO", X_width)
-    if Y is not None:
+    elif Y is not None:
         Y_width = get_width(Y)
         model.set_dim("nI", Y_width)
         model.set_dim("nO", Y_width)


svlandeg:
How is this little PR suddenly triggering `mypy` issues? mypy 0.8 was out already 15 days ago, and `master` was still building correctly 13 days ago...?

adrianeboyd:
This is probably also related to the newest version of numpy, but I'm not sure why it looks different than in #465, which didn't fail until the test suite.

svlandeg:
Hm, I guess we'll see when the other PR is merged whether it also fixes this one 🤞

justindujardin:
I'm late to this conversation, but you're right, the failures were related to the latest numpy release. 

I added some `# type: ignore` [comments in the test suite to fix them](https://github.com/explosion/thinc/pull/465/commits/1be4a17f5dcbdc0bf58a4a7e224fb30c0b75c100)

The latest numpy release [has its own type hints](https://numpy.org/doc/stable/release/1.20.0-notes.html#numpy-is-now-typed), so our usages in thinc don't completely line up.

At some point we could probably benefit from using the built-in numpy types in thinc, but for now a few comments did the trick. 😅

PR #469, opened 2021-02-16
Add check for contextvars ops vs. thread-local storage

adrianeboyd:
In order to detect cases in Jupyter notebooks where `contextvars` are
not preserved across cells, additionally save the ops settings in
thread-local storage as in thinc v7.

The new method `contextvars_eq_thread_ops` returns whether both are the
same ops type. The thread-local storage is not intended for general
use, only for this check in a restricted set of configurations.


adrianeboyd:
When/how/where should thinc test for this in notebooks?

PR #470, opened 2021-02-16
Reset torch tensor type in require_cpu

adrianeboyd:


honnibal:
I'll try reverting this to see if it's behind the recent failures.

PR #473, opened 2021-02-23
Revert "Reset torch tensor type in require_cpu"

honnibal:
Reverts explosion/thinc#470


honnibal:
We're getting failures so I just want to see if this passes

adrianeboyd:
I don't think this is the culprit, so closing for now.

PR #489, opened 2021-04-19
Remove codevoc from CI

svlandeg:
Remove codevoc from CI


azure-pipelines.yml
```diff
@@ -95,6 +95,3 @@ jobs:
       python -m ipykernel install --name thinc-notebook-tests --user
       python -m pytest --pyargs thinc --cov=thinc --cov-report=xml
```
adrianeboyd, line 96: I think we can also remove the `--cov` bits here and `pytest-cov` from `requirements.txt`?

svlandeg: Could do - but we could also use `pytest-cov` without `codevoc` and get a quick stats overview in the CI output?

honnibal: Leaving the coverage report in is fine. It's just the external service we want to drop.

honnibal (approved):

PR #490, opened 2021-04-21
Add tuplify

polm:
Tuplify replicates the input so that it can be modified separately by
downstream layers.

This needs tests and probably other metadata settings added still.

### DIFF
--- a/thinc/layers/tuplify.py
+++ b/thinc/layers/tuplify.py
@@ -10,12 +10,19 @@ MidT = TypeVar("MidT")
 
 @registry.layers("tuplify.v1")
 def tuplify(layer1: Model, layer2: Model, *layers) -> Model:
+    """Send a separate copy of the input to each child layer, and join the
+    outputs of the children into a tuple on the way out.
+
+    Typically used to provide both modified data and the original input to a
+    downstream layer.
+    """
+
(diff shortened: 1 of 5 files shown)


thinc/layers/tuplify.py
```diff
@@ -0,0 +1,44 @@
...
+from ..model import Model
+from ..config import registry
+
+InT = TypeVar("InT")
+OutT = TypeVar("OutT")
+MidT = TypeVar("MidT")
+
+
+@registry.layers("tuplify.v1")
+def tuplify(layer1: Model, layer2: Model, *layers) -> Model:
```
honnibal, line 12:
```suggestion
def tuplify(layer1: Model[InT, Any], layer2: Model[InT, Any], *layers) -> Model[InT, Tuple]:
```

polm:
So working with this more there are actually a few problems that should be fixed before it's merged.

1. There needs to be init handling for sublayers.
2. If the input is a generator that has to be handled appropriately.

honnibal:
Regarding 2, I don't think we need to support inputs as generators. Our own batching utilities will always be producing a list or a tuple or an array, and I don't think it's reasonable for people to expect a generator as a batch to work properly.

PR #491, opened 2021-04-21
Resizable container & force set_dim

svlandeg:
- Introduce `resizable` container layer with a `resize` method that can be used in `model.attrs["resize_output"]` that is checked in spaCy's `pipe.is_resizable` method. See also https://github.com/explosion/spaCy/pull/7862
- Add argument `force` that is `False` by default to allow overriding an existing dim value with `model.set_dim()` (to be used with caution ;-))


thinc/model.py
```diff
@@ -176,14 +176,14 @@ def get_dim(self, name: str) -> int:
         else:
             return value
 
-    def set_dim(self, name: str, value: int) -> None:
+    def set_dim(self, name: str, value: int, force: bool = False) -> None:
```
honnibal, line 179:
```suggestion
    def set_dim(self, name: str, value: int, *, force: bool = False) -> None:
```

We should make this keyword only I think?

honnibal:
Do we need the `force` flag on `set_dim` still? If we don't need it for this feature I wouldn't add it, as it does cause problems if parameters are already set.

svlandeg:
Yea I think it's still needed in the "container" layer to allow for proper dimension inference...

PR #494, opened 2021-05-03
🐛 Add update_forward_refs with local models

Kludex:
Hi there! :wave:

Let me start with: I don't know exactly how this start failing. :sweat_smile: 

Maybe you'll understand better looking at the explanation: https://github.com/samuelcolvin/pydantic/issues/2678#issuecomment-821437650

The solution I proposed is the one presented in the link above. An alternative could be: 

```python
ArgModel.update_forward_refs(**globals())
```

(but you'll get more than you need... and I'm not sure it works in all cases)

### DIFF
--- a/thinc/util.py
+++ b/thinc/util.py
@@ -48,10 +48,8 @@ try:  # pragma: no cover
 except ImportError:  # pragma: no cover
     has_mxnet = False
 
-from .types import ArrayXd, ArgsKwargs, Ragged, Padded, FloatsXd, IntsXd
-from .types import Floats1d, Floats2d, Floats3d, Floats4d
-from .types import Ints1d, Ints2d, Ints3d, Ints4d
-
+from .types import ArrayXd, ArgsKwargs, Ragged, Padded, FloatsXd, IntsXd  # noqa: E402
+from . import types  # noqa: E402
 
 def get_array_module(arr):  # pragma: no cover
     if is_cupy_array(arr):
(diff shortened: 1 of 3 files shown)


ines:
Omg, thanks! 🙏🎉 (Also, I feel kinda stupid for not seeing this... I even vaguely remember thinking "hmm, I wish there was a way to just tell it what the modules are 🤔" hahaha)

thinc/util.py
```diff
@@ -437,6 +439,16 @@ def validate_fwd_input_output(
...
     ArgModel = create_model("ArgModel", **sig_args)
+    ArgModel.update_forward_refs(
+        Floats2d=Floats2d,
+        Floats1d=Floats1d,
+        Floats3d=Floats3d,
+        Floats4d=Floats4d,
+        Ints1d=Ints1d,
+        Ints2d=Ints2d,
+        Ints3d=Ints3d,
+        Ints4d=Ints4d
```
ines, line 450: I wonder if we should just pass in all the `**types` here? Because in theory, there are more types that the forward functions could use. And using the `globals()` is more vague, and makes it less obvious that the imports in `util.py` matter.

ines: Okay, it can't be `**types` because that's a module, but you get the idea 😅

Kludex: I see. I guess it's fine now? :kissing:

ines: Yeah, I think the solution I came up with works and does what we want 🎉 Thanks again!

PR #495, opened 2021-05-05
Avoid concatting `None` gradient

svlandeg:
Originally opened to address https://github.com/explosion/spaCy/pull/8012

That was solved within `spacy`. This PR now only involves a more general implementation of the current functionality.

### DIFF
--- a/thinc/layers/concatenate.py
+++ b/thinc/layers/concatenate.py
@@ -60,14 +60,15 @@ def _array_forward(
         dY = model.ops.as_contig(d_output[:, : widths[0]])
         dX = callbacks[0](dY)
         start = widths[0]
+        add_gradients = hasattr(dX, "__add__") or hasattr(dX, "__iadd__")
+        add_gradients_data = hasattr(dX, "data") and (hasattr(dX.data, "__add__") or hasattr(dX.data, "__iadd__"))
         for bwd, width in zip(callbacks[1:], widths[1:]):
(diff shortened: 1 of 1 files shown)


thinc/layers/concatenate.py
```diff
@@ -62,10 +62,12 @@ def backprop(d_output: Array2d) -> InT:
...
-            if isinstance(dX, Ragged):
-                dX.data += bwd(dY).data
-            else:
-                dX += bwd(dY)
+            gradient = bwd(dY)
+            if gradient is not None:
+                if isinstance(dX, Ragged):
+                    dX.data += gradient.data
+                else:
+                    dX += gradient
```
honnibal, line 70: How about something like:

```suggestion
            gradient = bwd(dY)
            if add_gradients:
                dX += gradient
            elif add_gradients_data_attr:
                dX.data += gradient.data
```

And before the loop we do:

```
add_gradients = hasattr(dX, "__add__") or hasattr(dX, "__iadd__")
add_gradients_data = hasattr(dX, "data") and (hasattr(dX.data, "__add__") or hasattr(dX.data, "__iadd__"))
```

svlandeg: That won't solve the problem of concatenating `None` though. The problem is that one of the callbacks (the `Tok2VecListener`) doesn't return a gradient if the listener is used multiple times in the pipeline.

PR #499, opened 2021-05-18
Update pydantic requirement

adrianeboyd:
Increase upper pin to `<1.9.0`. Forbid versions with security vulnerability:
https://github.com/samuelcolvin/pydantic/security/advisories/GHSA-5jqp-qgf6-3pvh


svlandeg:
I guess we need to do `spacy` too?

PR #508, opened 2021-06-10
Restrict cython to <3.0

adrianeboyd:


adrianeboyd:
The failures are related to `mypy`, will rerun after #509.

PR #514, opened 2021-06-24
Include final ngram in NumpyOps.ngrams

adrianeboyd:
* Additionally return empty lists for invalid n-gram lengths

### DIFF
--- a/thinc/backends/numpy_ops.pyx
+++ b/thinc/backends/numpy_ops.pyx
@@ -345,11 +345,13 @@ class NumpyOps(Ops):
         return weights, gradient, mom1, mom2
 
     def ngrams(self, int n, const uint64_t[::1] keys):
+        if n < 1:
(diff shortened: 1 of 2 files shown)


thinc/backends/numpy_ops.pyx
```diff
@@ -345,11 +345,13 @@ class NumpyOps(Ops):
         return weights, gradient, mom1, mom2
 
     def ngrams(self, int n, const uint64_t[::1] keys):
+        if n < 1:
+            return self.alloc((0,), dtype="uint64")
         keys_ = <uint64_t*>&keys[0]
-        length = max(0, keys.shape[0]-n)
+        length = max(0, keys.shape[0]-(n-1))
```
honnibal, line 351: There's always one more off-by-one error than I expect :(. Thanks!

PR #521, opened 2021-08-11
Fix numpy_ops gemm output semantics when BLIS is used

danieldk:
I am reading various parts of the thinc source code to see how things work, I found a small inconsistency in the numpy `gemm` op:

The `out` keyword argument of the `gemm` op specifies an output array. However, the semantics were different depending on whether BLIS is used:

* `use_blis==False`: the values of out array are overwritten.
* `use_blis==True`: the values of out are added to the result.

With this change, the values of `out` are also overwritten with `use_blis=True`.

### DIFF
--- a/thinc/backends/numpy_ops.pyx
+++ b/thinc/backends/numpy_ops.pyx
@@ -88,7 +88,7 @@ class NumpyOps(Ops):
         y = self.as_contig(y)
         if out is not None:
             out = self.as_contig(out)
-        return blis.py.gemm(x, y, out=out, trans1=trans1, trans2=trans2)
+        return blis.py.gemm(x, y, out=out, trans1=trans1, trans2=trans2, beta=0.)
 
     def relu(self, np.ndarray X, inplace=False):
         cdef np.ndarray out = X if inplace else X.copy()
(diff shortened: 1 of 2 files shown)


thinc/backends/numpy_ops.pyx
```diff
@@ -88,7 +88,7 @@ class NumpyOps(Ops):
         y = self.as_contig(y)
         if out is not None:
             out = self.as_contig(out)
-        return blis.py.gemm(x, y, out=out, trans1=trans1, trans2=trans2)
+        return blis.py.gemm(x, y, out=out, trans1=trans1, trans2=trans2, beta=0.)
```
svlandeg, line 91: Hey Daniel, I'm a little confused by this PR. I don't see a change in behaviour for the `out`/return value as you mentioned in the PR description?

danieldk: Sorry if the commit message was a bit unclear. Here gemm computes:

out = x @ y + beta * out

And by default [`beta = 1.0`](https://github.com/explosion/cython-blis/blob/e42ef40b7af784a18f08fe6d3cc35680988613d2/blis/py.pyx#L66). So, without this change, this computes

out = x @ y + 1.0 * out

So, `out` is added to the matrix multiplication of `x` and `y`. This is not the case when BLIS is not used.

svlandeg: Ah, interesting, nice catch! It looks like we got lucky and didn't accidentally introduce bugs because of this, as it looks like `out` is typically `None` (converted to array of zeros) when we call this. Still, definitely good to fix his.

PR #525, opened 2021-08-25
🐛 Fix Mypy plugin crash on variadic arguments

### LINKED ISSUE #519: Mypy plugin crash analyzing `concatenate` called with `*args`
While working on [some changes to SpaCy's type signatures](https://github.com/explosion/spaCy/pull/8773), I noticed that running Mypy on the SpaCy code base produces the following error:

```python
Traceback (most recent call last):
  File "mypy/checkexpr.py", line 3911, in accept
  File "mypy/nodes.py", line 1558, in accept
  File "mypy/checkexpr.py", line 271, in visit_call_expr
  File "mypy/checkexpr.py", line 353, in visit_call_expr_inner
  File "mypy/checkexpr.py", line 858, in check_call_expr_with_callee_type
  File "mypy/checkexpr.py", line 917, in check_call
  File "mypy/checkexpr.py", line 1029, in check_callable_call
  File "mypy/checkexpr.py", line 729, in apply_function_plugin
  File ".../spaCy/.env/lib/python3.9/site-packages/thinc/mypy.py", line 31, in function_hook
    return get_reducers_type(ctx)
  File ".../spaCy/.env/lib/python3.9/site-packages/thinc/mypy.py", line 81, in get_reducers_type
    ctx.default_return_type.type, [arg_types[0].args[0], arg_types[0].args[1]]
IndexError: tuple index out of range
```

Based on some pdb debugging (using `mypy --pdb`) and reading the Thinc Mypy plugin code, I was able to come up with this minimal test case:

```
from thinc.api import chain, concatenate, Relu, Softmax

n_hidden = 32
dropout = 0.2

model1 = chain(
    Relu(nO=n_hidden, dropout=dropout),
    Relu(nO=n_hidden, dropout=dropout),
    Softmax()
)

model2 = chain(
    Relu(nO=n_hidden, dropout=dropout),
    Relu(nO=n_hidden, dropout=dropout),
    Softmax()
)

model3 = concatenate(*[model1, model2])
```

In this example, the use of variadic arguments (`*args`) is necessary for the error to occur.

I researched Mypy's typing system and looked into the Mypy plugin code, but wasn't able to figure out what the expected behavior would be here. The root problem is that [these lines](https://github.com/explosion/thinc/blob/a125b8cf6a518bf61a418dbd077d8849e8c30167/thinc/mypy.py#L53-L56), which build the `arg_types` list, result in only a single entry in `arg_types`:

```python
[builtins.list[thinc.model.Model*[thinc.types.Floats2d, thinc.types.Floats2d]]]
```

This is because the `get_reducers_type` function doesn't take into consideration the value of `ctx.arg_kinds`:

```python
(Pdb) p ctx
FunctionContext(..., arg_kinds=[[2]], ...)
```

(based on [this lookup table](https://github.com/python/mypy/blob/9f409c3d89c7fa65eff8cf67883473487e6effd4/mypy/nodes.py#L1525-L1537), 2 means `ArgKind.ARG_STAR`)

At this point I've probably researched the issue enough to fix it myself :joy: My proposal for how to fix this is to expand variadic arguments into two "fake arguments" during type-checking (in `get_reducers_type`). This would be an approximation of all of the possible inputs for the function, so it may not be totally sound, but might be better than the current crash?

Alternatively, this could be detected as a special-case and `Model[Any]` could be returned (or some generic indicator like that) 🙂

Thanks for making thinc! 😄

connorbrinton:
These changes fix a crash that would occur in the Thinc Mypy plugin whenever a Thinc API function was called with variadic arguments.

Previously the Thinc Mypy plugin would treat any argument with a generic type as a `Model` instance and attempt to infer a more specific type for the return type of the function call. However, the Thinc Mypy plugin was intended to only be used on functions that exclusively accept and return `Model` instances. Since the plugin accepted overly broad input, it would sometimes end up trying to access the second generic type argument for a generic type instance with only a single argument. This resulted in the plugin crashing.

This specifically manifested itself whenever variadic arguments were passed to a Thinc API function. Mypy passes the type of variadic arguments as `List[Model[InT, OutT]]`. The Thinc Mypy plugin would look at this type and attempt to treat it as an unwrapped `Model` type since the plugin would allow any generic type as an argument type annotation. Later on the plugin would attempt to retrieve the output type of the "Model". Since `List` accepts only a single generic argument, attempting to access `type.args[1]` would cause the crash.

These changes fix this by restricting the input on which the Thinc Mypy plugin operates. Specifically each argument and the return type must:
* Be a generic type `Instance`,
* Refer to the `thinc.model.Model` type and
* Have two generic type arguments

Notably these restrictions prevent the Thinc Mypy plugin from working on subclasses of `thinc.model.Model`. While this plugin previously had some support for `Model` subclasses, this behavior was not documented and unintentional. `Model` subclasses were only properly handled if their first two generic type arguments had the same meaning as `Model`'s generic type arguments.

Support for `Model` subclasses may be restored in the future through the use of Mypy's `map_instance_to_supertype` function. I just haven't figured out how to use it yet 😅

Fixes #519

### DIFF
--- a/thinc/mypy.py
+++ b/thinc/mypy.py
@@ -144,6 +144,8 @@ def get_reducers_type(ctx: FunctionContext) -> Type:
     for arg_type in itertools.chain(*ctx.arg_types):
         # TODO: Use `map_instance_to_supertype` to map subtypes to `Model` instances.
         assert isinstance(arg_type, Instance)
+        assert arg_type.type.fullname == thinc_model_fullname
+        assert len(arg_type.args) == 2
         arg_types.append(arg_type)
 
     # Collect neighboring pairs of arguments and their types
--- a/thinc/mypy.py
+++ b/thinc/mypy.py
@@ -117,8 +117,7 @@ def get_reducers_type(ctx: FunctionContext) -> Type:
 
     # Verify that the callable returns a `thinc.model.Model`
     # TODO: Use `map_instance_to_supertype` to map subtypes to `Model` instances.
-    # I haven't implemented this myself because I wasn't able to figure out how to look up the
-    # `TypeInfo` for a class outside of the module being type-checked
+    # (figure out how to look up the `TypeInfo` for a class outside of the module being type-checked)
     callee_return_type = callee_node_type.ret_type
     assert isinstance(callee_return_type, Instance)
     assert callee_return_type.type.fullname == thinc_model_fullname
--- a/thinc/mypy.py
+++ b/thinc/mypy.py
@@ -35,6 +35,73 @@ def function_hook(ctx: FunctionContext) -> Type:
 
 
 def get_reducers_type(ctx: FunctionContext) -> Type:
+    """
+    Determine a more specific model type for functions that combine models.
+
+    This function operates on function *calls*. It analyzes each function call
+    by looking at the function definition and the arguments passed as part of
+    the function call, then determines a more specific return type for the
+    function call.
+
+    This method accepts a `FunctionContext` as part of the Mypy plugin
+    interface. This function context provides easy access to:
+    * `args`: List of "actual arguments" filling each "formal argument" of the
+      called function. "Actual arguments" are those passed to the function
+      as part of the function call. "Formal arguments" are the parameters
+      defined by the function definition. The same actual argument may serve
+      to fill multiple formal arguments. In some cases the relationship may
+      even be ambiguous. For example, calling `range(*args)`, the actual
+      argument `*args` may fill the `start`, `stop` or `step` formal
+      arguments, depending on the length of the list.
+
+      The `args` list is of length `num_formals`, with each element
+      corresponding to a formal argument. Each value in the `args` list is a
+      list of actual arguments which may fill the formal argument. For
+      example, in the function call `range(*args, num)`, `num` may fill the
+      `start`, `end` or `step` formal arguments depending on the length of
+      `args`, so type-checking needs to consider all of these possibilities.
+    * `arg_types`: Type annotation (or inferred type) of each argument. Like
+      `args`, this value is a list of lists with an outer list entry for each
+      formal argument and an inner list entry for each possible actual
+      argument for the formal argument.
+    * `arg_kinds`: "Kind" of argument passed to the function call. Argument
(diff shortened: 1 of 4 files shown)


thinc/mypy.py
```diff
@@ -35,28 +35,129 @@ def function_hook(ctx: FunctionContext) -> Type:
...
+    assert isinstance(callee, NameExpr)
+    callee_node = callee.node
+    assert isinstance(callee_node, (FuncDef, Decorator))
+    callee_node_type = callee_node.type
+    assert isinstance(callee_node_type, CallableType)
+
+    # Verify that the callable returns a `thinc.model.Model`
+    # TODO: Use `map_instance_to_supertype` to map subtypes to `Model` instances.
+    # I haven't implemented this myself because I wasn't able to figure out how to look up the
+    # `TypeInfo` for a class outside of the module being type-checked
```
svlandeg, line 121:
```suggestion
    # TODO: Use `map_instance_to_supertype` to map subtypes to `Model` instances.
    # (figure out how to look up the `TypeInfo` for a class outside of the module being type-checked)
```

thinc/tests/regression/issue519/test_issue519.py
```diff
@@ -0,0 +1,26 @@
...
+import pytest
+
+
+@pytest.mark.slow
+def test_issue519():
+    """
+    Test ability of Thinc mypy plugin to handle variadic arguments.
+
+    This test usually takes around 45 seconds (at least on my machine), so
+    I've marked it as slow 🙂
```
svlandeg, line 17:
```suggestion
    This test can take up to 45 seconds, and is thus marked as slow.
```

svlandeg (approved):
Hi @connorbrinton,

This all looks pretty good, I particularly appreciate the detailed explanations, code comments and additional unit test :-) I made a few small nitpicking suggestions for some of the comments, as it makes sense to avoid statements with "I/my" in the code base, as you don't know later on who wrote what (and it shouldn't really matter either).

connorbrinton:
@svlandeg Thanks for those suggestions, they've been committed 🙂

svlandeg:
We'll need to bump the Thinc version and get this released before we can progress with https://github.com/explosion/spaCy/pull/8773, right?

PR #527, opened 2021-08-31
Allow config overrides to add new keys

ines:
Still need to test this against spaCy to check if we need to adjust any tests here. Also a bit worried about potential unitended side-effects (or some important reason why we included this feature that I forgot).

Related: https://github.com/explosion/spacy-transformers/pull/283#issuecomment-903513318

---

The config system currently only allows passing on overrides for existing settings, not new keys. For instance, given a config like this:

```ini
[section]
foo = 1
bar = 2
```

Passing in the overrides `{"section.foo": 3}` is valid, whereas `{"section.baz": 3}` would raise an error. In a lot of scenarios, this makes sense, because you'd expect the config to be complete and describing all available settings – so anything unexpected is likely a problem. But there are other cases where the config is more flexible (e.g. a dictionary with arbitrary keys passed to a function as an argument).

### DIFF
--- a/thinc/about.py
+++ b/thinc/about.py
@@ -1,2 +1,2 @@
-__version__ = "8.0.8"
+__version__ = "8.0.9"
 __release__ = True
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -352,7 +352,8 @@ class Config(dict):
             if "." not in key:
                 raise ConfigValidationError(errors=err, title=err_title)
             section, option = key.rsplit(".", 1)
-            if section not in config or option not in config[section]:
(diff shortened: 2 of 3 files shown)


svlandeg:
I had a look into the history and apparently you had it like this in your [original PR](https://github.com/explosion/thinc/pull/374), but then changed it: https://github.com/explosion/thinc/pull/374/commits/ac1e9cabf3bb05358e390b304fe47e0e769063a1

There wasn't another reason discussed other than structuring the code differently.

I wonder whether it would make sense to warn when a new key is defined though. To guard the user against making a typo, accidentally adding a new key and not updating the old one as they would have expected.

PR #528, opened 2021-08-31
try unpinning pip

svlandeg:
Testing the CI to see whether we can unpin `pip`, cf https://github.com/explosion/thinc/pull/493


svlandeg:
Hooray, it magically works again 🤷

PR #529, opened 2021-09-01
PyTorchShim: add support for mixed-precision

danieldk:
This change adds support for mixed-precision training and prediction to
the PyTorch shim. Mixed-precision support consists of two parts:

- Autocasting of the forward pass. This changes whitelisted ops to run
  in half precision rather than single precision.
- Gradient scaling. Gradients underflow quickly in half-precision.
  Gradient scaling increases the magnitude of the gradients during
  backprop to avoid underflow.

A separate class, PyTorchGradScaler is introduced to perform gradient
scaling and store the associated state.

---

I hope that I correctly implemented multi-GPU training, but I couldn't test it with my current single GPU workstation.

### DIFF
--- a/thinc/layers/pytorchwrapper.py
+++ b/thinc/layers/pytorchwrapper.py
@@ -34,6 +34,46 @@ def PyTorchWrapper(
     pytorch_model,
     convert_inputs: Optional[Callable] = None,
     convert_outputs: Optional[Callable] = None,
+) -> Model[Any, Any]:
+    """Wrap a PyTorch model, so that it has the same API as Thinc models.
+    To optimize the model, you'll need to create a PyTorch optimizer and call
+    optimizer.step() after each batch. See examples/wrap_pytorch.py
+
+    Your PyTorch model's forward method can take arbitrary args and kwargs,
+    but must return either a single tensor as output or a tuple. You may find the
+    PyTorch register_forward_hook helpful if you need to adapt the output.
+
+    The convert functions are used to map inputs and outputs to and from your
+    PyTorch model. Each function should return the converted output, and a callback
+    to use during the backward pass. So:
+
+        Xtorch, get_dX = convert_inputs(X)
+        Ytorch, torch_backprop = model.shims[0](Xtorch, is_train)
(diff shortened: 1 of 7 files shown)


thinc/layers/pytorchwrapper.py
```diff
@@ -34,6 +34,8 @@ def PyTorchWrapper(
     pytorch_model,
     convert_inputs: Optional[Callable] = None,
     convert_outputs: Optional[Callable] = None,
+    mixed_precision: bool = False,
+    grad_scaler: Optional[PyTorchGradScaler] = None,
```
svlandeg, line 38: We need to increment the versioning here to `PyTorchWrapper.v2` to make sure it's clear that these two parameters are only supported in the newer version and from a specific Thinc release. Both definitions can be here in the same file though.

thinc/shims/pytorch_grad_scaler.py
```diff
@@ -0,0 +1,133 @@
...
+                device_tensors, found_inf_device, inv_scale_device
+            )
+
+            self._found_inf += found_inf_device.to(self._found_inf.device)
+
+        return bool(self._found_inf != 0)
+
+    def update(self):
+        """
+        Update the the scale factor and clear information about infinities.
```
svlandeg, line 116:
```suggestion
        Update the scale factor and clear information about infinities.
```

svlandeg (commented):
Some first quick comments:

thinc/layers/pytorchwrapper.py
danieldk: Ah, right, should be fixed now.

honnibal:
It's hard to love the mechanism here. 🤔  It's just sort of disappointing to need this helper class.

I don't really have a better suggestion though, and it doesn't affect the rest of the API, it's specific to the PyTorchShim. So I'm okay with merging this as-is. I'd definitely try to find some other way of doing things if it were on the main `Model` class though!

svlandeg:
We can have it as an ondocumented beta feature for now, allowing us to still revisit the implementation details at some point if we want. So I'll go ahead and merge!

PR #530, opened 2021-09-01
Add wrap_model_recursive

danieldk:
This is a simple wrapper function that recursively wraps a Model and its layers recursively.

Useful for e.g. recursively wrapping forward/backprop passes in NVTX ranges.

### DIFF
--- a/thinc/tests/model/test_model.py
+++ b/thinc/tests/model/test_model.py
@@ -1,25 +1,10 @@
 import pytest
 import threading
 import time
-from thinc.api import (
-    CupyOps,
-    prefer_gpu,
-    Linear,
-    Dropout,
-    Model,
-    Shim,
-    change_attr_values,
-)
-from thinc.api import (
-    set_dropout_rate,
-    chain,
-    concatenate,
-    with_debug,
-    wrap_model_recursive,
-    Relu,
-    Softmax,
-    Adam,
-)
+from thinc.api import Adam, CupyOps, Dropout, Linear, Model, Relu
+from thinc.api import Shim, Softmax, chain, change_attr_values
+from thinc.api import concatenate, prefer_gpu, set_dropout_rate
+from thinc.api import with_debug, wrap_model_recursive
 import numpy
 
 from ..util import make_tempdir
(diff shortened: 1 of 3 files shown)


thinc/tests/model/test_model.py
```diff
@@ -443,3 +452,39 @@ def test_model_gpu():
...
+    n_debug = 0
+    for model in concat_debug.walk():
+        if model.name.startswith("debug"):
+            n_debug += 1
+
+    # There should be 5 unique debug wrappers:
+    # * Around concatenate. (= 1)
+    # * One around each chain in concatenate. (= 2)
+    # * One around each relu in the chain. (= 2)
+    assert n_debug == 5
```
svlandeg, line 490: Love these tests. You could extend the `test_recursive_double_wrap` one with 
```
assert concat_debug.layers[0].layers[0].layers[0].layers[0].name == "debug(relu)"
assert concat_debug.layers[0].layers[0].layers[0].layers[1].name == "debug(relu)"
assert concat_debug.layers[0].layers[1].layers[0].layers[0].name == "debug(relu)"
assert concat_debug.layers[0].layers[1].layers[0].layers[1].name == "debug(relu)"
```
to ensure that the `seen` parameter works correctly. Because if not, these will be `debug(debug(relu))` instead of just `debug(relu)`.

svlandeg (approved):
Nice. I'm kind of surprised we didn't have this before yet - it'll certainly be useful for debugging etc.

thinc/tests/model/test_model.py
```diff
@@ -10,7 +10,16 @@
...
+from thinc.api import (
+    set_dropout_rate,
+    chain,
+    concatenate,
+    with_debug,
+    wrap_model_recursive,
+    Relu,
+    Softmax,
+    Adam,
+)
```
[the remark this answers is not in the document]
danieldk: Fixed, I also changed the preceding import which also had the same formatting.

thinc/tests/model/test_model.py
danieldk: Great idea! I have added these.

PR #533, opened 2021-09-07
Fix default ops for get_array_ops

adrianeboyd:
`get_array_ops` should default to `NumpyOps` rather than the current ops so that numpy arrays are handled correctly.


adrianeboyd:
I was worried that defaulting to `NumpyOps` would cause the ops to switch from `AppleOps` to `NumpyOps` unintentionally, but this is clearly incorrect for numpy arrays and current `CupyOps`.

It looks like `get_array_ops` is only used in a fairly limited context, but we still may have to keep an eye out in the future for interactions with `AppleOps`.

PR #535, opened 2021-09-09
Improve wrapping in with_{debug,nvtx_range}

danieldk:
This PR combines two changes:

1. In `wrap_model_recursive`, also make sure that node references are pointing to the wrapped nodes.
2. Pass through dims, refs, attrs, and ops, so that these can be queried when a layer is wrapped.

### DIFF
--- a/thinc/model.py
+++ b/thinc/model.py
@@ -380,6 +380,24 @@ class Model(Generic[InT, OutT]):
                 if ref is not None and ref not in tree:
                     node.set_ref(name, None)
 
+    def replace_node(self, old: "Model", new: "Model") -> bool:
+        """Replace a node anywhere it occurs within the model. Returns a boolean
+        indicating whether the replacement was made."""
+        seen = False
+
+        for node in list(self.walk()):
+            if node is old:
+                seen = True
+            else:
+                node._layers = [
+                    new if layer is old else layer for layer in node._layers
+                ]
+                for name in node.ref_names:
+                    if node.get_ref(name) is old:
+                        node.set_ref(name, new)
+
+        return seen
+
     def get_gradients(self) -> Dict[Tuple[int, str], Tuple[FloatsXd, FloatsXd]]:
         """Get non-zero gradients of the model's parameters, as a dictionary
         keyed by the parameter ID. The values are (weights, gradients) tuples.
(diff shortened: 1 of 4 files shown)


thinc/model.py
```diff
@@ -363,6 +363,22 @@ def walk(self) -> Iterable["Model"]:
...
 
+    def replace_layer(self, old: "Model", new: "Model") -> "Model":
+        """Replace one of the model's immediate children with a new layer. If the
+        layer is not found within the children, an IndexError is raised. Returns
+        the model."""
+
+        if not old in self.layers:
+            raise IndexError("Cannot replace non-existing layer")
+
+        self._layers = [new if layer == old else layer for layer in self.layers]
```
honnibal, line 374: I think we probably want `is` here, rather than `==`

thinc/model.py
```diff
@@ -363,6 +363,22 @@ def walk(self) -> Iterable["Model"]:
...
+
+        for (name, item) in self._refs.items():
+            if item == old:
```
honnibal, line 377:
```suggestion
            if item is old:
```

thinc/model.py
honnibal: I first thought we'd use `layers.index()`, but that would be wrong -- the layers could have the node in there twice. So your implementation is correct.

thinc/model.py
```diff
@@ -363,6 +363,22 @@ def walk(self) -> Iterable["Model"]:
...
+        """Replace one of the model's immediate children with a new layer. If the
+        layer is not found within the children, an IndexError is raised. Returns
+        the model."""
+
+        if not old in self.layers:
+            raise IndexError("Cannot replace non-existing layer")
+
+        self._layers = [new if layer == old else layer for layer in self.layers]
+
+        for (name, item) in self._refs.items():
```
honnibal, line 376: I was going to say "Oh, we should walk the tree and replace the reference", but then I realised you can't have a reference higher up in the tree, only further down.

honnibal:
Looking at this now, do you think the method would be more useful if it replaced the layer anywhere in the tree?

The issue is that references can be higher up than the point at which we're making the replacement. So to do the wrapping correctly, the user has to be quite careful. They'll have to use `node.walk()` from the top of their model to do it reliably. So maybe we should just do this for them. This would also match the behaviour in `remove_node` better.

So we'd have something like this:

```python
class Model:
    def replace_node(self, old: Model, new: Model) -> bool:
        """Replace a node anywhere it occurs within the model. Returns a boolean indicating
        whether the replacement was made."""
        seen = False
        for node in list(self.walk()):
            if node is old:
                seen = True
            else:
                node._layers = [new if child is old else child for child in node._layers]
                for name in node.ref_names:
                    if node.get_ref(name) is old:
                        node.set_ref(name, new)
        return seen
```

danieldk:
> So maybe we should just do this for them. This would also match the behaviour in `remove_node` better.

Yeah, I agree, this would be much nicer and more consistent with the behavior of `remove_node`.

danieldk:
Updated the PR with the updated `replace_node`. This also simplified `wrap_model_recursive` a lot, which is a good sign.

honnibal:
Sorry I didn't see this sooner. Looks good!

honnibal (approved):

PR #536, opened 2021-09-16
CategoricalCrossEntropy: support missing values for int arrays

danieldk:
`CategoricalCrossEntropy` provides a `missing_value` option that can be used
to mask instances that have a special label. However, this functionality
only works with non-integer categorical labels.

This change adds support for masking missing values for int lists and xp
arrays.

### DIFF
--- a/thinc/loss.py
+++ b/thinc/loss.py
@@ -97,7 +97,7 @@ class CategoricalCrossentropy(Loss):
             truths = xp.asarray(truths, dtype="i")
             mask = _make_mask(guesses, missing)
         else:
-            mask = _make_xp_mask(truths, guesses, missing_value)
+            mask = _make_mask_by_value(truths, guesses, missing_value)
         if truths.ndim != guesses.ndim:
             # transform categorical values to one-hot encoding
             truths = to_categorical(cast(Ints1d, truths), n_classes=guesses.shape[-1])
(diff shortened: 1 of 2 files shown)


thinc/loss.py
```diff
@@ -342,6 +351,22 @@ def _make_mask(missing, guesses) -> Floats2d:
     return mask
 
 
+def _make_xp_mask(guesses, truths, missing_value) -> Floats2d:
```
svlandeg, line 354: This looks good to me in general, but it annoys me slightly that the signature of `_make_xp_mask` is so different to `_make_mask` in the order of the arguments. Maybe `_make_mask` should have had the order of arguments reversed, and it is a private function so maybe it wouldn't even hurt to change this on the fly, just to have some kind of consistency between the function declarations?

svlandeg: We should probably also reconsider the naming of these two functions...

danieldk: I changed the order of `_make_xp_mask` to correspond to `convert_truths`.

Do you have suggestions for the name? Maybe `_missing_value_mask`?

svlandeg: `_make_mask_by_value`?

danieldk: 👍🏻 Done!

svlandeg: Oh hm wait, I'm confused, I thought we'd have `_make_mask` as before and `_make_mask_by_value` as the latter checks on the missing value?

danieldk: Makes sense! I'll update the names.

danieldk: I did do a force push this time, to avoid a commit which fixes a commit ;).

PR #537, opened 2021-09-21
Fix replace_node on nodes with indirect node refs

danieldk:
`Model.replace_node` failed if a node *n* contained a reference to another node *m*, when *m* only occurs in one of the (grand)children of *n*.  Since `replace_node` used the breadth-first `Model.walk` method, the reference to *m* would be replaced before *m* is replaced in the (grand)children of *n*. However, this fails because `set_ref` verifies that the a new node reference occurs in the graph.

This change resolves this issue by replacing nodes using a depth-first search.

### DIFF
--- a/thinc/model.py
+++ b/thinc/model.py
@@ -352,7 +352,22 @@ class Model(Generic[InT, OutT]):
             for name, param in backup.items():
                 self.set_param(name, param)
 
-    def walk(self) -> Iterable["Model"]:
+    def walk(self, *, order: str = "bfs") -> Iterable["Model"]:
+        """Iterate out layers of the model.
+
+        Nodes are returned in breadth-first order by default. Other possible
+        orders are "dfs_pre" (depth-first search in preorder) and "dfs_post"
+        (depth-first search in postorder)."""
+        if order == "bfs":
+            return self._walk_bfs()
+        elif order == "dfs_pre":
+            return self._walk_dfs()
+        elif order == "dfs_post":
+            return self._walk_dfs(post_order=(True))
+        else:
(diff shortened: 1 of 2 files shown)


thinc/model.py
```diff
@@ -352,7 +352,22 @@ def use_params(self, params: Dict[Tuple[int, str], FloatsXd]):
...
+
+        Nodes are returned in breadth-first order by default. Other possible
+        orders are "dfs_pre" (depth-first search in preorder) and "dfs_post"
+        (depth-first search in postorder)."""
+        if order == "bfs":
+            return self._walk_bfs()
+        elif order == "dfs_pre":
+            return self._walk_dfs()
+        elif order == "dfs_post":
+            return self._walk_dfs(post_order=(True))
```
svlandeg, line 366:
```suggestion
            return self._walk_dfs(post_order=True)
```

thinc/model.py
```diff
@@ -352,7 +352,22 @@ def use_params(self, params: Dict[Tuple[int, str], FloatsXd]):
...
+    def walk(self, *, order: str = "bfs") -> Iterable["Model"]:
+        """Iterate out layers of the model.
+
+        Nodes are returned in breadth-first order by default. Other possible
+        orders are "dfs_pre" (depth-first search in preorder) and "dfs_post"
+        (depth-first search in postorder)."""
+        if order == "bfs":
+            return self._walk_bfs()
+        elif order == "dfs_pre":
+            return self._walk_dfs()
```
svlandeg, line 364:
```suggestion
            return self._walk_dfs(post_order=False)
```

svlandeg (commented):
I think it doesn't hurt to be specific

svlandeg (approved):

PR #538, opened 2021-09-28
Fix name of the layer created with with_nvtx_range

danieldk:
The layer was named `debug(...)`, should be `nvtx_range(...)`. This is
only a cosmetic change, but looks nicer when debugging models.


danieldk:
I will close this and tack it on the larger NVTX PR that I'll make soon.

PR #539, opened 2021-09-30
Wrap layer callbacks in with_{debug,nvtx_range}

danieldk:
Before this change, with_{debug,nvtx_range} instrumented a `Model`'s init/forward/backprop by wrapping it in another `Model`. However, this turns out to be tricky, since in some cases we have to propagate information from the wrapped nodes (e.g. references) while in other cases we shouldn't.
    
With this change, we instead add instumentation by wrapping a `Model`'s init and forward callbacks. This makes the nodes in the graph appear as they were before wrapping, but adds behavior in the wrapped callbacks.
    
The with_{debug,nvtx_range} still returns `Model` to maintain compatibility with the existing API. Though the returned `Model` is the `Model` that is wrapped.

### DIFF
--- a/thinc/layers/with_debug.py
+++ b/thinc/layers/with_debug.py
@@ -1,28 +1,32 @@
-from typing import Optional, Callable, Any, Tuple
+from typing import Optional, Callable, Any, Tuple, TypeVar
 
 from ..model import Model, wrap_with_callbacks
+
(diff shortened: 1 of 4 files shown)


honnibal:
We tried to avoid modifying the internal `_func` here, but this way really is better. I'll consider adding a `replace_function` method later -- but for now, let's get this merged :)

PR #540, opened 2021-10-05
get_ops: support returning the best avilable CPU backend

danieldk:
When using `{get,use}_ops`, we often do no want a particular CPU backend, but the best CPU backend that is available (e.g. Accelerate-based thinc ops on a Mac).

This change adds a special op value `cpu` that returns the best available CPU backend. E.g.:

```python
with use_ops("cpu"):
   # ...
```


honnibal:
LGTM!

PR #552, opened 2021-10-25
mixed-precision: do not backprop Inf/NaN out of PyTorch layers

danieldk:
With mixed-precision enabled in a PyTorchWrapper, an update is skipped and the gradient scale is adjusted when one or more of the gradients is Inf/NaN. However, Inf/NaN gradients were still backpropagated into preceding Thinc layers.

This change fixes this issue by clearing gradients of inputs to PyTorchWrapper if Inf/NaN occurs in any of the gradients.

Also:

* Change `PyTorchGradScaler.scale` to accept `Tensor` besides `Sequence[Tensor]`.
* Expose `PyTorchWrapper_v2` in the API
* Add a unit test for mixed-precision training.

---

~Set to draft because:~

- [x] ~Still needs testing on a transformer model.~ Similar accuracies in a tagger -> morphologizer pipeline ✅ 
- [x] ~Needs to be rebased after #553  and #554.~ ✅

### DIFF
--- a/thinc/api.py
+++ b/thinc/api.py
@@ -25,6 +25,7 @@ from .layers import CauchySimilarity, ParametricAttention, Logistic
 from .layers import resizable, sigmoid_activation, Sigmoid, SparseLinear
 from .layers import PyTorchWrapper, PyTorchRNNWrapper, PyTorchLSTM
 from .layers import TensorFlowWrapper, keras_subclass, MXNetWrapper
+from .layers import PyTorchWrapper_v2
 
 from .layers import add, bidirectional, chain, clone, concatenate, noop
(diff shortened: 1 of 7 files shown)


honnibal:
Nice! Looks good.

Btw you could use Hypothesis to test a range of input tensors instead of one arbitrary one if you wanted. I think the test looks fine as it is though.

danieldk:
> Btw you could use Hypothesis to test a range of input tensors instead of one arbitrary one if you wanted. I think the test looks fine as it is though.

That's a good idea, I'll adjust the test to do that.

danieldk:

PR #559, opened 2021-11-03
Support big endian platform by providing new ops implementation

andrewsi-z:
Based on discussion in spaCy [issue 9428](https://github.com/explosion/spaCy/issues/9428), there are a set of modules related to the load and use of pre-trained language pipelines and/or models on a platform with a different byteorder.   This leads to incorrect results produced by the use of these pretrained models on platforms such as s390x.

As discussed in the spaCy issue referenced above, this PR adds support for thinc-bigendian-ops , which is created and located here: https://github.com/andrewsi-z/thinc-bigendian-ops .  

thinc-bigendian-ops follows the precedent and use of thinc-apple-ops. When imported, it provides a custom ops class that implements the following:
- numpy_ops.pyx hash has logic that maps a unsigned char array to the input ids array (uint64). The approach used maps the underlying storage and "views" as char array, and this implementation will result in different values depending on system byteorder. [BigEndianOps modifies this algorithm](https://github.com/andrewsi-z/thinc-bigendian-ops/blob/951081541467fdc7759fadee9b3a96e4166e3385/thinc_bigendian_ops/bigendian_ops.pyx#L43-L66) and provides an alternative based on bit-shift operations, which will return the same value regardless of system byte order.
- A new [asarray](https://github.com/andrewsi-z/thinc-bigendian-ops/blob/951081541467fdc7759fadee9b3a96e4166e3385/thinc_bigendian_ops/bigendian_ops.pyx#L18-L38) method is implemented, which checks the byteorder of the output before returning it to caller. In cases where the byteorder is littleendian, it is swapped. 

The logic in thinc proper is meant to mirror the logic added for thinc_apple_ops in a minimally intrusive way. 

Notes:

* ran test suite with thinc-bigendian-ops on s390x (big endian)
* ran spaCy _sm model pipelines to validate original use cae.


honnibal:
I think this looks like a good match for how we've been doing things. @adrianeboyd do you have any comments? It looks okay to me. If it all works, I'm pleased to resolve this long-standing platform support issue!

I do think the mechanism we've designed for the ops selection could be better though. It's quite unsightly the way we have to reference the subclasses in Thinc, and it leads to the circular import problems.

Here's a suggestion we could try to improve this:

* Ops classes implement a classmethod `get_hardware_suitability(name, **kwargs)` that returns an integer. A negative integer indicates incompatibility. A higher integer indicates that the ops class claims it should be used. `kwargs` can be passed in from above to influence the match.
* In `get_ops`, we sort the registered ops by their hardware suitability score, and select the highest available. If no compatible ops are found, we raise an error.

This will prevent Thinc from having to know all of the ops classes in order to resolve which one should be used. Generally ops classes should return `-1` for incompatible, `0` for compatible but not optimised, and `1` for optimised. If another plugin wants to overrule existing ones in specific circumstances it can return a higher match score under those conditions (including by supporting arbitrary keyword arguments).

I think we could merge the current PR as-is, as it doesn't seem to make things worse (that I can see). But I think the patch does indicate we could improve our system here.

andrewsi-z:
For reference on associated issues, I should mention I have the needed murmurhash changes here:  https://github.com/andrewsi-z/murmurhash
This uses the murmurhash3 endian agnostic design introduced for (I believe) node use of it. 
I followed this same design and applied it to murmurhash2.cpp hash64a, which caused errors. I will open up an issue against explosion/murmurhash to discuss whether you'd like these fixes or prefer to keep a separate murmurhash library for big endian (which also functionally works fine).

adrianeboyd:
I think I'd reverse/massage the logic in `get_ops` in bit, but given the current state of `thinc`, the basics for this PR seem fine.

Adding `AppleOps` is simpler in a number of ways than `BigEndianOps`:
* you can't install it on systems where it wouldn't run correctly
* backing off to `numpy` accidentally is slower, not broken

So we need to be more careful around `BigEndianOps`. I also hesitate a bit to add this without any CI testing on our end, but I'm not sure that we have any feasible options.

thinc/backends/__init__.py
```diff
@@ -91,10 +95,15 @@ def get_ops(name: str, **kwargs) -> Ops:
     cls: Optional[Callable[..., Ops]] = None
     if name == "cpu":
         _import_extra_cpu_backends()
+        
         cls = ops_by_name.get("apple", ops_by_name.get("numpy"))
+        
+        if "bigendian" in ops_by_name:
+            cls = ops_by_name.get("bigendian", ops_by_name.get("numpy"))
+
```
adrianeboyd, line 103:
```suggestion
        cls = ops_by_name.get("numpy")
        cls = ops_by_name.get("apple", cls)
        cls = ops_by_name.get("bigendian", cls)
```

adrianeboyd: This is where a ranking from the ops classes themselves would be better.

thinc/backends/__init__.py
```diff
@@ -113,6 +122,9 @@ def get_array_ops(arr):
 def use_ops(name: str, **kwargs):
     """Change the backend to execute on for the scope of the block."""
     current_ops = get_current_ops()
+    # avoid fallback to base NumpyOps if on big endian platform
+    if current_ops.name == "bigendian" and name == "numpy":
+        name = current_ops.name
```
adrianeboyd, line 127: I think we've removed all the `use_ops("numpy")` from our code, so I think I'd rather have at most a warning/error rather than having `use_ops` silently not do what you just told it to do.

andrewsi-z:
Hi @adrianeboyd, thank you for your comments and I'll start working on the requested changes.

You brought up CI... I know github actions doesn't support s390x yet, but there are a number of supporting options that can potentially (?) be triggered by a github action. These would be community available no cost resources .

I can explore this as a follow-on action and present some options for discussion (and help get it going if you and the explosion team believe it is worthwhile to purse).

thinc/backends/__init__.py
andrewsi-z: Updated with suggested change and retested.

thinc/backends/__init__.py
andrewsi-z: I removed the highlighted code... At best, this was my attempt to keep an explicit use_ops("numpy") specification from reverting back to numpy ops when explicitly requested from outside of thinc. It is best not to ignore an explicit request.

svlandeg:
While we probably want to think about more generic solutions in the future, this should be fine to merge as is for now, and will become available in Thinc's next release 8.0.114.

PR #566, opened 2021-12-12
Try out making Ragged generic

honnibal:
The `Ragged` type is troublingly vague at the moment, because it doesn't tell you what type it holds. Try making it generic. Could require lots of adjustments, I'm not sure?


richardpaulhudson:
I hadn't seen this when I started on https://github.com/explosion/thinc/pull/599, but seem to have implemented everything that is here in the course of implementing https://github.com/explosion/thinc/pull/599.

PR #568, opened 2021-12-14
New hard activations

kadarakos:
This is the first PR in a series of PRs adding new activation functions to thinc. 

The general goal is to add activations to be able to reproduce different models and have more options. Specifically in this PR other than Swish the remaining activation functions are designed to be fast and to work well in low precision.  

The new function in this PR are:

1. HardSigmoid https://paperswithcode.com/method/hard-sigmoid
2. HardTanh
4. ClippedLinear: generic linear nonlinearity with parametrizeable `slope`, `offset`, `min` and `max`. Its used to implement the previous three.
5. Swish https://blog.paperspace.com/swish-activation-function/
6. HardSwish  https://www.scitepress.org/Papers/2019/74696/74696.pdf 
7. HardSwishMobilenet https://arxiv.org/abs/1905.02244v5
8. GELU: "almost exact" GELU with approximate `erf`
9. Approximate GELU: the approximation from the original paper https://paperswithcode.com/method/gelu

There is also a `relu_n` function in `thinc/backends/ops.py`, because I wasn't clear yet how to implement the `ReLU` with a threshold above (useful for low precision). At the moment its internally used by HardSwishMobilenet, but in the future it will be somehow merged with `relu`. 

The forward and backward ops are implemented in `thinc/backends/ops.py`, but the optimized `Cython` and `CUDA` versions are deferred to a next PR. 

For each op there is a separate dense layer implemented in `thinc/layers/`, however, the next PR will attempt to refactor this into a single Dense() API that takes the activation function as an argument.


thinc/backends/ops.py
```diff
@@ -364,7 +364,7 @@ def get_dropout_mask(self, shape: Shape, drop: Optional[float]) -> FloatsXd:
         if drop is None or drop <= 0:
             return self.xp.ones(shape, dtype="f")
         elif drop >= 1.0:
-            return self.alloc(shape)
+           return self.alloc(shape)
```
svlandeg, line 367: Nitpick: Have you run `black` on this file? There should be 4 spaces here instead of 3 :-)

compare_to_torch.py
```diff
@@ -0,0 +1,65 @@
+import torch
```
svlandeg, line 1: I'm not sure we want to keep this script in the main repo, but if we do, perhaps we can move it in the `bin` folder?

thinc/backends/ops.py
```diff
@@ -569,13 +569,13 @@ def as_contig(self, data: ArrayT, dtype: Optional[DTypes] = None) -> ArrayT:
...
             self.xp.exp(-X, out=X)
-            X += 1.0 # type: ignore
+            X += 1.0 # type: ignore 
             X **= -1.0 # type: ignore
             return cast(FloatsType, X)
         else:
             return cast(FloatsType, 1.0 / (1.0 + self.xp.exp(-X)))
 
-    def dsigmoid(self, Y: FloatsType, *, inplace: bool = False) -> FloatsType:
+    def dsigmoid(self, Y: FloatsT, *, inplace: bool = False) -> FloatsT:
```
svlandeg, line 578:
```suggestion
    def dsigmoid(self, Y: FloatsType, *, inplace: bool = False) -> FloatsType:
```
Let's revert this back so this PR focuses on the new functions. 

@honnibal for some context: we've been discussing some of the type annotations throughout the `ops` module here, and the best way forward to deal with some of the remaining `mypy` issues. I suggested to discuss this on a separate PR though, and focus on the actual implementations here.

svlandeg (commented):
Nice first PR, Ákos! I'm sure Matt will want to review as well, but I already had a few nitpicking comments ;-)

compare_to_torch.py
kadarakos: I really wasn't sure what to do with this. Its really helpful to me when developing, but I also don't think the main repo is the right place. `bin` also sounds good to me, or I can also remove it from the repo.

kadarakos: I use this for development and I kinda like it that I can change the range for example of the values.

danieldk: I think code like that tends to bit rot. Maybe it's better to turn this into actual tests that are guarded with `has_torch`? We already have a bunch of tests that use PyTorch and install PyTorch in CI. That would have the benefit that these tests are always run in CI as well.

svlandeg: Great idea, let's do that :-)

kadarakos: Thanks for the suggest :-). I've pushed the test in `thinc/tests/backends/test_ops.py` with `has_torch`

svlandeg:
FYI: the new `mypy` failure should hopefully be fixed by https://github.com/explosion/thinc/pull/569

kadarakos:
> FYI: the new `mypy` failure should hopefully be fixed by #569

phew that's good to know!

svlandeg:
@kadarakos : you can try pulling and merging `upstream/master` and push to this PR again, and the CI will hopefully behave :-)

kadarakos:
> @kadarakos : you can try pulling and merging `upstream/master` and push to this PR again, and the CI will hopefully behave :-)

Thanks! Enjoying watching the green check marks again.

thinc/backends/ops.py
```diff
@@ -680,6 +682,167 @@ def backprop_relu(
...
+        if inplace:
+            self.xp.clip(X, 0, n, out=X)
+            return X
+        return self.xp.clip(X, 0, n)
+
+    def backprop_relu_n(
+        self, dY: FloatsType, X: FloatsType, n: float = 6.0, inplace: bool = False
+    ) -> FloatsType:
+        if inplace:
+            dY *= (0 < X) & (X < n)
```
danieldk, line 697: One operation less, though I don't know if it is faster:

```
dY *= self.xp.clip(X, 0, n) != 0.
```

Which is of course:

```
dY *= relu_n(X) != 0.
```

thinc/backends/ops.py
```diff
@@ -680,6 +682,167 @@ def backprop_relu(
...
+            X *= 0.2  # type: ignore
+            X += 0.5  # type: ignore
+            return cast(FloatsType, self.xp.clip(X, 0, 1, out=X))
+        out = X * 0.2 + 0.5
+        return cast(FloatsType, self.xp.clip(out, 0, 1))
+
+    def backprop_hard_sigmoid(
+        self, dY: FloatsType, X: FloatsType, inplace: bool = False
+    ) -> FloatsType:
+        slope = self.alloc1f(1) + 0.2
```
danieldk, line 713: I am not sure if I understand why this needs an array? Also, since `alloc1f` creates a `float32` array by default, this would reduce precision if `X` is a `float64` array.

thinc/backends/ops.py
```diff
@@ -680,6 +682,167 @@ def backprop_relu(
...
+        a3 = 1.421413741
+        a4 = -1.453152027
+        a5 = 1.061405429
+        p = 0.3275911
+
+        t = 1.0 / (1.0 + p * X)
+        y = 1.0 - (((((a5 * t + a4) * t) + a3) * t + a2) * t + a1) * t * self.xp.exp(
+            -X * X
+        )
+        return sign * y
```
danieldk, line 791: I think we'd need copyright attribution for such a large fragment?

If we want a naive implementation that we override with `NumpyOps`, I would use something like (untested):

```
erf = np.vectorize(lambda x: math.erf(x))
return erf(X)
```

thinc/backends/ops.py
```diff
@@ -680,6 +682,167 @@ def backprop_relu(
...
+
+    def sechsq(self, X: FloatsType) -> FloatsType:
+        return (1 / self.xp.cosh(X)) ** 2
+
+    # https://github.com/huggingface/transformers/blob/master/src/transformers/activations.py#L37
+    def gelu_approx(self, X: FloatsType, inplace: bool = False) -> FloatsType:
+        tmp = 1.0 + self.xp.tanh(
+            self.xp.sqrt(2 / math.pi) * (X + 0.044715 * self.xp.power(X, 3))
```
danieldk, line 799: Better to make this a constant:

`self.xp.sqrt(2 / math.pi)`

To avoid recomputing it over and over again.

thinc/backends/ops.py
```diff
@@ -680,6 +682,167 @@ def backprop_relu(
...
+        )
+        tmp += 0.5
+        dX += tmp
+        if inplace:
+            dY *= dX
+            return dY
+        return dY * dX
+
+    def gelu(self, X: FloatsType, inplace: bool = False) -> FloatsType:
+        tmp = 1.0 + self.erf(X / self.xp.sqrt(2.0))
```
danieldk, line 828: `self.xp.sqrt(2.0)`

This could be a constant as well.

danieldk (commented):
Really nice to have these functions 🎉. I only had time so far to look at the `Ops` part and added some comments.

thinc/backends/ops.py
danieldk:
```
erf = np.vectorize(math.erf)
```

might also work

thinc/backends/ops.py
```diff
@@ -680,6 +682,167 @@ def backprop_relu(
         dY *= Y > 0
         return dY
 
+    def relu_n(
```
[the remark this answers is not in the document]
kadarakos: I was thinking about this, but I was wondering whether its better to do it in the next PR where I'll try to improve the consistency of the activation function `ops` + add a generic `Dense(dim=256, activation="gelu")` wrapper layer.

thinc/backends/ops.py
kadarakos: I wasn't sure where is the best place to store this constant. Do you have a suggestion?

thinc/backends/ops.py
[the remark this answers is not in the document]
danieldk: Since this is a new function, I think it makes sense to do it right the first time.

thinc/backends/ops.py
danieldk: I think it's good custom to place constants somewhere near the top of the file. (Probably between the typevars and the `Ops` class?)

thinc/backends/ops.py
kadarakos: When the result of the activation goes propagates to `thinc.backends.numpy_ops.NumpyOps.gemm` on the backward pass I've gotten this error, when returning `float64`: 

```
File "blis/py.pyx", line 64, in blis.py.gemm
ValueError: Buffer dtype mismatch, expected 'const double' but got 'float'
```
So my understanding was that `thinc` ops should live in `float32`

thinc/backends/ops.py
kadarakos: okok, I was debating whether i should make it an attribute of something or is it just better to put it as an actual constant. Thanks!

thinc/backends/ops.py
danieldk: Do you have a reproducer for this error. I think that this might happen when the input is float32, but when you outputting float64, because then `gemm` will be called with matrices of different types. I think it should work correctly if when the inputs are float32, you return float32 and when the inputs are float64, you return float64.

The issue is that you loose the type with the conversion to bools and then the result gets whatever the result of `slope` is. If you do something like

```
dX = self.xp.where(self.xp.abs(X) < 2.5, 0.2 * X, 0.)
```

`dX` will always have the same dtype as `X`.

thinc/backends/ops.py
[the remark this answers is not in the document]
kadarakos: Then I guess `relu_n` can just become a `ops.clip(min_val, max_val)` with an adjoint `dY * self.clip(X, min_val, max_val) != 0`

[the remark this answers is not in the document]
kadarakos: I was just also not sure where should this stop because it can be even more general with a slope and an offset.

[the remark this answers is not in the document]
kadarakos: Or like if you look at the `hard_swish` functions they are the same but then times x. 

like the very general version I guess would be like `(x or 1) *  self.xp.clip(x * slope + offset, min_val, max_val)`.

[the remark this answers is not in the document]
kadarakos: For hard sigmoids these are the ones ive came across

Hard-sigmoid:

https://arxiv.org/pdf/1603.00391.pdf

`clip(x * 0.25 + 0.5, 0, 1)`

https://arxiv.org/pdf/1511.00363v3.pdf

`clip(x * 0.5 + 0.5, 0, 1)`


https://pytorch.org/docs/stable/generated/torch.nn.Hardsigmoid.html

`x * 0.166667 + 0.5, x[x <= -3] = 0, x[x >=3] = 1`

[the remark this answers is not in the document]
danieldk: There are many different abstraction levels. My main motivation was to have a function that is similar to what is offered in other libraries.

But it was only a minor suggestion anyway, feel free to ignore.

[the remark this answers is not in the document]
kadarakos: I agree that it would be nice, but I was just not sure what is that other libraries offer for these hard functions. I looked at the hard sigmoid in Keras:

https://github.com/keras-team/keras/blob/2c48a3b38b6b6139be2da501982fd2f61d7d48fe/keras/backend.py#L5196

and Pytorch

https://pytorch.org/docs/stable/generated/torch.nn.Hardsigmoid.html

but they didn't seem to be the same.  For the MobileNetV3 application its like the Pytorch version then: 

https://github.com/keras-team/keras/blob/v2.7.0/keras/applications/mobilenet_v3.py#L455

So would you think it makes sense to implement a generic `ops.clipped_linear` with

`clip(x * slope + 0.5, min_val, max_val)`. This can do `relu`, `relu_n`, `hard_tanh` and `hard_sigmoid`. And this then comes with a generic `ops.backprop_clipped_linear`?

thinc/backends/ops.py
kadarakos: Oh I think the `clip` in the backprop is not correct, because it should be strictly `<` and `>` and not `<=` and `=<` like `clip` does.

thinc/backends/ops.py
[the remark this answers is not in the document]
kadarakos: I've pushed the generic `clipped_linear`, let me know what do you think! `test_clipped_linear` shows how it can implement `relu6` and `hard_sigmoid`. `hard_tanh` would be `ops.clipped_linear(min_val=-1, max_val=1)`, but it could do other stuff too.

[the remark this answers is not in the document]
danieldk: Nice 👍🏻🎉!

thinc/backends/ops.py
danieldk: I would be surprised if it matters in practice. Weights are rarely integers.

kadarakos: I doesn't but `hypothesize` fails the test with the evil value of `6.0`

danieldk: Ah, I see! 😀

thinc/backends/ops.py
honnibal: Does cupy support `vectorize`? If not we can't do it that way in the `ops` function so easily.

danieldk: cupy has `vectorize`, but it doesn't know how to deal with `math` functions.

But does every `Ops` method have to run with CuPy arrays? Overriding this in `CupyOps` should be straightforward, since CuPy does have an `erf` function.

svlandeg:
Maintenance note: there's a few open questions/comments here that might not be addressed until after the holidays, so I'll put this "in draft" for the time being :-)

thinc/backends/ops.py
kadarakos: I thought that the general design is that `ops` has the basic implementations that run both with `cupy` and `numpy` and then the `NumpyOps` and `CupyOps` can optimize it further. Is it hard to credit the website for the`erf` implementation somehow?

thinc/backends/ops.py
kadarakos: With the new `clipped_linear` do you think it would make sense to replace the `relu`, `relu_n` and `hard_sigmoid` with just a `ClippedLinear` layer that calls `clipped_linear` in the `ops`?

kadarakos: I pushed a `ClippedLinear` layer, if you change the `Relu` in `examples/mnist.py` to `ClippedLinear` it will throw the 

```
  File "thinc/backends/numpy_ops.pyx", line 93, in thinc.backends.numpy_ops.NumpyOps.gemm
    return blis.py.gemm(x, y, out=out, trans1=trans1, trans2=trans2, beta=0.)
  File "blis/py.pyx", line 64, in blis.py.gemm
ValueError: Buffer dtype mismatch, expected 'const double' but got 'float'
```

danieldk: See my comment above, I think this happens if `gemm` is passed arrays of different types. Things like

```
dX = slope * ((low < X) & (X < high))
```

Are not safe, since they discard the data type of X and result in an array that has the same data type as `slope`. E.g., if `slope` is a Python `float`, this fill create a `np.float64` array. However, this is incorrect when the input as a `np.float32` array. Then in the caller of the backward function, a `gemm` is attempted of an `np.float32` and an `np.float64` array.

See my earlier comment for an approach (with `np.where`) that should not have this issue.

danieldk: By the way, the current implementation should also not typecheck, since `dY`, `X` and the return type must be the same type (`FloatsType` is a generic type).

honnibal:
Really nice work!

I don't really have any corrections to add, except to note that we may want activation-only layers for these functions as well possibly?

Regarding the tests, the PyTorch comparison is a nice utility. Are the tests that use it marked as needing pytorch? I'm pretty sure we have a marker for that.

thinc/backends/ops.py
kadarakos: Oh okay sorry! Because of you typecheck comment @danieldk I've remembered that both `numpy` and `cupy`has `zeros_like` so I've implemented `clipped_linear` with that, because I kinda like that pattern. The `hard_sigmoid` and `relu_n` are now calling `clipped_linear` and its `backprop` and we have `hard_tanh` now too.

thinc/api.py
```diff
@@ -12,7 +12,13 @@
...
 from .types import Ragged, Padded, ArgsKwargs, Unserializable
 from .util import fix_random_seed, is_cupy_array, set_active_gpu
-from .util import prefer_gpu, require_gpu, require_cpu, DataValidationError, data_validation
+from .util import (
+    prefer_gpu,
+    require_gpu,
+    require_cpu,
+    DataValidationError,
+    data_validation,
+)
```
svlandeg, line 21:
```suggestion
from .util import prefer_gpu, require_gpu, require_cpu,
from .util import DataValidationError, data_validation
```
Battling `black` is like a full-time job ;-)

thinc/backends/ops.py
danieldk: Resolved by giving credit to John D. Cook, who put this code in the public domain (see comments attached to the article).

thinc/layers/clipped_linear.py
```diff
@@ -0,0 +1,150 @@
...
+        nI=nI,
+        init_W=init_W,
+        dropout=dropout,
+        normalize=normalize,
+        min_val=-1.0,
+        max_val=1.0,
+    )
+
+
+@registry.layers("ReluK.v1")
```
danieldk, line 131: `ReluK` or `ReluN`?

kadarakos: I've renamed the op to `relu_k` I quite prefer that `k` for some reason :)

kadarakos:
> Really nice work!
> 
> I don't really have any corrections to add, except to note that we may want activation-only layers for these functions as well possibly?
> 
> Regarding the tests, the PyTorch comparison is a nice utility. Are the tests that use it marked as needing pytorch? I'm pretty sure we have a marker for that.

Sorry for replying late, yes the test has the `has_torch`

danieldk (approved):
Looks good to me! I did a detailed review of the `Ops` part. The layers all follow the same pattern, so I mostly reviewed `ClippedLinear`.

thinc/backends/ops.py
```diff
@@ -680,6 +685,191 @@ def backprop_relu(
...
+            X *= self.relu_k(X + 3) / 6
+            return X
+        return X * (self.relu_k(X + 3) / 6)
+
+    def backprop_hard_swish_mobilenet(
+        self, dY: FloatsType, X: FloatsType, inplace: bool = False
+    ) -> FloatsType:
+        dX = (1 / 6) * (X * 2.0 + 3.0)
+        dX[X > 3.0] = 1.0
+        dX[X < -3.0] = 0
```
honnibal, line 798: This looks like it'll return `float64` I think

thinc/backends/ops.py
```diff
@@ -680,6 +685,191 @@ def backprop_relu(
...
+        a3 = 1.421413741
+        a4 = -1.453152027
+        a5 = 1.061405429
+        p = 0.3275911
+
+        t = 1.0 / (1.0 + p * X)
+        y = 1.0 - (((((a5 * t + a4) * t) + a3) * t + a2) * t + a1) * t * self.xp.exp(
+            -X * X
+        )
+        return sign * y
```
honnibal, line 822: Float type?

thinc/backends/ops.py
kadarakos: Seems like it, but it passes the test for some reason.

PR #572, opened 2022-01-16
Add assert that size hasn't changed for reduce mean backprop

polm:
Behavior of this with numpyops seems wrong, as instead of giving an
error it produces nans, as though it's accessing out of bounds memory or
something. See https://github.com/explosion/spaCy/pull/9669.

Some more general issues / questions around this:

1. Do we want to add this kind of check for all cases where appropriate?
2. Is there a better place (like in numpyops) for this check or an equivalent check?
3. Other asserts in Thinc seem to not use messages, should the message here be removed?


svlandeg:
Ah, good catch on the spaCy PR!

I would personally only use `assert` statements to check for logic that is entirely governed by the internal code, i.e. not checking for user input. In the context of this function being defined in Thinc, I'm not sure that holds up? I think this becomes apparent from the fact that you want to throw a user-facing error message - i.e. maybe this should be an actual check + raising an error.

I definitely think it would be good to check for other places where such checks would be appropriate, as it'll definitely help with user experience to catch these kind of mismatches earlier.

polm:
OK, I have moved the logic into the actual layer implementation. Currently this covers all the `reduce_*` layers and, somewhat arbitrarily, softmax. 

I think there are more things that could be covered by this, but it might make sense to use a decorator for the backprop function. Something like this:

```
def forward(...):
    Y = ...

    @consistent_backprop(Y.size)
    def backprop(dY: ..., ...)
        ....
```

Inspecting arguments in a decorator can get complicated, but it looks like all our backprop functions are unary, so that should actually be pretty simple to do.

svlandeg (commented):
I like this approach, it definitely feels more elegant to add this type of decorator instead of having to add this kind of functionality to each of the relevant layers.

I noticed it's still in draft - are you still working on this?

polm:
This is OK as-is, but I left it in draft to think about whether there was a more general way to apply it, and whether it should be applied to other functions.

It would be helpful to check this for basically any layer that has an array as input and output; this mainly covers the `reduce` family because that's where I noticed the error. So there's also a question of which other layers exactly it should be applied to.

thinc/util.py
```diff
@@ -510,6 +510,30 @@ def set_torch_tensor_type_for_ops(ops):
...
 
+def consistent_backprop(forward):
+    """Decorator to check that backprop input is consistent.
+
+    This should be placed on the forward function, and only works with backprop
+    that takes a single Array2D input.
+    """
+
+    print("HERE", forward)
+
```
adrianeboyd, line 521:
```suggestion
```

polm:
For now I think we don't want to expand this to other functions, so I'd just consider it ready as-is. I would like to consider adding more checks like this later, since I think when debugging it can help avoid cryptic errors. On the other hand it really is more important for some of these reduce functions which will silently and unpredictably produce NaNs with bad input.

svlandeg (approved):
I think this approach is rather elegant :-)

polm:
After thinking about this a bit, this includes a non-decorator approach to this. The basic idea is to have a dataclass that contains information about the array necessary for consistency checking, and to call a method on the dataclass on the gradient that checks consistency. That looks a little like this:

```
def forward(...):
    ...
    ainfo = create_arrayinfo(Y)
    def backward(dY: OutT):
        ainfo.check_consistency(dY)
```

The consistency check throws an exception if something is wrong. 

This is slightly more complicated than the decorator but not much. Error handling code is still in one place, and the factory function takes care of getting rid of any references that would hold up GC. You have to add two lines (besides an import) instead of the one decorator, but that's not bad. You do need to provide arguments, but again, no big deal.

The `create_arrayinfo` call is a little weird and there are several alternatives to consider

1. `ArrayInfo.create(...)`: maybe more awkward?
2. override constructor `ArrayInfo(...)`: Not sure if there are any downsides to doing that with a dataclass, probably not here?

If this looks OK I can replace the decorator with it wherever necessary.

svlandeg:
> The `create_arrayinfo` call is a little weird and there are several alternatives to consider
> 
>     1. `ArrayInfo.create(...)`: maybe more awkward?
> 
>     2. override constructor `ArrayInfo(...)`: Not sure if there are any downsides to doing that with a dataclass, probably not here?

Maybe `ArrayInfo.from_array()`?

honnibal:
> Maybe `ArrayInfo.from_array()`?

Yeah that's what I would call it, and we definitely prefer not to override the dataclass constructors, the classmethod is the way to go 👍

honnibal:
There's another question around whether to make the check a method or a function. I probably would've chosen a function but I don't really have a strong motive either way.

polm:
OK, switched to create with `ArrayInfo.from_array()` everywhere, so I think this is good now. 

> There's another question around whether to make the check a method or a function. I probably would've chosen a function but I don't really have a strong motive either way.

It's not a big deal but I think one advantage of using the method is reducing the number of names you have to import.

polm:
Note: this needed a merge because the recent softmax normalization changes caused a minor conflict.

thinc/layers/reduce_first.py
```diff
@@ -20,7 +21,10 @@ def forward(model: Model[Ragged, OutT], Xr: Ragged, is_train: bool) -> Tuple[Out
     x_shape = Xr.dataXd.shape
     lengths = Xr.lengths
 
+    ainfo = ArrayInfo.from_array(Y)
```
svlandeg, line 24: Should we just call this variable `array_info` for the sake of clarity? It's not that much longer...

svlandeg:
Looks good to me! There just seem to be a few issues with `mypy` that still need to get fixed.

thinc/layers/reduce_first.py
polm: Sure, I'll change it.

polm:
OK, partly resolved the typing issues.

The remaining issue is that for ArrayXd types, the `dtype` is supposedly a string, which means it doesn't have a `name` member. But in practice that's definitely not the case - with numpy ops the arrays are just numpy arrays, where `dtype` is a particular kind of type object.

Should the ArrayXd type annotation here be fixed, and if fixed what would it look like? Or should I just put in a command to ignore typing?

svlandeg:
> The remaining issue is that for ArrayXd types, the `dtype` is supposedly a string, which means it doesn't have a `name` member. But in practice that's definitely not the case - with numpy ops the arrays are just numpy arrays, where `dtype` is a particular kind of type object.

Can't we just remove the `.name` bit in all three places?

polm:
That works too!

PR #573, opened 2022-01-17
Do not use an explicit loop in unflatten

danieldk:
...to avoid many small GPU <-> CPU transfers.

I still have to benchmark this change to see if this leads to a meaningful training time difference.

### DIFF
--- a/thinc/backends/ops.py
+++ b/thinc/backends/ops.py
@@ -247,13 +247,14 @@ class Ops:
         """The reverse/backward operation of the `flatten` function: unflatten
         a large array into a list of arrays according to the given lengths.
         """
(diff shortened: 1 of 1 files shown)


danieldk:

honnibal (approved):

svlandeg:
@danieldk : looks good! Feel free to merge in once you're done with your benchmarks and everything looks good :-)

danieldk:
Just a status update: I want to benchmark this a little more.

danieldk:
I did more benchmarks and this change seems to make training slightly slower (2-3%) and inference slightly faster by the same amount. Since a declarative style is nicer in general and since this should scale better (the looping over the array will get more expensive with longer lengths arrays), I'll merge this.

PR #579, opened 2022-01-25
Speed up maxout by exploiting parallelism better

danieldk:
I was profiling the parser refactor and noticed that the maxout kernel is taking more time than I expected:

Screen Shot 2022-01-25 at 14 45 07

The maximum amount of parallelism by the kernel is determined by its batch size. However, this leaves the GPU underused. E.g. by default our amount of parallelism is nr_blocks: 128 * nr_threads_per_block: 128 = 16384, which is much larger than the typical batch size.

This PR changes the maxout kernel to parallelize at the output level (each thread computes one output). This makes the maxout kernel about 4.7 times faster with a batch size of 1024 on a RTX 2060 Super:

Screen Shot 2022-01-25 at 14 44 55

I haven't updated the backprop variant yet, since it barely appears in the profiles.

### DIFF
--- a/thinc/backends/_custom_kernels.cu
+++ b/thinc/backends/_custom_kernels.cu
@@ -88,26 +88,23 @@ void maxout(float* best, int* which,
 {
     int _loop_start = blockIdx.x * blockDim.x + threadIdx.x;
     int _loop_stride = blockDim.x * gridDim.x;
-    for (int b = _loop_start; b < B; b += _loop_stride)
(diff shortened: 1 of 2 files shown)


honnibal:
Awesome!

danieldk:
Added a test which checks `which` output as well.

danieldk:

danieldk:
The test failure is unrelated, but interesting (inconsistent results in the grad scaler test).

danieldk:

PR #580, opened 2022-01-28
Remove deadline of test_scale_random_inputs

danieldk:
By default the hypothesis deadline is 200ms. The test_scale_random_inputs test would frequently fail because on the first run the deadline was exceeded because GPU initialization takes some time.


danieldk:

PR #581, opened 2022-01-31
Add expand_window.v2 with support for Ragged

danieldk:
This version of expand_window uses the new seq2col lengths
functionality.


honnibal:
Do we need the second version here? What if we changed `expand_window` to work over a Union[List[Floats2d], Ragged]`?

danieldk:
> Do we need the second version here? What if we changed `expand_window` to work over a Union[List[Floats2d], Ragged]`?

Works for me too, I wasn't to sure how much we want to overload the existing version. I'll make a new PR.

danieldk:
Closed in favor of #582.

PR #582, opened 2022-01-31
Add support for Ragged to expand_window.v1

danieldk:
Used type variables, so that the caller can be sure that when
they put in a type, they get it out as well.﻿

### DIFF
--- a/thinc/layers/expand_window.py
+++ b/thinc/layers/expand_window.py
@@ -5,20 +5,18 @@ from ..config import registry
 from ..types import Floats2d, Ragged
 
 
-InOutT = TypeVar("InOutT", Floats2d, Ragged)
+InT = TypeVar("InT", Floats2d, Ragged)
 
 
 @registry.layers("expand_window.v1")
-def expand_window(window_size: int = 1) -> Model[InOutT, InOutT]:
+def expand_window(window_size: int = 1) -> Model[InT, InT]:
     """For each vector in an input, construct an output vector that contains the
     input and a window of surrounding vectors. This is one step in a convolution.
     """
     return Model("expand_window", forward, attrs={"window_size": window_size})
 
 
-def forward(
-    model: Model[InOutT, InOutT], X: InOutT, is_train: bool
-) -> Tuple[InOutT, Callable]:
+def forward(model: Model[InT, InT], X: InT, is_train: bool) -> Tuple[InT, Callable]:
     if isinstance(X, Ragged):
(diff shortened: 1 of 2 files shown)


thinc/layers/expand_window.py
```diff
@@ -1,31 +1,58 @@
...
 
 from ..model import Model
 from ..config import registry
-from ..types import Floats2d
+from ..types import Floats2d, Ragged
 
 
-InT = Floats2d
-OutT = Floats2d
+InOutT = TypeVar("InOutT", Floats2d, Ragged)
```
svlandeg, line 8: In these cases we'd just call the type `InT`, like we do for `residual` for instance, we'd just have `Model[InT, InT]`.

thinc/layers/expand_window.py
```diff
@@ -1,31 +1,58 @@
...
         Y = model.ops.tile(X, (nW * 2) + 1)
 
-    def backprop(dY: OutT) -> InT:
+    def backprop(dY: Floats2d) -> Floats2d:
         return model.ops.backprop_seq2col(dY, nW)
 
     return Y, backprop
+
+
+def _expand_window_ragged(model: Model[Ragged, Ragged], Xr: Ragged):
```
svlandeg, line 44:
```suggestion
def _expand_window_ragged(model: Model[Ragged, Ragged], Xr: Ragged) -> Tuple[Ragged, Callable]::
```

svlandeg (commented):
Looks good! I just had a few minor comments.

thinc/layers/expand_window.py
danieldk: Fixed.

thinc/layers/expand_window.py
danieldk: Fixed.

svlandeg:
Now I wonder whether we should bump this version, too, because technically it did change / supports extended functionality...

danieldk:
> Now I wonder whether we should bump this version, too, because technically it did change / supports extended functionality...

That is https://github.com/explosion/thinc/pull/581 :). We are just extending the API, so from the perspective of semantic versioning, it's fine to keep the version the same I think. But I also see the point that people might program against this newer version and then it doesn't work with older `thinc` versions.

danieldk:

svlandeg (approved):
Internally discussed to keep the version when it's just "type widening" - so fine to merge as is.

PR #583, opened 2022-02-02
Add the `normalize` option to the Softmax layer

danieldk:
When this option is disabled, unnormalized probabilities (logits) are
returned during inference.

I was not too sure if adding a keyword argument requires making a
new version of the layer. The default behavior is unchanged.

Using unnormalized probabilities during inference can give nice
speedups. E.g. on my M1 Pro, a German tagging/parsing pipeline is
~27% faster with unnormalized probabilities due to not having to
use the expensive `expf` function.

### DIFF
--- a/thinc/layers/softmax.py
+++ b/thinc/layers/softmax.py
@@ -43,10 +43,14 @@ def forward(model: Model[InT, OutT], X: InT, is_train: bool) -> Tuple[OutT, Call
         model.inc_grad("W", model.ops.gemm(dY, X, trans1=True))
         return model.ops.gemm(dY, W)
 
+    def backprop_unnormalized(dY: InT):
+        msg = "backprop is not supported for an unnormalized Softmax layer"
+        raise ValueError(msg)
+
     if normalized:
         return Y, backprop
     else:
-        return Y, lambda dY: []
+        return Y, backprop_unnormalized
 
 
 def init(
--- a/thinc/layers/softmax.py
+++ b/thinc/layers/softmax.py
@@ -26,12 +26,12 @@ def Softmax(
         init=partial(init, init_W, init_b),
         dims={"nO": nO, "nI": nI},
         params={"W": None, "b": None},
-        attrs={"normalize": normalize},
+        attrs={"normalize_softmax": normalize},
     )
 
 
 def forward(model: Model[InT, OutT], X: InT, is_train: bool) -> Tuple[OutT, Callable]:
-    normalized = model.attrs["normalize"] or is_train
+    normalized = model.attrs["normalize_softmax"] or is_train
     W = cast(Floats2d, model.get_param("W"))
     b = cast(Floats1d, model.get_param("b"))
     Y = model.ops.affine(X, W, b)
--- a/thinc/layers/softmax.py
+++ b/thinc/layers/softmax.py
@@ -18,7 +18,7 @@ def Softmax(
     *,
     init_W: Callable = zero_init,
     init_b: Callable = zero_init,
-    normalize: bool = True,
+    normalize_outputs: bool = True,
 ) -> Model[InT, OutT]:
     return Model(
         "softmax",
@@ -26,7 +26,7 @@ def Softmax(
         init=partial(init, init_W, init_b),
         dims={"nO": nO, "nI": nI},
         params={"W": None, "b": None},
-        attrs={"normalize_softmax": normalize},
+        attrs={"normalize_softmax": normalize_outputs},
     )
 
 
--- a/thinc/layers/softmax.py
+++ b/thinc/layers/softmax.py
@@ -18,6 +18,24 @@ def Softmax(
     *,
     init_W: Callable = zero_init,
     init_b: Callable = zero_init,
+) -> Model[InT, OutT]:
+    return Model(
+        "softmax",
+        forward,
+        init=partial(init, init_W, init_b),
+        dims={"nO": nO, "nI": nI},
+        params={"W": None, "b": None},
+        attrs={"normalize_softmax": True},
+    )
+
+
+@registry.layers("Softmax.v2")
+def Softmax_v2(
+    nO: Optional[int] = None,
+    nI: Optional[int] = None,
+    *,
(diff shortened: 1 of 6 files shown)


honnibal:
This is great -- the attribute gives runtime control of this, which another layer wouldn't let you do. Nice!

Just make the attr name more specific though, so that it's easier to manipulate the attribute within `model.walk()` from the top-level model.

danieldk:
> Just make the attr name more specific though, so that it's easier to manipulate the attribute within model.walk() from the top-level model.

Ah, yes, there may be name clashes. I have renamed it to `normalize_softmax`.

honnibal:
Yeah that's great. The attribute names basically have a global namespace when you're doing `model.walk()` so it's better to be a bit more specific. But yeah great idea!

thinc/layers/softmax.py
```diff
@@ -17,29 +17,36 @@ def Softmax(
...
     def backprop(dY: InT) -> OutT:
         model.inc_grad("b", dY.sum(axis=0))
         model.inc_grad("W", model.ops.gemm(dY, X, trans1=True))
         return model.ops.gemm(dY, W)
 
-    return Y, backprop
+    if normalized:
+        return Y, backprop
+    else:
+        return Y, lambda dY: []
```
honnibal, line 49:
```suggestion
        return Y, lambda dY: []
```

Better to return `ops.xp.zeroslike(dY)` or raise an error I think. The empty list will give type errors or confusing situations.

danieldk: Raised an error, better to notice when someone tries to backprop into an unnormalized softmax.

We could also consider computing the normalized probabilities when someone tries to backprop. It would give the best of both worlds: fast, unnormalized softmax during inference and training still works.

thinc/layers/softmax.py
```diff
@@ -17,29 +17,40 @@ def Softmax(
     nI: Optional[int] = None,
     *,
     init_W: Callable = zero_init,
-    init_b: Callable = zero_init
+    init_b: Callable = zero_init,
+    normalize: bool = True,
```
svlandeg, line 21: I would consider renaming this argument, too. Because in other layers where the constructor takes a `normalize` argument, it defines whether or not a `LayerNorm` should be chained to the layer. Could we call this something like `normalize_outputs`?

thinc/tests/layers/test_softmax.py
```diff
@@ -0,0 +1,23 @@
...
+    dtype="f",
+)
+
+
+def test_unnormalized_softmax_backprop():
+    model = Softmax(normalize=False)
+    model.initialize(inputs, outputs)
+    _, backprop = model(inputs, is_train=False)
+    with pytest.raises(ValueError, match="backprop is not supported"):
+        backprop(OPS.xp.zeros_like(outputs))
```
svlandeg, line 23: Maybe this test could also double check that the backprop does in fact work when `is_train=True`?

svlandeg (commented):
Looks great! I would in fact bump the version though. You expect the same configuration to work across any spaCy/Thinc version, but if users create a config using a v1 with the new argument, then try running it on an older installation, it will unexpectedly crash. Instead, they should use v2 for the new argument and update the older installation when they want to use that v2.

thinc/tests/layers/test_softmax.py
danieldk: I think this should already be covered by the tests in `test_layers_api`?

thinc/layers/softmax.py
danieldk: Fixed.

thinc/tests/layers/test_softmax.py
svlandeg: Is it? I meant specifically to test when `normalize_outputs=False` but `is_train=True`, which is a sort of special case because of the `or` logic.

danieldk: Ah, now I understand, good idea!

danieldk: Added a check. Also named the layer with the `normalize_outputs` argument `Softmax.v2`.

thinc/layers/softmax.py
```diff
@@ -17,29 +17,58 @@ def Softmax(
...
+
+
+@registry.layers("Softmax.v2")
+def Softmax_v2(
+    nO: Optional[int] = None,
+    nI: Optional[int] = None,
+    *,
+    init_W: Callable = zero_init,
+    init_b: Callable = zero_init,
+    normalize_outputs: bool = True,
```
kadarakos, line 39: Wonder whether `v2` should also have `temperature`. I usually think of `softmax` having temperature, but not sure if its important to have it in the implementation. Just wondering, because in `Pytorch` users can add it easily, but in `thinc` I feel like users have to deal with the derivative themselves and that might not be that nice. Not sure.

danieldk: I think it would be nice to have temperature! I think it should be a separate PR, but we could add this to v2 before cutting a new release?

Related: it would also be nice to have label smoothing in the cross entropy loss.

kadarakos: Okok, let's have maybe two separate small PRs then for label-smoothing and temperature!

website/docs/api-layers.md
```diff
@@ -442,19 +442,49 @@ between 0 and 1, so each vector can be interpreted as a probability
...
+- **Parameters:** <ndarray shape="nO, nI">W</ndarray>,
+  <ndarray shape="nO,">b</ndarray>
+
+</inline-list>
+
+A dense layer with a softmax activation. This is usually used as a prediction
+layer. Vectors produced by the softmax function sum to 1, and have values
+between 0 and 1, so each vector can be interpreted as a probability
+distribution. In contrast to the `Softmax` function, `Softmax_v2` also supports
+outputting unnormalized probabilities during inference.
```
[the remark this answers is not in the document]
danieldk: Added rationale (though formulated slightly differently).

svlandeg:
Merging - we can continue the discussion over at https://github.com/explosion/thinc/pull/589

PR #585, opened 2022-02-04
Ragged: avoid concatenate when computing sequence spans

danieldk:
When profiling the parser refactor, I found that quite a lot of time is spent on initializing the arc eager oracle. Turns out that this (indirectly) does a lot of indexing on `Ragged`.

In `Ragged`, `concatenate` is used for computing the end indices of sequences from their start indices. This creates a new copy of an array that is largely identical to the array of start indices.

This PR changes `Ragged` to compute a single array for start/end indices, avoiding `concatenate` during the array construction.

Before this change:

Screen Shot 2022-02-04 at 11 28 59

After this change:

Screen Shot 2022-02-04 at 11 34 37

Both the time spent in `init_gold` and `Ragged.__getitem__` is reduced by quite a bit.

### DIFF
--- a/thinc/types.py
+++ b/thinc/types.py
@@ -841,7 +841,7 @@ class Ragged:
     data: Array2d
     lengths: Ints1d
     data_shape: Tuple[int, ...]
-    cumsums: Optional[Ints1d] = None
+    starts_ends: Optional[Ints1d] = None
 
     def __init__(self, data: _Array, lengths: Ints1d):
         self.lengths = lengths
@@ -874,8 +874,7 @@ class Ragged:
             return Ragged(self.data[s:e], self.lengths[index : index + 1])
         elif isinstance(index, slice):
             lengths = self.lengths[index]
-            cumsums = self._get_cumsums()
-            start = cumsums[index.start - 1] if index.start >= 1 else 0
+            start = ends[index.start - 1] if index.start >= 1 else 0
             end = start + lengths.sum()
             return Ragged(self.data[start:end].reshape(self.data_shape), lengths)
         else:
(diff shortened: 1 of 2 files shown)


thinc/types.py
```diff
@@ -874,8 +874,7 @@ def __getitem__(self, index: Union[int, slice, Array1d]) -> "Ragged":
             return Ragged(self.data[s:e], self.lengths[index : index + 1])
         elif isinstance(index, slice):
             lengths = self.lengths[index]
-            cumsums = self._get_cumsums()
-            start = cumsums[index.start - 1] if index.start >= 1 else 0
+            start = ends[index.start - 1] if index.start >= 1 else 0
```
danieldk, line 877: Just noticed that this can be simplified by using `starts`.

svlandeg (approved):
Neat little trick! I had a detailed look and I can't find any reason why this wouldn't work, but I'll leave this unmerged for a bit longer to give anyone else a chance to review if they like.

danieldk:

thinc/types.py
```diff
@@ -841,7 +841,7 @@ class Ragged:
     data: Array2d
     lengths: Ints1d
     data_shape: Tuple[int, ...]
-    cumsums: Optional[Ints1d] = None
+    starts_ends: Optional[Ints1d] = None
```
svlandeg, line 844: Should this be underscored? It shouldn't really be publicly accessed, should it? Now I'm also worrying about bwd-compat if we merge this change, though I don't think anyone would have been accessing this field directly?

danieldk: Yeah, it would probably be better. Interestingly `cumsums` was made public relatively recently in https://github.com/explosion/thinc/commit/4e8cf6f68628f563c509d4cb7b9057fe0b82cc4a.

svlandeg: Oh. pinging @adrianeboyd in that case - it looks like there was a good reason to do this.

adrianeboyd: I think Daniël's getting years mixed up, that wasn't that recent?

adrianeboyd: It's possible newer versions of pydantic don't have the same behavior or someone else can figure out how to get it to work, but I don't know any more details than what's mentioned in the PR.

danieldk: Having extra fields is disallowed here (which is probably good?):

https://github.com/explosion/thinc/blob/95b1ec978091431994e6ab8a7b366e177fa4a5f8/thinc/util.py#L427

But it seems like we cannot change this specifically for the `Ragged` class. I am not sure if I understand the issue correctly, but this may be due to:

https://github.com/samuelcolvin/pydantic/issues/986

danieldk: I could add a `cumsums` property to ensure we are not breaking the API.

svlandeg: Discussed internally, we can keep this as is. This wasn't really part of the public API and isn't documented as such.

PR #587, opened 2022-02-05
CategoricalCrossEntropy.get_grad: fix invariants check

danieldk:
get_grad checked the guesses/target invariants as follows:

```
if guesses.any() > 1 or guesses.any() < 0:
```

However, `guesses.any()` only has two possible outputs, True or False,
which are reinterpreted as 1 or 0, so the condition is always false.

This change updates the conditions to actually check that guesses
and target are in [0, 1].


svlandeg (commented):
Could you add a test that failed to fail before this PR?

danieldk:
> Could you add a test that failed to fail before this PR?

Fixed!

svlandeg (approved):
Great!

PR #589, opened 2022-02-09
Softmax_v2: add support for softmax with temperature

danieldk:
Temperature can be used to change the confidence/entropy of a model.

### DIFF
--- a/thinc/backends/ops.py
+++ b/thinc/backends/ops.py
@@ -602,7 +602,16 @@ class Ops:
         else:
             return 1 - Y ** 2
 
-    def softmax(self, x: FloatsT, *, inplace: bool = False, axis: int = -1) -> FloatsT:
+    def softmax(
+        self,
+        x: FloatsT,
+        *,
+        inplace: bool = False,
+        axis: int = -1,
+        temperature: float = 1.0,
+    ) -> FloatsT:
+        if temperature != 1.0:
+            x = x / temperature
         maxes = self.xp.max(x, axis=axis, keepdims=True)
         shifted = x - maxes
         new_x = self.xp.exp(shifted)
@@ -622,7 +631,12 @@ class Ops:
         new_x /= summed
         return new_x
 
-    def backprop_softmax(self, Y: FloatsT, dY: FloatsT, *, axis: int = -1) -> FloatsT:
+    def backprop_softmax(
+        self, Y: FloatsT, dY: FloatsT, *, axis: int = -1, temperature: float = 1.0
+    ) -> FloatsT:
+        if temperature != 1.0:
+            dY = dY / temperature
+
         dX = Y * dY
         dX -= Y * dX.sum(axis=axis, keepdims=True)
         return dX
--- a/thinc/layers/softmax.py
+++ b/thinc/layers/softmax.py
@@ -59,11 +59,8 @@ def forward(model: Model[InT, OutT], X: InT, is_train: bool) -> Tuple[OutT, Call
     b = cast(Floats1d, model.get_param("b"))
     Y = model.ops.affine(X, W, b)
 
-    if temperature != 1.0:
-        Y /= temperature
-
     if normalize:
-        Y = model.ops.softmax(Y)
+        Y = model.ops.softmax(Y, temperature=temperature)
 
     def backprop(dY: InT) -> OutT:
         if temperature != 1.0:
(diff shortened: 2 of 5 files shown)


danieldk:
Rebased.

thinc/layers/softmax.py
```diff
@@ -25,7 +25,7 @@ def Softmax(
         init=partial(init, init_W, init_b),
         dims={"nO": nO, "nI": nI},
         params={"W": None, "b": None},
-        attrs={"normalize_softmax": True},
+        attrs={"normalize_softmax": True, "softmax_temperature": 1.0},
```
svlandeg, line 28: Ah it annoys me slightly that one attr is `XXX_softmax` and the other `softmax_XXX`. We can probably rename either of the two, as the first one has not been part of a release yet

thinc/layers/softmax.py
```diff
@@ -37,26 +37,38 @@ def Softmax_v2(
     init_W: Callable = zero_init,
     init_b: Callable = zero_init,
     normalize_outputs: bool = True,
+    temperature=1.0,
```
kadarakos, line 40: Hmm, actually I'm not sure whether the `temperature` is better here in the `Softmax` layer itself or whether it should descend into the `softmax` ops themselves. So I'm not sure where people use `softmax` with temperature outside of the `Softmax` layer, but for example in an `attention` layer we can make the `softmax` sharper for more `hard` decisions. Also maybe we will implement the straight-through gumbel-softmax at some point and then that would not necessarily run the `Softmax` layer, maybe just the `softmax` op itself right?

danieldk: I think this would lead to us handling temperature half in the op (for forward) and half in the op (for backward), since the backward is tailored to the cross-entropy loss case and doesn't use the `backprop_softmax` op (which is for softmax as an activation).

Doing it in the layer would mean that we could always push handling down to the op in the future if we want to, since it wouldn't change the API. However, if we put it in the op now, we have to commit to that API.

thinc/layers/softmax.py
danieldk: I think then having `softmax` at the beginning would be nicer.

danieldk: Renamed.

thinc/layers/softmax.py
danieldk: Doing it in the op would allow more efficient implementations though (getting the maximum value and applying temperature could be fused into one loop).

I still propose to do it as-is now and move it down to the op when needed.

danieldk: Added a commit that applies temperature in the ops. We can always drop the commit before merging the PR if we don't like it.

thinc/tests/backends/test_ops.py
```diff
@@ -593,6 +596,38 @@ def test_softmax_works_inplace(ops, X):
...
+    Yt.backward(dYt)
+
+    return cast(Floats2d, torch2xp(Yt)), cast(Floats2d, torch2xp(Xt.grad))
+
+
+@pytest.mark.skipif(not has_torch, reason="needs PyTorch")
+@pytest.mark.parametrize("ops", ALL_OPS)
+@pytest.mark.parametrize("temperature", [0.5, 1.0, 2.0])
+def test_softmax_temperature(ops, temperature):
+    X = ops.xp.arange(-10, 10, 0.2, dtype="f").reshape(10, 10)
```
kadarakos, line 619: Should `X` be also parametrized?

thinc/backends/ops.py
```diff
@@ -602,7 +602,16 @@ def dtanh(self, Y: FloatsT, *, inplace: bool = False) -> FloatsT:
...
+    def softmax(
+        self,
+        x: FloatsT,
+        *,
+        inplace: bool = False,
+        axis: int = -1,
+        temperature: float = 1.0,
+    ) -> FloatsT:
+        if temperature != 1.0:
+            x = x / temperature
```
svlandeg, line 614: We need to guard against `temperature` 0.0 as well - probably by throwing an error?

thinc/layers/softmax.py
```diff
@@ -37,26 +37,35 @@ def Softmax_v2(
...
```
svlandeg, line 40:
```suggestion
    temperature: float = 1.0,
```

svlandeg (commented):
Looks good in general, only had a few more small comments. And we probably want to run the naming by Matt :-)

thinc/backends/ops.py
danieldk: Yeah, that would be good to validate. We should probably also reject negative values.

Fixed and added test.

PR #590, opened 2022-02-15
Update example-notebook

Jette16:
Updated the example notebook with following changes:

- Made indentation consistent in notebook 00
- Changed install statements --> removed references to alpha and dev versions
- Updated `mathy` to `mathy_core` in notebook 06
- Deleted notebook `parallel_training_ray` as agreed in Slack
- Checked if everything can be executed & cleared outputs (except notebook 03)
- Checked if something should be updated for latest Thinc version (but here, I found nothing relevant; I checked it against git and website code)


svlandeg (approved):
Looks great, thanks so much for going through these in detail! 🙏

PR #594, opened 2022-02-18
Config: do not sort positional arguments

danieldk:
In a configuration such as the following:

```
[model]

[model.chain]
@layers = "chain.v1"

[model.chain.*.hashembed]
@layers = "HashEmbed.v1"
nO = 8
nV = 8

[model.chain.*.expand_window]
@layers = "expand_window.v1"
window_size = 1
```

The positional arguments in the chain were sorted, changing the chain order. This change masks the names of positional arguments. Since `sorted` is a stable sort, the order of positional arguments is retained.

### DIFF
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -465,7 +465,7 @@ def mask_positional_args(name: str) -> List[Optional[str]]:
 
     stable_name = cast(List[Optional[str]], name.split("."))
 
-    # Remove names of sections that are a positional arugment.
+    # Remove names of sections that are a positional argument.
     for i in range(1, len(stable_name)):
         if stable_name[i - 1] == "*":
             stable_name[i] = None
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -342,7 +342,7 @@ class Config(dict):
         sort_map = {section: i for i, section in enumerate(self.section_order)}
         sort_key = lambda x: (
             sort_map.get(x[0].split(".")[0], len(sort_map)),
-            mask_positional_args(x[0]),
+            _mask_positional_args(x[0]),
         )
         return dict(sorted(data.items(), key=sort_key))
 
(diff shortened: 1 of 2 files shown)


thinc/config.py
```diff
@@ -456,6 +459,20 @@ def from_disk(
         return self.from_str(text, interpolate=interpolate, overrides=overrides)
 
 
+def mask_positional_args(name: str) -> List[Optional[str]]:
```
svlandeg, line 462: Perhaps make this a private function?

thinc/config.py
```diff
@@ -456,6 +459,20 @@ def from_disk(
...
+    """Create a section name representation that masks names
+    of positional arguments to retain their order in sorts."""
+
+    stable_name = cast(List[Optional[str]], name.split("."))
+
+    # Remove names of sections that are a positional arugment.
+    for i in range(1, len(stable_name)):
+        if stable_name[i - 1] == "*":
+            stable_name[i] = None
```
svlandeg, line 471: I guess you prefer using `None` here to avoid any (however unlikely) clashes with names of other arguments? It would make the typing a little cleaner to have only strings, and I get a little nervous running the `None` values through `sort`, but I guess it's really just fine.

thinc/config.py
```diff
@@ -456,6 +459,20 @@ def from_disk(
         return self.from_str(text, interpolate=interpolate, overrides=overrides)
 
 
+def mask_positional_args(name: str) -> List[Optional[str]]:
+    """Create a section name representation that masks names
+    of positional arguments to retain their order in sorts."""
+
+    stable_name = cast(List[Optional[str]], name.split("."))
+
+    # Remove names of sections that are a positional arugment.
```
svlandeg, line 468:
```suggestion
    # Remove names of sections that are a positional argument.
```

thinc/config.py
danieldk: Done.

svlandeg (approved):
Looks good, thanks again for the bug fix!

PR #595, opened 2022-02-21
Label-smoothing

kadarakos:
Suggestion for the implementation of `label-smoothing`: https://proceedings.neurips.cc/paper/2019/file/f1748d6b0fd9d439f71450117eba2725-Paper.pdf . In the `CategoricalCrossentropy`, whenever the `truths` are one-hot (generated by `to_categorical`) the smoothing is applied if the smoothing parameter `alpha` larger than 0.


thinc/loss.py
```diff
@@ -46,11 +46,14 @@ def __init__(
         names: Optional[List[str]] = None,
         missing_value: Optional[Union[str, int]] = None,
         neg_prefix: Optional[str] = None,
+        alpha: float = 0.0
```
danieldk, line 49: `alpha` is not very descriptive. Maybe `label_smoothing` or `label_smoothing_prob`?

thinc/loss.py
```diff
@@ -368,6 +374,19 @@ def _make_mask_by_value(truths, guesses, missing_value) -> Floats2d:
     return mask
 
 
+def to_smooth(truths: Floats2d, alpha: float):
```
danieldk, line 377: I think it is nicer/simpler to add this to `to_categorical`. E.g.:

```
def to_categorical(Y: IntsXd, n_classes: Optional[int] = None, label_smoothing: float = 0.0) -> FloatsXd:
    xp = get_array_module(Y)
    if n_classes is None:
        n_classes = int(numpy.max(Y) + 1)  # type: ignore
    label_distr = xp.full((n_classes, n_classes), label_smoothing / (n_classes - 1))
    xp.fill_diagonal(label_distr, 1 - label_smoothing)
    return label_distr[Y]
```

thinc/loss.py
```diff
@@ -101,6 +104,9 @@ def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]
         if truths.ndim != guesses.ndim:
             # transform categorical values to one-hot encoding
             truths = to_categorical(cast(Ints1d, truths), n_classes=guesses.shape[-1])
+            # Only smooth one-hot encoding
+            if self.alpha > 0:
```
danieldk, line 108: We should probably throw an exception if label smoothing was enabled but not applied (e.g. because the user provided a truth matrix).

danieldk (changes requested):

thinc/loss.py
kadarakos: `smooth_prob`?

thinc/loss.py
danieldk: I think raising an exception would be the best, since this would be a programming error (use of incompatible arguments).

thinc/loss.py
danieldk: It's called `label_smoothing` in Tensorflow/Keras. If we need something shorter, I like `smoothing`.

thinc/loss.py
```diff
@@ -46,11 +46,13 @@ def __init__(
         names: Optional[List[str]] = None,
         missing_value: Optional[Union[str, int]] = None,
         neg_prefix: Optional[str] = None,
+        label_smoothing: Optional[float] = 0.0
```
svlandeg, line 49: Make sure to always format with `black` before submitting - I'm pretty sure that'll add a `,` at the end of this line.

thinc/util.py
```diff
@@ -203,12 +203,21 @@ def copy_array(dst: ArrayXd, src: ArrayXd) -> None:  # pragma: no cover
...
+) -> FloatsXd:
     xp = get_array_module(Y)
     if n_classes is None:
         n_classes = int(numpy.max(Y) + 1)  # type: ignore
     # Unfortunately, cupy does not support put_along_axis.
-    return xp.eye(n_classes, dtype="float32")[Y]
+    one_hot = xp.eye(n_classes, dtype="float32")[Y]
+    if label_smoothing > 0.0:
+        z = label_smoothing / (n_classes - 1)
+        xp.maximum(0, one_hot - label_smoothing - z, out=one_hot)
```
danieldk, line 218: I like the implementation that I suggested more. I had to read this implementation several times to understand how it works. The other implementation is just: every probability is `label_smoothing / (n_classes - 1)`, except the diagonal, which is `1 - label_smoothing`.

This also does many more allocations and operations.

The other implementation:

* Allocate + fill`label_distr`
* Fill diagonal
* Allocate + fill `label_dist[Y]`

This implementation:

* Allocate + fill identity matrix
* Allocate + fill `one_hot`
* Allocate and compute `one_hot - label_smoothing`
* Allocate and compute `one_hot - label_smoothing - z`
* Compute maximum 0 and previous step of `one_hot`
* Add `z` to `one_hot`

danieldk: We should probably also check that `label_smoothing >= 0` and `label_smoothing < 1`.

thinc/util.py
```diff
@@ -203,12 +203,22 @@ def copy_array(dst: ArrayXd, src: ArrayXd) -> None:  # pragma: no cover
...
+) -> FloatsXd:
+    assert label_smoothing >= 0 and label_smoothing < 1
     xp = get_array_module(Y)
     if n_classes is None:
         n_classes = int(numpy.max(Y) + 1)  # type: ignore
-    # Unfortunately, cupy does not support put_along_axis.
-    return xp.eye(n_classes, dtype="float32")[Y]
+    label_distr = xp.full((n_classes, n_classes), label_smoothing / (n_classes - 1))
+    xp.fill_diagonal(label_distr, 1 - label_smoothing)
+    if Y.dtype.itemsize == 4:  # type: ignore
```
danieldk, line 217: Why do you want to size the floats to the integer width? This changes the API of `to_categorical` (since it would always return `float32` before). But the width of the integer is also not really related to the width of the labels?

If we want to provide flexibility, `to_categorical` could take an optional `dtype` argument and pass that to `xp.full`.

kadarakos: I'm not sure why is it supposed to always return `float32`. A bunch of the `ops` in `backend` are using the `like` mechanism right? So if the func gets `float32` it supposed to return `float32` also same for `float64`. Since `to_categorical: Ints1d -> Floats2d` we can't get keep the type, but I was just thinking that the most intuitive behavior is to keep the width. But I'll revert it to the previous version of always returning `float32`.

danieldk: I don't see the integer width as related to the floating point width. Even if you are doing math in `float64` for higher precision, 32-bit integers are fine for most occasions such as class labels, indices, lengths, etc. 32-bit is also the default integer size on e.g. CUDA GPUs.

I agree that it would be nice if the function could also return `float64`, but I think it makes more sense then to add a `dtype` argument. Then a caller could do the right thing based on the information that the caller has.

kadarakos: I think its good enough to stay with `float32`, because the only other value we could accept is `float64` atm. (right? :D )

thinc/loss.py
```diff
@@ -46,11 +46,13 @@ def __init__(
         names: Optional[List[str]] = None,
         missing_value: Optional[Union[str, int]] = None,
         neg_prefix: Optional[str] = None,
+        label_smoothing: float = 0.0,
```
danieldk, line 49: `SequenceCategoricalCrossentropy` probably needs a similar argument?

thinc/loss.py
```diff
@@ -46,11 +46,13 @@ def __init__(
...
     ):
         self.normalize = normalize
         self.names = names
         self.missing_value = missing_value
         self.neg_prefix = neg_prefix
+        self.label_smoothing = label_smoothing
```
danieldk, line 55: Would be nice to have a test for label smoothing.

danieldk (commented):
We probably also need `SequenceCategoricalCrossentropy.v3` and `CategoricalCrossentropy.v3`?

thinc/loss.py
kadarakos: Pushed tests for `label_smoothing` in the `test_utils` for the test of `to_categorical`.

thinc/loss.py
kadarakos: added

danieldk:

danieldk (approved):
Looks good!

thinc/util.py
danieldk: Yeah, no surprises. We can always choose to add a `dtype` argument later if necessary.

thinc/loss.py
```diff
@@ -179,24 +210,26 @@ def __init__(
...
         missing_value: Optional[Union[str, int]] = None,
         neg_prefix: Optional[str] = None,
+        label_smoothing: float = 0.0,
     ):
         self.cc = CategoricalCrossentropy(
             normalize=False,
             names=names,
             missing_value=missing_value,
             neg_prefix=neg_prefix,
+            label_smoothing=label_smoothing
```
svlandeg:
```suggestion
            label_smoothing=label_smoothing,
```

thinc/loss.py
```diff
@@ -241,6 +274,23 @@ def configure_SequenceCategoricalCrossentropy_v2(
...
+    missing_value: Optional[Union[str, int]] = None,
+    neg_prefix: Optional[str] = None,
+    label_smoothing: float = 0.0
+) -> SequenceCategoricalCrossentropy:
+    return SequenceCategoricalCrossentropy(
+        normalize=normalize, names=names,
+        missing_value=missing_value,
+        neg_prefix=neg_prefix,
+        label_smoothing=label_smoothing
+    )
```
svlandeg, line 292: Could you run `black` on all the files of the PR?

thinc/util.py
```diff
@@ -203,12 +203,20 @@ def copy_array(dst: ArrayXd, src: ArrayXd) -> None:  # pragma: no cover
         numpy.copyto(dst, src)  # type: ignore
 
 
-def to_categorical(Y: IntsXd, n_classes: Optional[int] = None) -> FloatsXd:
+def to_categorical(
+    Y: IntsXd,
+    n_classes: Optional[int] = None,
```
svlandeg, line 208: I think we should make `n_classes` and `label_smoothing` keyword-only arguments. Unfortunately this would be breaking for `n_classes`, but we could do it at least for `label_smoothing` already.

thinc/util.py
```diff
@@ -203,12 +203,20 @@ def copy_array(dst: ArrayXd, src: ArrayXd) -> None:  # pragma: no cover
...
+    label_smoothing: float = 0.0,
+) -> FloatsXd:
+    assert label_smoothing >= 0 and label_smoothing < 0.5
```
svlandeg, line 211: 0.5 seems like a very high value for label smoothing? Is that realistic?

website/docs/api-loss.md
```diff
@@ -57,8 +57,18 @@ Calculate the scalar loss. Typically returns a float.
...
 
+A flexible implementation of the common categorical cross-entropy loss that
+works on various data types. The **Guesses** should represent probabilities and
+are expected to be in the range of <tt>[0, 1]</tt>. They can both represent
+exclusive classes from multi-class cross-entropy (generally coming from a
+<tt>softmax</tt> layer) or could be classwise binary decisions for multi-label
+cross-entropy (<tt>sigmoid</tt> layer). The **Truths** are most commonly
+provided as labels in <tt>Ints1d</tt>, <tt>List[int]</tt> or <tt>List[str]</tt>
+format. Alternatively, users can provide a <tt>Floats2d</tt> which to encode
+label-confidence.
```
svlandeg, line 68: In a text like this, I would format these differently:
```suggestion
A flexible implementation of the common categorical cross-entropy loss that
works on various data types. The `guesses` should represent probabilities and
are expected to be in the range of <tt>[0, 1]</tt>. They can both represent
exclusive classes from multi-class cross-entropy (generally coming from a
<tt>softmax</tt> layer) or could be classwise binary decisions for multi-label
cross-entropy (<tt>sigmoid</tt> layer). The `truths` are most commonly
provided as labels in <tt>Ints1d</tt>, <tt>List[int]</tt> or <tt>List[str]</tt>
format. Alternatively, users can provide a <tt>Floats2d</tt> which to encode
label-confidence.
```
I'm also unsure about the rendering of the `<tt>` tags within the text. Can't we just use backticks for those? At any rate, they don't seem to be appropriate for marking up words like `softmax`, cf https://github.com/explosion/thinc/tree/master/website#tt-type-annotation.

Also, this text should be moved outside of the `<inline-list>` scope.

website/docs/api-loss.md
```diff
@@ -81,22 +91,29 @@ normalize = true
...
 </grid>
 
-| Argument       | Type          |  Description                                      |
-| -------------- | ------------- | ------------------------------------------------- |
-| _keyword-only_ |               |                                                   |
-| `normalize`    | <tt>bool</tt> | Normalize and divide by number of examples given. |
+| Argument          | Type                     |  Description                                                        |
+| ----------------- | ------------------------ | ------------------------------------------------------------------- |
+| _keyword-only_    |                          |                                                                     |
+| `names`           | <tt>List[str]</tt>       | Label names. Has to be provided when using with List[str] as truths. |
```
svlandeg, line 97: Can you run `prettier` formatting on the `md` files?

website/docs/api-loss.md
```diff
@@ -81,22 +91,29 @@ normalize = true
...
 - **Guesses:** <tt>List[Floats2d]</tt>
-- **Truths:** <tt>List[Union[Ints1d, Floats2d]]</tt>
+- **Truths:** <tt>List[Union[Ints1d, List[int], List[str], Floats2d]]</tt>
 - **Gradient:** <tt>List[Floats2d]</tt>
 - **Loss:** <tt>List[float]</tt>
 
 </inline-list>
 
+This loss runs the <tt>CategoricalCrossentropy<\tt> over a <tt>List<\tt> of
+**Guesses** and **Truths**.
```
svlandeg, line 115: Same comment about formatting

svlandeg (commented):
It will be great to have optional label smoothing, thanks for adding this Ákos!

I mainly had some house-keeping nitpicks ;-)

thinc/util.py
svlandeg: Also, this shouldn't be an `assert`, but a proper value check + raising an informative error, because the value can be a mistake by a user of Thinc. `assert` should only be used to check internal consistencies, not for catching wrong input.

website/docs/api-loss.md
```diff
@@ -81,22 +91,29 @@ normalize = true
...
-| Argument       | Type          |  Description                                      |
-| -------------- | ------------- | ------------------------------------------------- |
-| _keyword-only_ |               |                                                   |
-| `normalize`    | <tt>bool</tt> | Normalize and divide by number of examples given. |
+| Argument          | Type                     |  Description                                                        |
+| ----------------- | ------------------------ | ------------------------------------------------------------------- |
+| _keyword-only_    |                          |                                                                     |
+| `names`           | <tt>List[str]</tt>       | Label names. Has to be provided when using with List[str] as truths. |
+| `normalize`       | <tt>bool</tt>            | Normalize and divide by number of examples given.                   |
+| `neg_prefix`      | <tt>str</tt>             | Symbol that indicates that a label is negative e.g. "!sci-fi".      |
```
svlandeg, line 99: Alternatively: "Prefix used to indicate that ..."

website/docs/api-loss.md
```diff
@@ -81,22 +91,29 @@ normalize = true
...
+| `missing_value`   | <tt>Union[str, int]</tt> | Symbol for "missing value" among the labels.                        |
```
svlandeg, line 100: Alternatively: "Specific label that indicates the value is missing and should not be considered for training/evaluation purposes, e.g. empty string `""` or `0`." ?

thinc/loss.py
svlandeg: I guess I would intuitively call this `epsilon`, but we should probably follow whatever convention is most used.

svlandeg: And `epsilon` invokes other meanings when thinking about losses and comparing float values...

svlandeg: (just rambling out loud here)

thinc/util.py
svlandeg: I agree that it's the appropriate mathematical upperbound, it just feels like way too drastic.

Actually, just checked over at TF and they use the interval [0, 1] there. That's just ... silly. Anyway I guess 0.5 is fine, though I couldn't imagine anything above 0.2 being really useful in actual applications.

website/docs/api-loss.md
kadarakos: I'll swap all of them to backticks, makes more sense I agree.

thinc/util.py
```diff
@@ -203,12 +203,26 @@ def copy_array(dst: ArrayXd, src: ArrayXd) -> None:  # pragma: no cover
...
+            "but {} was given".format(label_smoothing)
+        )
     xp = get_array_module(Y)
     if n_classes is None:
         n_classes = int(numpy.max(Y) + 1)  # type: ignore
-    # Unfortunately, cupy does not support put_along_axis.
-    return xp.eye(n_classes, dtype="float32")[Y]
+    label_distr = xp.full(
+        (n_classes, n_classes), label_smoothing / (n_classes - 1), dtype="float32"
+    )
```
svlandeg, line 223: Let's add a sanity check for `n_classes` here as well, basically that it's bigger than 1, otherwise raise an error?

website/docs/api-loss.md
```diff
@@ -58,12 +58,21 @@ Calculate the scalar loss. Typically returns a float.
...
 </inline-list>
 
+A flexible implementation of the common categorical cross-entropy loss that
+works on various data types. The `guesses` should represent probabilities and
+are expected to be in the range of `[0, 1]`. They can both represent exclusive
+classes from multi-class cross-entropy (generally coming from a `softmax` layer)
+or could be classwise binary decisions for multi-label cross-entropy (`sigmoid`
+layer). The `truths` are most commonly provided as labels in `Ints1d`,
+`List[int]` or `List[str]` format. Alternatively, users can provide a `Floats2d`
+which to encode label-confidence.
```
svlandeg, line 74: There's something wrong with this last sentence, it doesn't quite parse?

thinc/util.py
```diff
@@ -203,12 +203,26 @@ def copy_array(dst: ArrayXd, src: ArrayXd) -> None:  # pragma: no cover
...
+    Y: IntsXd,
+    n_classes: Optional[int] = None,
+    *,
+    label_smoothing: float = 0.0,
+) -> FloatsXd:
+    if not 0.0 <= label_smoothing < 0.5:
+        raise ValueError(
+            "label_smoothing should be greater or "
+            "equal to 0.0 and less than 0.5, "
+            "but {} was given".format(label_smoothing)
```
svlandeg, line 216: we can just use an f-string here:
```suggestion
            f"but {label_smoothing} was given"
```

website/docs/api-loss.md
```diff
@@ -81,22 +90,29 @@ normalize = true
...
+| _keyword-only_    |                          |                                                                                                                                                  |
+| `names`           | <tt>List[str]</tt>       | Label names. Has to be provided when using with List[str] as truths.                                                                             |
+| `normalize`       | <tt>bool</tt>            | Normalize and divide by number of examples given.                                                                                                |
+| `neg_prefix`      | <tt>str</tt>             | Prefix used to indicate that a label is negative e.g. "!sci-fi".                                                                                 |
+| `missing_value`   | <tt>Union[str, int]</tt> | Specific label that indicates the value is missing and should not be considered for training/evaluation purposes, e.g. empty string `""` or `0`. |
+| `label_smoothing` | <tt>float</tt>           | Smoothing-coefficient for label-smoothing in range of [0, 0.5].                                                                                  |
```
svlandeg, line 100:
```suggestion
| `label_smoothing` | <tt>float</tt>           | Smoothing-coefficient for label-smoothing in range of [0, 0.5[.                                                                                  |
```

svlandeg (approved):
I think this is good to go?

PR #596, opened 2022-02-22
test_softmax_temperature: use assert_allclose from ops

danieldk:
This avoids that the test fails with CupyOps.


danieldk:

kadarakos (commented):
Nice catch!

kadarakos (approved):

PR #597, opened 2022-02-22
Improve bound and type checks in CUDA kernels

danieldk:
This PR combines some improvements that should remove most accidental out of bounds reads/writes:

* Check that floating point kernel inputs have dtype `float32`.
* Check that integer kernel inputs have dtype `int32`.
* Check in elementwise backprop kernels that the shapes of `X` and `Y` match that of `dY` (where applicable).
* Check that output arrays have the correct shape when they are provided.
* Check that lengths are `>= 0` and that lengths sum up to the number of instances.
* Check that `which` used in `backprop_maxout` and `backprop_reduce_max` are in bounds.

The `out` option of `mish` is replaced by `inpace`, to give `mish` the same interface as other elementwise kernels. This should not break the API, since `_custom_kernels` is a private module.

### DIFF
--- a/thinc/backends/_custom_kernels.py
+++ b/thinc/backends/_custom_kernels.py
@@ -489,7 +489,11 @@ def _check_lengths(lengths, n_elems: int):
 
 
 def _check_which(which, B: int, I: int, P: int, check_values: bool = False):
-    assert which.dtype == "int32", "which should be encoded as 32-bit integers"
-    assert which.shape == (B, I), "which has incorrect shape"
+    assert (
+        which.dtype == "int32"
+    ), "maximum index (which) should be encoded as 32-bit integers"
+    assert which.shape == (B, I), "maximum index (which) has incorrect shape"
     if check_values:
-        assert cupy.all((which >= 0) & (which < which_len)), "which value out of bounds"
+        assert cupy.all(
+            (which >= 0) & (which < which_len)
+        ), "maximum index (which) value out of bounds"
--- a/thinc/backends/_custom_kernels.py
+++ b/thinc/backends/_custom_kernels.py
@@ -73,6 +73,8 @@ def clipped_linear(
     threads_per_block=128,
     num_blocks=128,
 ):
+    assert X.dtype == "float32", "CUDA clipped_linear kernel can only handle float32"
+
     out = X
     if not inplace:
         out = cupy.zeros_like(X, dtype="f")
@@ -85,6 +87,8 @@ def clipped_linear(
 
 
 def gelu(X, inplace=False, threshold=6.0, threads_per_block=128, num_blocks=128):
+    assert X.dtype == "float32", "CUDA gelu kernel can only handle float32"
+
     out = X
     if not inplace:
         out = cupy.zeros_like(X, dtype="f")
@@ -104,11 +108,11 @@ def check_seq2col_lengths(lengths, B):
 
 
 def seq2col(X, nW, *, lengths=None, out=None, threads_per_block=128, num_blocks=128):
+    assert X.dtype == "float32", "CUDA seq2col kernel can only handle float32"
+
     B = X.shape[0]
     nF = nW * 2 + 1
     I = X.shape[1]
-
-    assert X.dtype == "float32", "CUDA seq2col kernel can only handle float32"
 
     lengths = check_seq2col_lengths(lengths, B)
     nL = lengths.shape[0]
@@ -127,6 +131,8 @@ def seq2col(X, nW, *, lengths=None, out=None, threads_per_block=128, num_blocks=
 
 
 def maxout(X, out=None, threads_per_block=128, num_blocks=128):
+    assert X.dtype == "float32", "CUDA maxout kernel can only handle float32"
+
     B, I, P = X.shape
 
     out_shape = (B, I)
@@ -135,7 +141,8 @@ def maxout(X, out=None, threads_per_block=128, num_blocks=128):
         which = cupy.zeros(out_shape, dtype="i")
     else:
         best, which = out
-        assert maxes.shape == out_shape, "maxes has incorrect shape"
+        assert best.dtype == "float32", "CUDA maxout kernel can only handle float32"
+        assert best.shape == out_shape, "best has incorrect shape"
         assert which.shape == out_shape, "which has incorrect shape"
 
     maxout_kernel((num_blocks,), (threads_per_block,), (best, which, X, B, I, P))
@@ -143,6 +150,8 @@ def maxout(X, out=None, threads_per_block=128, num_blocks=128):
 
 
 def mish(X, inplace=False, threshold=5, threads_per_block=128, num_blocks=128):
+    assert X.dtype == "float32", "CUDA mish kernel can only handle float32"
+
     out = X
     if not inplace:
         out = cupy.zeros_like(X, dtype="f")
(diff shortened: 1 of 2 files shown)


svlandeg (commented):
Happy to see this kind of PR!

About the use of `assert` versus an `if`+`raise`. I would not use `assert` if the wrong input can be caused by a user of Thinc. The fact that we're throwing user-facing error messages here implies to me that maybe `assert` is less applicable? To me, `assert` statements should only be used on internal code paths, to signal a bug within the library itself, not to check for user input.

thinc/backends/_custom_kernels.py
```diff
@@ -363,3 +475,35 @@ def hash(ids, seed, out=None, threads_per_block=128, num_blocks=128):
...
+        ), "maximum index (which) value out of bounds"
+
+def _check_which_reduce_max(which, shape: Tuple, lengths=None):
+    assert (
+        which.dtype == "int32"
+    ), "maximum index (which) should be encoded as 32-bit integers"
+    assert which.shape == shape, "maximum index (which) has incorrect shape"
+    if lengths is not None:
+        print(which.shape)
+        print(lengths.shape)
```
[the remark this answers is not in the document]
danieldk: Whoops, that's a leftover.

danieldk:
> About the use of `assert` versus an `if`+`raise`. I would not use `assert` if the wrong input can be caused by a user of Thinc. The fact that we're throwing user-facing error messages here implies to me that maybe `assert` is less applicable? To me, `assert` statements should only be used on internal code paths, to signal a bug within the library itself, not to check for user input.

`_custom_kernels` is a private module, but I think there are several different cases:

* Checking types should probably be assertions, because if the wrong type of array is passed, this is an error in `CupyOps`, which should dispatch to a kernel when the types are correct or to `Ops` otherwise.
* Incorrect shapes should probably be exceptions, unless we start checking shapes in `Ops`.
* Checking `lengths`/`which` should be exceptions. I don't think it makes sense to check these in `Ops`, since CuPy/NumPy implementations do bounds checks and otherwise we would be doing the same checks twice.
* Some kernels have an `out` type. However, since `_custom_kernels` is a private module and these `out` arguments are not used by `CupyOps`, maybe they should be removed altogether?

danieldk:
Incorrect shapes, `which`, `lengths` are now exceptions. I have also added missing dispatch to some `CupyOps` methods, for when our CUDA kernels cannot handle a data type.

danieldk:
Putting this in draft, the number of changes is quite large, so I think we should postpone this for after the upcoming release, so that it gets more testing.

thinc/backends/_custom_kernels.py
```diff
@@ -273,64 +299,80 @@ def backprop_gelu(
...
 
 
-def backprop_mish(dY, X, out=None, threshold=5, threads_per_block=128, num_blocks=128):
-    if out is None:
-        out = cupy.empty_like(X)
+def backprop_mish(
+    dY, X, inplace: bool = False, threshold=5, threads_per_block=128, num_blocks=128
+):
+    _check_array(dY)
+    _check_array(X, dY.shape)
```
kadarakos, line 322: To me it might be clearer to have something like `_check_compatible_shapes(X, Y)` or if its too long then maybe `_check_compatible(X, Y)`, when the question is whether the shape of `X` and `Y` are the same.

thinc/backends/cupy_ops.py
```diff
@@ -218,28 +226,52 @@ def seq2col(self, seq, nW, *, lengths=None):
...
+            return _custom_kernels.reduce_mean(X, lengths)
+        else:
+            super().reduce_mean(X, lengths)
 
     def backprop_reduce_mean(self, d_means, lengths):
-        return _custom_kernels.backprop_reduce_mean(d_means, lengths)
+        if d_means.dtype == "float32" and lengths.dtype == "int32":
+            return _custom_kernels.backprop_reduce_mean(d_means, lengths)
+        else:
+            super().reduce_mean(d_means, lengths)
```
kadarakos, line 250: should this be `backprop_reduce_mean` ?

danieldk: Whoops 😅, thanks! I'll fix that tomorrow.

danieldk: Fixed.

thinc/backends/_custom_kernels.py
danieldk: Renamed to `_check_compatible`. `_check_compatible_shapes` is too specific, since you can also leave out the shape and it will just check the `dtype`.

danieldk:
I have retarget this to `develop` and removed the draft status.

thinc/backends/_custom_kernels.py
```diff
@@ -363,3 +409,56 @@ def hash(ids, seed, out=None, threads_per_block=128, num_blocks=128):
         (out, ids, out_size, in_size, ids.shape[0], seed),
     )
     return out
+
+
+def _check_compatible(out, shape: Optional[Tuple] = None):
+    assert out.dtype == "float32", "CUDA kernel can only handle float32"
+    if shape is not None and out.shape != shape:
+        msg = f"array has incorrect shape, expected: {shape}, was: {out.shape}"
+        raise ValueError(msg)
```
svlandeg, line 418: It's not a big deal because it's an internal function, but I'd make the name a bit more explicit, maybe something like `_check_cuda_compatible` or `_check_float32_compat` or such. Because it comes across as a bit weird otherwise when this is called with just one argument (-> compatible to _what_?)

thinc/backends/_custom_kernels.py
```diff
@@ -124,88 +125,101 @@ def seq2col(X, nW, *, lengths=None, out=None, threads_per_block=128, num_blocks=
     return out
 
 
-def maxout(X, out=None, threads_per_block=128, num_blocks=128):
+def maxout(X, threads_per_block=128, num_blocks=128):
```
svlandeg, line 128: I wonder whether we should define most of these functions with keyword-only arguments, in this case it'd be
```
maxout(X, *, threads_per_block=128, num_blocks=128):
```
Things would break (if at all) in a more straightforward way if we change the list of arguments like this PR does.

(To be clear: I'm not saying things do break or that there's an issue with this PR, but it's just something that came to mind seeing the arguments getting changed for many of these functions. It feels like we probably always want to call these functions with named arguments.)

thinc/backends/_custom_kernels.py
```diff
@@ -363,3 +409,56 @@ def hash(ids, seed, out=None, threads_per_block=128, num_blocks=128):
...
+        raise ValueError("all sequence lengths must be >= 0")
+    if cupy.sum(lengths) != n_elems:
+        raise IndexError("lengths must sum up to the batch size")
+
+
+def _check_which_maxout(which, B: int, I: int, P: int):
+    shape = (B, I)
+    assert (
+        which.dtype == "int32"
+    ), "maximum index (which) should be encoded as 32-bit integers"
```
svlandeg, line 433: This would probably be a bit more readible if it uses a `msg` variable like the `if` on the next line does -> then the `assert` can go on one line.

thinc/backends/_custom_kernels.py
```diff
@@ -363,3 +409,56 @@ def hash(ids, seed, out=None, threads_per_block=128, num_blocks=128):
...
+    )
+    if cupy is not None
+    else None
+)
+
+
+def _check_which_reduce_max(which, shape: Tuple, lengths):
+    assert (
+        which.dtype == "int32"
+    ), "maximum index (which) should be encoded as 32-bit integers"
```
svlandeg, line 459: Same comment about one-line-`assert`

svlandeg (approved):
Looks good to me, though I'm not 100% confident with all the details of this PR. Added some nitpicks as well ;-)

thinc/backends/_custom_kernels.py
danieldk: I think this needs be renamed later anyway, since the `float32` check is replaced in the next PR which makes the kernels generic (also supporting `float64`).

svlandeg: Right! Ok sure, then we can leave as is for now.

thinc/backends/_custom_kernels.py
danieldk: I agree. Especially since some of these functions (like the backprop kernels) take multiple arguments that are tensors with the same shape, it's very easy to accidentally make an error and only notice it when training models. I'll make this my todo item after the generic kernel PR.

thinc/backends/_custom_kernels.py
danieldk: Fixed

thinc/backends/_custom_kernels.py
danieldk: Fixed

thinc/backends/_custom_kernels.py
danieldk: See https://github.com/explosion/thinc/pull/621

PR #601, opened 2022-03-02
Fix Ops.backprop_reduce_max

danieldk:
Suppose that we have a sequence of length 2 and dimensionality 3:

```
1: a b c
2: d e f
```

The `which` array specifies for each dimension which sequence contained
the maximum value. For example, if `which = [0, 1, 0]`, the maximum values
were:

```
a e c
```

However, the `Ops` implementation of this function was using which to
select columns. So,

```
a b a
d e d
```

would be selected. This change rectifies this, to select per column the
row indicated in `which`.


svlandeg (approved):
Nice find!

PR #603, opened 2022-03-07
custom_kernels: make all CUDA kernels generic

danieldk:
This change makes all CUDA kernels generic using C++ templates, so that they can be used for float and double arrays.


thinc/backends/_custom_kernels.cu
```diff
@@ -3,8 +3,27 @@
 // This pattern ensures that all of the loop values are visited once, no matter
 // what grid parameters are used for the function.
 
-extern "C" __global__
-void seq2col(float* output, const float* X, const int* lengths,
+// We cannot include CUDA header for mathetmatical constants, since it requires
```
svlandeg, line 6:
```suggestion
// We cannot include CUDA header for mathematical constants, since it requires
```

thinc/backends/_custom_kernels.cu
```diff
@@ -3,8 +3,27 @@
...
+
+template <>
+struct Constants<double> {
+    static constexpr double INV_SQRT_2 = 0.7071067811865475;
+    static constexpr double INV_SQRT_2PI = 0.3989422804014327;
+};
+
+template <>
+struct Constants<float> {
+    static constexpr float INV_SQRT_2 = 0.70710677;
```
svlandeg, line 20: Slightly surprised that this isn't `0.70710678` considering the `double` value of this one?

thinc/backends/_custom_kernels.py
```diff
@@ -74,26 +134,40 @@ def clipped_linear(
...
-    )
+        out = cupy.zeros_like(X)
+    if X.dtype == "float32":
+        clipped_linear_kernel_float(
+            (num_blocks,),
+            (threads_per_block,),
+            (out, X, slope, offset, min_val, max_val, X.size),
+        )
+    else:
+        clipped_linear_kernel_double(
```
svlandeg, line 149: Should we add an additional `assert` here on the `dtype`? Or is that not necessary because we've already done the `_is_float_array` check?

thinc/backends/cupy_ops.py
```diff
@@ -170,39 +170,43 @@ def backprop_clipped_linear(
...
-        if X.dtype == "float32" and dY.dtype == "float32":
+        if X.dtype == dY.dtype and X.dtype in ("float32", "float64"):
             return _custom_kernels.backprop_hard_swish_mobilenet(dY, X, inplace=inplace)
         else:
             return super().backprop_hard_swish_mobilenet(dY, X, inplace=inplace)
 
     def mish(self, X, threshold=20.0, inplace=False):
-        if X.dtype == "float32":
+        if X.dtype in ("float32", "float64") and not inplace:
             return _custom_kernels.mish(X, inplace=inplace, threshold=threshold)
```
svlandeg, line 186: Why do we need the `not inplace` check here?

thinc/backends/cupy_ops.py
```diff
@@ -170,39 +170,43 @@ def backprop_clipped_linear(
...
+        if X.dtype in ("float32", "float64"):
             return _custom_kernels.swish(X, inplace=inplace, threshold=17.0)
         else:
             return super().swish(X, inplace=inplace)
 
     def backprop_swish(self, dY, X, Y, inplace=False):
-        if X.dtype == "float32" and dY.dtype == "float32":
+        if (
+            X.dtype == dY.dtype
+            and X.dtype == Y.dtype
```
svlandeg, line 207:
```suggestion
            X.dtype == dY.dtype == Y.dtype
```
This works too, right?

thinc/backends/cupy_ops.py
```diff
@@ -226,49 +230,57 @@ def seq2col(self, seq, nW, *, lengths=None):
...
     def reduce_mean(self, X, lengths):
-        if X.dtype == "float32" and lengths.dtype == "int32":
+        if X.dtype in ("float32", "float64") and lengths.dtype == "int32":
             return _custom_kernels.reduce_mean(X, lengths)
         else:
             super().reduce_mean(X, lengths)
 
     def backprop_reduce_mean(self, d_means, lengths):
-        if d_means.dtype == "float32" and lengths.dtype == "int32":
+        if X.dtype in ("float32", "float64") and lengths.dtype == "int32":
```
svlandeg, line 255:
```suggestion
        if d_means.dtype in ("float32", "float64") and lengths.dtype == "int32":
```

svlandeg (commented):
Looks good! Left a few small comments/questions/nitpicks/corrections.

Should we also add some tests to test the functionality with `double` arrays?

thinc/backends/_custom_kernels.cu
danieldk: `0.70710678` is not representable using a 32-bit float.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0f9a148c5d96bf7fe081a15b10676c5b

thinc/backends/_custom_kernels.py
danieldk: Yeah, the idea is that the invariants are already checked at the beginning of each function, so we can rely on `dtype` to be `float64` in that case.

thinc/backends/cupy_ops.py
danieldk: Fixed.

thinc/backends/cupy_ops.py
danieldk: Nice, I didn't know that! Fixed.

danieldk:
> Should we also add some tests to test the functionality with double arrays?

I was looking into that. We already test the activation functions for `float32` and `float64`. I think for most of the other functions we also need generic implementations in `NumpyOps`. I was already looking into that. What do you think, does it make more sense to add that to this PR or wait until I have done generic implementations of `NumpyOps` as well?

svlandeg:
Intuitively I would say it'd make sense to add the tests here, but I don't feel strongly. As long as it's on the TODO list, that's fine by me :-) 

> > Should we also add some tests to test the functionality with double arrays?
> 
> I was looking into that. We already test the activation functions for `float32` and `float64`. I think for most of the other functions we also need generic implementations in `NumpyOps`. I was already looking into that. What do you think, does it make more sense to add that to this PR or wait until I have done generic implementations of `NumpyOps` as well?

danieldk:
> Intuitively I would say it'd make sense to add the tests here, but I don't feel strongly. As long as it's on the TODO list, that's fine by me :-)

I'll add them and disable them for ops that don't support `float64` yet. I'll put this in draft so that it doesn't get merged accidentally.

danieldk:
> Intuitively I would say it'd make sense to add the tests here, but I don't feel strongly. As long as it's on the TODO list, that's fine by me :-)

Done!

Added a small function `ops_with_dtypes`, which we can remove once `NumpyOps` properly supports float64 arrays as well.

danieldk:

PR #607, opened 2022-03-11
to_categorical: permit having one class without label smoothing

danieldk:
Also add tests to verify that the n_classes kwarg is correctly checked.


svlandeg (approved):
Good catch!

PR #614, opened 2022-03-14
class-weights for cross-entropy

kadarakos:
It adds class-weights to the cross-entropy loss to potentially help with class imbalance.


thinc/backends/ops.py
```diff
@@ -131,6 +131,15 @@ def _iter_items():
 
         return SizedGenerator(_iter_items, len(sizes))
 
+    # Computed as in "Logistic regression in rare events data"
+    # G King, L Zeng - Political analysis, 2001
+    def balanced_class_weights(self, labels: Ints1d):
```
danieldk, line 136: Unused?

thinc/loss.py
```diff
@@ -1,3 +1,5 @@
+from math import isclose
```
danieldk, line 1: Unused.

thinc/loss.py
```diff
@@ -47,12 +49,15 @@ def __init__(
         missing_value: Optional[Union[str, int]] = None,
         neg_prefix: Optional[str] = None,
         label_smoothing: float = 0.0,
+        class_weights: Optional[List[float]] = None,
```
danieldk, line 52: It would be nice if it was also possible to pass the class weights as `Floats1d`.

thinc/loss.py
```diff
@@ -47,12 +49,15 @@ def __init__(
...
     ):
         self.normalize = normalize
         self.names = names
         self.missing_value = missing_value
         self.neg_prefix = neg_prefix
         self.label_smoothing = label_smoothing
+        self.class_weights = class_weights
```
danieldk, line 59: Maybe the list -> array conversion should be done here already? Avoids allocating/freeing the same array repeatedly.

thinc/loss.py
```diff
@@ -143,6 +148,25 @@ def get_grad(self, guesses: Floats2d, truths: IntsOrFloatsOrStrs) -> Floats2d:
...
+            if target.shape[1] != len(self.class_weights):
+                raise ValueError(
+                    "The number of classes in the "
+                    "target and in class weights "
+                    "has to be equal, but found "
+                    f"{target.shape[1]} and "
+                    f"{len(self.class_weights)}"
+                )
+            cp = xp.asarray(self.class_weights)
+            cp = xp.tile(cp, (guesses.shape[0], 1))
```
danieldk, line 165: I don't think this is really necessary right? The `cp` array can just be broadcast in

```
sample_weights = (norm_target * cp).sum(axis=1)
```

I think it already works without any changes, but otherwise I think

```
cp = xp.asarray(self.class_weights).reshape(1, target.shape[1])
```

should do the trick? Avoids a potentially large allocation :).

thinc/loss.py
```diff
@@ -143,6 +148,25 @@ def get_grad(self, guesses: Floats2d, truths: IntsOrFloatsOrStrs) -> Floats2d:
...
+            row_sum = xp.expand_dims(xp.sum(target, axis=1), axis=1)
```
danieldk, line 166: You probably don't need the `expand_dims` when using the `keepdims=True` argument of `sum`.

thinc/loss.py
```diff
@@ -143,6 +148,25 @@ def get_grad(self, guesses: Floats2d, truths: IntsOrFloatsOrStrs) -> Floats2d:
...
+            norm_target = target / row_sum
+            sample_weights = (norm_target * cp).sum(axis=1)
+            difference *= xp.expand_dims(sample_weights, 1)
```
danieldk, line 169: This `expand_dims` may also not be necessary by using the `keepdims` argument of `sum`.

thinc/backends/ops.py
kadarakos: it is intended to use if the user would like to compute class weights to provide to the `CategoricalCrossEntropy`. But they can give any other kinds of weights, its just there for convenience.

danieldk: Ah, though I wonder if it doesn't better fit in a utility module, since `Ops` is mostly for low-level array operations.

thinc/loss.py
kadarakos: I was thinking the same but we don't know whether we will get `cupy` or `numpy` later on.

thinc/backends/ops.py
svlandeg: Agreed that moving to `loss.py` would make sense!

thinc/loss.py
kadarakos: What do you think about the current solution @danieldk ?

danieldk: Hmm, then we'd still reallocate it over and over again if the user passes a list. What about something more drastic, like just requiring `class_weights` to be `Floats1d`?

kadarakos: Yeah ok we can do that too, but then still since the `CategoricalCrossEntropy` itself is not bound to be `cupy` or `numpy` and it could get a different type of array at every call, we would still need to reallocate right?

danieldk: Right, but I don't think that happens much in practice. Probably when someone passes in `class_weights` as a CuPy array they'll also use CuPy arrays down the line. I think at this point it probably doesn't make much sense to optimize for mixed use.

thinc/backends/ops.py
```diff
@@ -603,7 +603,7 @@ def dtanh(self, Y: FloatsT, *, inplace: bool = False) -> FloatsT:
             Y += 1.0
             return Y
         else:
-            return 1 - Y ** 2
+            return 1 - Y**2
```
danieldk, line 606: Not sure which variant comes from the old version of `black` and which from the new, but it seems like we are flipping between the formats between different PRs 😅.

thinc/loss.py
```diff
@@ -143,6 +146,26 @@ def get_grad(self, guesses: Floats2d, truths: IntsOrFloatsOrStrs) -> Floats2d:
...
+            if target.shape[1] != len(self.class_weights):
+                raise ValueError(
+                    "The number of classes in the "
+                    "target and in class weights "
+                    "has to be equal, but found "
+                    f"{target.shape[1]} and "
+                    f"{len(self.class_weights)}"
+                )
+            if not is_xp_array(self.class_weights):
+                self.class_weights = xp.asarray(self.class_weights)
```
danieldk, line 163: Maybe we should do this in the constructor, so that it is only done once?

danieldk: Oh, it is only done once. But I guess it'd still make sense to do this in the constructor? Simplifies the invariants within the class.

thinc/loss.py
```diff
@@ -47,12 +47,15 @@ def __init__(
         missing_value: Optional[Union[str, int]] = None,
         neg_prefix: Optional[str] = None,
         label_smoothing: float = 0.0,
+        class_weights: Optional[Union[List[float], Floats1d]] = None,
```
danieldk, line 50: Could you document the parameter in `api-loss.md`?

thinc/backends/ops.py
svlandeg: cf https://github.com/explosion/spaCy/pull/10209 --> the change from the PR seems the right "current" `black` format. We should probably pin 
```
black>=22.0,<23.0
```
and make sure that everyone updates their `black` in their local `venv`

svlandeg (commented):
`master` has changed significantly since this PR was opened, so probably best to merge in the current branch & double check merge conflicts 🙏

PR #618, opened 2022-03-17
NumpyOps: add bound checks

danieldk:
This PR adds various bounds checks to `NumpyOps` to prevent out-of-bound reads and writes:

- Check that `which` indices are in bounds for `backprop_{maxout, reduce_max}`.
- Check that `dY` and `X` shapes match in backprop for elementwise functions.
- Hook up ReLU to the PyTorch comparison tests.
- Check that lengths are valid in `backprop_reduce_{mean, max, sum}`.


danieldk:

honnibal:
Nice! Love the advanced syntax for the `nogil except`. That's either relatively new, or I just never knew about it 🤯 . Either way, neat.

PR #621, opened 2022-03-21
_custom_kernels: make all arguments kwarg-only

danieldk:
Some kernels take multiple arguments, which are all arrays with the same shape. It's easy to accidentally mix up the order and introduce bugs in that way. This change makes all arguments kwarg-only, so that we always specify what we intend to pass at the call site.

Since `_custom_kernels` is an internal module, this does not result in any public API changes.

### DIFF
--- a/thinc/backends/_custom_kernels.py
+++ b/thinc/backends/_custom_kernels.py
@@ -125,6 +125,7 @@ backprop_swish_kernel_float = _get_kernel("backprop_swish<float>")
 
 
 def clipped_linear(
+    *,
     X,
     inplace=False,
     slope=1.0,
@@ -154,7 +155,7 @@ def clipped_linear(
     return out
 
 
-def gelu(X, inplace=False, threshold=6.0, threads_per_block=128, num_blocks=128):
+def gelu(*, X, inplace=False, threshold=6.0, threads_per_block=128, num_blocks=128):
     _is_float_array(X)
 
     out = X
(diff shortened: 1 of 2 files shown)


danieldk:
I didn't specify kwargs for `super()` calls. Since they are currently not tested, we can't be sure that nothing breaks :(.

thinc/backends/_custom_kernels.py
```diff
@@ -154,7 +155,7 @@ def clipped_linear(
     return out
 
 
-def gelu(X, inplace=False, threshold=6.0, threads_per_block=128, num_blocks=128):
+def gelu(*, X, inplace=False, threshold=6.0, threads_per_block=128, num_blocks=128):
```
svlandeg, line 158: I think I would personally opt to not make `X` keyword-only, as the name isn't important and it feels intuitive that the first argument should always be the matrix.

svlandeg (commented):
Thanks for looking into this!

How bwd-compat breaking will this be?

thinc/backends/_custom_kernels.py
danieldk: I first had that as well, but some of the backprop functions also have `X` as an argument, but not as the first argument. But maybe we could make an exception for `X` for the forwards and `dY` for the backwards?

danieldk: > How bwd-compat breaking will this be?

I don't think this should break anything, since we only use these in `CupyOps` and it's a private module.

svlandeg: To me it feels a bit weird to have "required" keyword arguments, i.e. keyword-arguments that don't have a default value. You'd need to know exactly what the name is, like `seq=...` for `seq2col`, where really the name feels arbitrary to me...

danieldk: I have updated the PR to make only non-required arguments kwarg-only.

thinc/backends/_custom_kernels.py
```diff
@@ -459,7 +467,7 @@ def backprop_gelu(
     return out
 
 
-def backprop_maxout(dY, which, P, threads_per_block=128, num_blocks=128):
+def backprop_maxout(dY, which, P, *, threads_per_block=128, num_blocks=128):
```
kadarakos, line 470: If we are dealing with the kwargs already and we are creating potential backward compatibility issues, what about the variable name `P`? To me its not really descriptive. But maybe its descriptive enough because the number of pieces in the maxout layer is `nP` ?

thinc/backends/cupy_ops.py
```diff
@@ -150,8 +150,8 @@ def backprop_clipped_linear(
     ):
         if X.dtype == dY.dtype and X.dtype in ("float32", "float64"):
             return _custom_kernels.backprop_clipped_linear(
-                dY=dY,
-                X=X,
+                dY,
+                X,
                 slope=slope,
```
kadarakos, line 155: Should the `slope`, `offset`, `min_val` and `max_val` become keyword-only?

thinc/backends/_custom_kernels.py
danieldk: > creating potential backward compatibility issues

This PR should not create backwards compat issues, since `_custom_kernels` is a private module. If we'd change the naming in the public ops interface, we'd break backwards compat, not sure if it's worth doing that.

thinc/backends/cupy_ops.py
danieldk: This PR makes them. Or do you mean in `CupyOps`?

thinc/backends/_custom_kernels.py
svlandeg: Yea no I think it's fine as is

svlandeg (approved):
Looks good!

PR #623, opened 2022-03-23
GradScaler: Do not enable when training on CPU

danieldk:
- Disable gradient scaling and emit a warning when but CPU training is
  performed.
- Disable gradient scaling and emit a warning when a PyTorch version
  without gradient scaling is used (rather than raising an exception).

See https://github.com/explosion/spaCy/issues/10527

**Note:** putting this in draft. I'd like to add a similar check and warning to the shim.

### DIFF
--- a/thinc/shims/pytorch_grad_scaler.py
+++ b/thinc/shims/pytorch_grad_scaler.py
@@ -1,6 +1,7 @@
 from typing import Dict, Iterable, List, Union, cast
+import warnings
 
-from ..util import has_torch_amp, is_torch_array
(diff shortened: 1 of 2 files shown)


danieldk:
This turns out to be quite tricky, since Thinc may not be switched to GPU yet when `PyTorchGradScaler` is constructed. So, checking whether we are using a GPU in the constructor may disable gradient scaling unintentionally.

I will close this PR in favor of #624, which replaces the assertion by an exception and describes how this error can be avoided.

PR #624, opened 2022-03-24
Mixed-precision training: improve checks

danieldk:
1. Before this change, using gradient scaling on a non-CUDA tensor would trigger an assertion. However, it is possible to trigger this error outside Thinc by enabling mixed-precision training on a CPU. So this should be a proper exception rather than an `AssertionError`. Besides raising a `ValueError`, the error message is also extended to describe how the error can be avoided.

2.  We checked that gradient scaling is supported when construction a PyTorchGradScaler. However, this gives issues when someone uses a model that was trained with gradient scaling. In such cases, it's safe to contruct a grad scaler, since it is not used. This change moves the check to the actual scaling.

3. Also remove the check that verifies that mixed-precision scaling is available (when enabled) from the constructor of PyTorchShim. PyTorch will at most give a warning when trying to autocast when there is no support.

### DIFF
--- a/thinc/shims/pytorch.py
+++ b/thinc/shims/pytorch.py
@@ -43,11 +43,6 @@ class PyTorchShim(Shim):
         mixed_precision: bool = False,
         grad_scaler: Optional[PyTorchGradScaler] = None,
     ):
-        if mixed_precision and not has_torch_amp:
-            raise ValueError(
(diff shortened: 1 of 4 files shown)


danieldk:
Related issues: https://github.com/explosion/spaCy/issues/10527 and https://github.com/explosion/spaCy/discussions/10543

PR #625, opened 2022-03-29
Lstm benchmark example fix

kadarakos:
The example didn't work when I tried running it and this PR fixes it.

### DIFF
--- a/examples/benchmarks/lstm_tagger.py
+++ b/examples/benchmarks/lstm_tagger.py
@@ -16,15 +16,8 @@ import typer
 import tqdm
 import numpy.random
 from timeit import default_timer as timer
-from thinc.api import (
-    Model,
-    Config,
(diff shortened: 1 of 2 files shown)


examples/benchmarks/lstm_tagger.py
```diff
@@ -16,10 +16,18 @@
...
-from thinc.api import Model, Config, registry, chain, list2padded, with_array
+from thinc.api import (
+    Model,
+    Config,
+    registry,
+    chain,
+    list2padded,
+    with_array,
+    with_padded,
+)
```
svlandeg, line 27: When this happens due to `black` formatting, we'd typically spread the `import` statements over 2 lines "manually"

svlandeg (commented):
Thanks for looking into this Ákos! 🙏

PR #627, opened 2022-03-30
Make NumpyOps CPU kernels generic

danieldk:
This PR makes most CPU kernels generic, so that they can take both `float32` and `float64` arrays (and hopefully in the future `float16`). I experimented with kernels in Cython + fused types and kernels as C++ with templates, I found the C++ template route more promising:

- More compact/ergonomic implementations with fewer compile-time conditionals.
- Opens up the possibility to easily use SIMD intrinsics in the future.

To allow genericity in the `NumpyOps` method arguments, we use:

- Fused types when we require a specific dimensionality;
- `np.ndarray` otherwise.

Some of the kernels are not (yet) made generic:

- `cpu_scatter_add`: needs tests to verify that the op still works
  correctly.
- `cpu_position_encode`: the position_encode op doesn't take float
  array(s).
- lstm kernels: I need to look more deeply into them.

### DIFF
--- a/thinc/backends/cpu_kernels.hh
+++ b/thinc/backends/cpu_kernels.hh
@@ -93,7 +93,7 @@ void cpu_reduce_max(A* maxes__bo, L* which__bo, const A* X__to,
 
         T -= *length;
 
-        memcpy(maxes__bo, X__to, O * sizeof(*maxes__bo));
+        std::memcpy(maxes__bo, X__to, O * sizeof(*maxes__bo));
         X__to += O;
         for (L i = 1; i < *length; ++i) {
             for (L j = 0; j < O; ++j) {
--- a/thinc/backends/cpu_kernels.hh
+++ b/thinc/backends/cpu_kernels.hh
@@ -1,0 +1,394 @@
+#ifndef CPU_KERNELS_HH
+#define CPU_KERNELS_HH
+
+#include <algorithm>
+#include <cmath>
+#include <cstring>
+#include <stdexcept>
+#include <string>
+#include <type_traits>
+
+
+// All elementwise functions, such as most activations, work in-place.
+
+template <typename A, typename L>
+L argmax(A* arr, L len)
+{
+    static_assert(std::is_floating_point<A>::value,
+        "Array should be floating point");
+    static_assert(std::is_integral<L>::value, "Array length should be integral");
+
+    L max = 0;
+    for (L i = 1; i < len; ++i) {
+        if (arr[i] > arr[max]) {
+            max = i;
+        }
+    }
+
+    return max;
+}
+
+template <typename A, typename L>
+void vec_add(A* X, const A* Y, A scale, L N)
+{
+    static_assert(std::is_floating_point<A>::value,
+        "Array should be floating point");
+    static_assert(std::is_integral<L>::value, "Array length should be integral");
+
+    for (L i = 0; i < N; ++i)
+        X[i] += scale * Y[i];
+}
+
+template <typename A, typename L>
+void cpu_maxout(A* best__bo, L* which__bo, const A* cands__bop, L B, L O, L P)
+{
+    static_assert(std::is_floating_point<A>::value,
+        "Array should be floating point");
+    static_assert(std::is_integral<L>::value, "Array length should be integral");
+
+    for (int i = 0; i < B * O; ++i) {
+        which__bo[i] = argmax(cands__bop + i * P, P);
+        best__bo[i] = cands__bop[i * P + which__bo[i]];
+    }
+}
+
+template <typename A, typename L>
+void cpu_backprop_maxout(A* dX__bop, const A* dX__bo, const L* which__bo,
+    L B, L O, L P)
+{
+    static_assert(std::is_floating_point<A>::value,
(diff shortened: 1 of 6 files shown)


thinc/backends/cpu_kernels.hh
```diff
@@ -0,0 +1,394 @@
...
+            continue;
+        else if (*length < 0)
+            throw std::invalid_argument(std::string("all sequence lengths must be >= 0, was: ") + std::to_string(*length));
+        else if (*length > T) {
+            throw std::out_of_range("lengths must sum up to the number of rows");
+        }
+
+        T -= *length;
+
+        memcpy(maxes__bo, X__to, O * sizeof(*maxes__bo));
```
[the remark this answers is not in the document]
danieldk: Good catch! Fixed.

thinc/backends/cpu_kernels.hh
```diff
@@ -0,0 +1,394 @@
...
+template <typename A, typename L>
+void cpu_relu(A* X, L N)
+{
+    static_assert(std::is_floating_point<A>::value,
+        "Array should be floating point");
+    static_assert(std::is_integral<L>::value, "Array length should be integral");
+
+    for (L i = 0; i < N; ++i) {
+        if (X[i] < 0) {
+            X[i] = 0.0;
```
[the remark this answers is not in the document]
danieldk: I don't think `-0.0` has any practical ramifications, but I am all in for erasing it :).

thinc/backends/numpy_ops.pyx
```diff
@@ -93,21 +93,17 @@ class NumpyOps(Ops):
...
-                    data[i] = 0.
-            return Y
+            cpu_relu(<float *>Y.data, <int>Y.size)
+        elif X.dtype == "float64":
+            cpu_relu(<double *>Y.data, <int>Y.size)
         else:
-            return super().relu(X, inplace)
+            return super().relu(X, inplace=inplace)
+
+        return Y
```
svlandeg, line 105: It reads a little weird to me that for 2 out of 3 `if` clauses, the `return` is at the end of the function, and for the other it's immediate.

(the same comment about some of the other function rewrites below)

thinc/backends/numpy_ops.pyx
```diff
@@ -93,21 +93,17 @@ class NumpyOps(Ops):
...
-                Y = X.copy()
-            data = <weight_t*>Y.data
-            for i in range(size):
-                if data[i] < 0:
-                    data[i] = 0.
-            return Y
+            cpu_relu(<float *>Y.data, <int>Y.size)
+        elif X.dtype == "float64":
+            cpu_relu(<double *>Y.data, <int>Y.size)
         else:
```
svlandeg, line 102: I suppose this `else` won't actually happen a lot, right? Otherwise I was going to ask whether it's an issue that we might have already been copying `Y` unnecessarily.

thinc/backends/numpy_ops.pyx
```diff
@@ -117,6 +113,7 @@ class NumpyOps(Ops):
         cdef const weight_t* Y_ptr = <const weight_t*>Y.data
         cdef np.ndarray dX
         if dY.dtype == "float32" and Y.dtype == "float32":
+            dX = _inplace_or_copy(dY, inplace)
```
svlandeg, line 116: You forgot to remove the lines below

thinc/backends/numpy_ops.pyx
```diff
@@ -159,84 +156,143 @@ class NumpyOps(Ops):
...
         cdef np.ndarray Y
+
         if X.dtype == "float32":
-            if inplace:
-                Y = X
-            else:
-                Y = self.xp.empty_like(X)
-            cpu_mish(<float*>Y.data, <float *>X.data, threshold, X.size)
-            return Y
+            Y = _inplace_or_copy(X, inplace)
```
svlandeg, line 196: The original code initialized `Y` to an empty array with same shape as `X` - but here we'd be copying `X` entirely, is that correct?

danieldk: Yes. The kernel now applies the activation in-place, which I did for all element-wise activations. It should not lead to any performance regressions, since most platforms vectorize `memcpy` and before this change we were doing scalar copies in the kernel itself (when values `>= threshold`).

thinc/backends/numpy_ops.pyx
danieldk: Fixed.

thinc/backends/numpy_ops.pyx
danieldk: Fixed by adding the `_inplace_or_copy` inside the branches of the `if` statement.

thinc/backends/numpy_ops.pyx
danieldk: Jikes, thanks!

danieldk:
> Ugh - do we have a flaky test?

Oh no, again the dreadful `Unreliable test timings! On an initial run, this test took XYZms, which exceeded the deadline of 200.00ms`.

svlandeg:
Rerunning it 😐

PR #630, opened 2022-04-01
SWA and Lookahead optimzer

kadarakos:
# Optimizer wrappers

Currently the `Optimizer` is quite large class. It implements a version of weight averaging that starting from the first step it keeps an exponential moving average of the weights. This PR suggests to move this functionality out of the `Optimizer` into the `SWA` class.

The `SWA` is the Stochastic Weight Averaging as implemented in https://arxiv.org/abs/1803.05407. It has attributes:

1. `start_step`: the training-step where to start recording the moving averages from.
2. `freq`: the number of steps between updating the moving average.
3. `lr`: the new learning-rate for the `SWA` steps.

The PR contains also a `Lookahead` optimizer wrapper that implements a similar algorithm to `SWA`. It's from https://arxiv.org/abs/1907.08610. It keeps track of "slow" weights and "fast" weights. Every `freq` train-steps it updates the "slow" weights with an exponential moving average of the "slow" weights and replaces the "fast" with the "slow". Essentially it let's the optimizer run the optimization forward `k` iterates, but then it pulls it back with a factor `pullback`. 

The example `optimizer_wrappers.py` shows how to run the `Lookahead` with `Adam` from a fixed number of epochs and then to swap it to `SWA`. This uses the `SWA.start_swa()` functionality that let's the caller immediately run `SWA` without setting `start_step`.


examples/optimizer_wrappers.py
```diff
@@ -0,0 +1,60 @@
+"""
+PyTorch version: https://github.com/pytorch/examples/blob/master/mnist/main.py
+TensorFlow version: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/examples/tutorials/mnist/mnist.py
```
svlandeg, line 3: This is copy/pasted from `mnist.py`, but I'm not sure we'd want to repeat it here? Or do you want to replace the `mnist.py` example with this one?

kadarakos: Not exactly sure what to do with this example just yet, for now I just kinda meant it as a demonstration for the reviewers of the PR :-).

thinc/optimizers.py
```diff
@@ -354,4 +354,127 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
...
+        lr: FloatOrSeq,
+        *,
+        freq: int = 1,
+        start_step: Optional[int] = 0
+    ):
+        self.optimizer = optimizer
+        self.start_step = start_step
+        self.freq = freq
+        self.lr = lr
+        self.nr_update: Dict[Tuple[int, str], int] = defaultdict(int)
```
danieldk, line 376: Just an idea: I like `collections.Counter` for such purposes, since the name clearly states the semantics. It also has some useful additional methods (which you probably don't need here :)).

thinc/optimizers.py
```diff
@@ -354,4 +354,127 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
...
+        weights: FloatsXd,
+        gradient: FloatsXd,
+        *,
+        lr_scale: float = 1.0,
+    ):
+        self.nr_update[key] += 1
+        nr_update = self.nr_update[key]
+        weights, gradient = self.optimizer(key, weights, gradient, lr_scale=lr_scale)
+        # Update running mean
+        if nr_update == self.start_step:
```
danieldk, line 423: I am not sure if I understand this. SWA is triggered once one of the variables reaches the given number of updates. Wouldn't we want to decide this per variable? If so, then I think changing the condition below to:

```
        if nr_update >= self.start_step and nr_update % self.freq == 0:
             self._update_averages(key, weights)
```

Also removes some state from the class (`self.run_swa`).

thinc/optimizers.py
```diff
@@ -354,4 +354,127 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
...
 
 
+class SWA(object):
+    """
+    After the optimization ran 'start_step' number of steps
+    it switches the learning-rate of the optimizer and starts
+    recording the moving averages of all weights.
+    """
+
+    def __init__(
```
danieldk, line 364: Would be nice here to have documentation of the arguments.

thinc/optimizers.py
```diff
@@ -354,4 +354,127 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
...
+
+class Lookahead(object):
+    """
+    Keeps a "slow" exponential moving average
+    of all "fast" weights. At each 'freq' update
+    it updates the slow-weights and replaces the
+    "fast" weights with them.
+    """
+
+    def __init__(self, optimizer, *, freq: int, pullback: float):
```
danieldk, line 439:
```suggestion
    def __init__(self, optimizer: Optimizer, *, freq: int, pullback: float):
```

Would be nice here to have documentation of the arguments. I was also wondering if there are some reasonable defaults that could be set here?

danieldk (commented):
Looks really nice, also great that these are wrappers for `Optimizer`.

This still needs tests I think? Should be relatively easy to give it checkable weights with a mock optimizer?

danieldk:
Oh, one more thing about wrapping: there are various methods in Thinc and spaCy that take `Optimizer` arguments. Using these classes would not typecheck, since they do not derive from `Optimizer`. I am not sure whether deriving from `Optimizer` does make sense, since that would also inherit a lot of internal state? But ideally, `Optimizer` would be a base class.

thinc/optimizers.py
kadarakos: haha I was thinking about the same thing that I would use a `Counter` to khm count. But yeah, I'm not using anything from it :D

svlandeg (commented):
`master` has changed significantly since this PR was open, so probably best to merge in the current branch & double check merge conflicts 🙏

thinc/optimizers.py
```diff
@@ -354,4 +354,133 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
         )
 
 
+class SWA(object):
```
honnibal, line 357: I think we don't need this inheritance from `object` in Python3

honnibal:
@danieldk I don't think I'll have time to look at this in proper detail, so I'm relying on you to approve it 🙏 😄 

I think the idea is good though, so I'm happy to see it merged once all the details are approved!

danieldk:
> @danieldk I don't think I'll have time to look at this in proper detail, so I'm relying on you to approve it 🙏 😄
> 
> I think the idea is good though, so I'm happy to see it merged once all the details are approved!

Ok, I'll do another review round (probably tomorrow).

thinc/optimizers.py
```diff
@@ -111,8 +129,8 @@ class Optimizer(object):
     schedules: Dict[str, Generator]
     nr_update: Dict[KeyT, int]
     last_seen: Dict[KeyT, int]
-    grad_clip: float
     learn_rate: float
+    grad_clip: float
```
shadeMe, line 133: Why was this reordered?

thinc/optimizers.py
```diff
@@ -354,4 +375,201 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
...
+    ):
+        """
+        Update moving average of weights.
+        """
+        if key not in self.averages:
+            ops = get_array_ops(weights)
+            self.averages[key] = ops.alloc(weights.shape, dtype="float32")
+            self.averages[key] += weights
+        else:
+            self.averages[key] *= self.nr_update[key]
```
danieldk, line 436: I wonder if it makes sense to change this a bit:

* Make a private member `_weight_sums` and only add the weights here.
* Add a property `averages` that returns a dict with `_weight_sums` divided by `nr_update`.

Benefits:

* During training updates are done more often than getting the averages (I think). And you avoid do one multiplication and division less.
* You lose less precision as a result of repeatedly multiplying and dividing (though probably not much of an issue in neural nets).

Downside:

* Takes more memory when retrieving the averages, since you have both the sums and averages in memory.

It's just an idea that we might consider.

thinc/optimizers.py
```diff
@@ -354,4 +375,201 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
...
+        lr_scale: float = 1.0,
+    ):
+        self.nr_update[key] += 1
+        nr_update = self.nr_update[key]
+        weights, gradient = self.optimizer(key, weights, gradient, lr_scale=lr_scale)
+        # Update running mean
+        if nr_update == self.start_step:
+            self._swap_lr()
+        if nr_update >= self.start_step and nr_update % self.freq == 0:
+            self._update_averages(key, weights)
```
danieldk, line 466: If I understand correctly, I think there is an ordering error here. `nr_update[key]` is incremented at the beginning of the function, but `_update_averages` assumes the old count?

thinc/optimizers.py
```diff
@@ -354,4 +375,201 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
         )
 
 
-__all__ = ["Adam", "RAdam", "SGD", "Optimizer", "ADAM_DEFAULTS", "SGD_DEFAULTS"]
+class SWA(OptimizerABC):
+    """
+    After the optimization ran 'start_step' number of steps
```
danieldk, line 380: Add a bibliographic reference is there is one.

thinc/optimizers.py
```diff
@@ -354,4 +375,201 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
...
+
+        return weights, gradient
+
+
+class Lookahead(OptimizerABC):
+    """
+    Keeps a "slow" exponential moving average
+    of all "fast" weights. At each 'freq' update
+    it updates the slow-weights and replaces the
+    "fast" weights with them.
```
danieldk, line 476: Add a bibliographic reference is there is one.

thinc/optimizers.py
```diff
@@ -354,4 +375,201 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
...
+        """
+        Update exponental moving average of weights.
+        """
+        if key not in self.averages:
+            ops = get_array_ops(weights)
+            self.averages[key] = ops.alloc(weights.shape, dtype="float32")
+            self.averages[key] += weights
+        else:
+            average = self.averages[key]
+            self.averages[key] += self.pullback * (weights - average)
```
danieldk, line 512:
```suggestion
            average += self.pullback * (weights - average)
```

(same dict)

thinc/optimizers.py
```diff
@@ -354,4 +375,201 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
...
+        nr_update = self.nr_update[key]
+        weights, gradient = self.optimizer(key, weights, gradient, lr_scale=lr_scale)
+        # Update running mean and return interpolated weights
+        if nr_update % self.freq == 0:
+            self._update_averages(key, weights)
+            weights = self.averages[key]
+        return weights, gradient
+
+
+@registry.optimizers("SWA.v1")
```
danieldk, line 532: Shouldn't the name be `AdamSWA.v1`?

thinc/optimizers.py
```diff
@@ -100,7 +106,22 @@ def SGD(
     )
 
 
-class Optimizer(object):
+class OptimizerABC(Generic[GradT, WeightT]):
```
danieldk, line 109: Since we now have a `v9` branch where we can have larger changes, maybe it's good to discuss the naming and the class hierarchy a bit more internally?

thinc/optimizers.py
danieldk, line 436: Maybe you could look up `self.averages[key]` just once and reuse the dict? It's only a small performance improvement, but why not? 😄

thinc/optimizers.py
```diff
@@ -354,4 +375,201 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
...
+        learn_rate: FloatOrSeqOrStr,
+        *,
+        freq: int = 1,
+        start_step: int = 0,
+        on_plateau: bool = False
+    ):
+        """
+        optimizer: The inner optimizer whose iterates to average.
+        learn_rate: Learning-rate (or schedule) to run with after
+            the averaging is turned on.
```
danieldk, line 404: Add possible values.

kadarakos:
Closed this PR, because over time the priorities shifted.

PR #632, opened 2022-04-11
Reduce unnecessary zero-init'd allocations

shadeMe:
I've added a new keyword arg `uninitialized` to the allocator methods in the `NumpyOps` and `Ops` classes. Also replaced some zero-init'd allocations where the elements in the destination array were immediately replaced.

### DIFF
--- a/thinc/backends/_custom_kernels.py
+++ b/thinc/backends/_custom_kernels.py
@@ -533,7 +533,7 @@ def backprop_reduce_sum(d_sums, lengths, *, threads_per_block=128, num_blocks=12
     O = d_sums.shape[1]
     _check_lengths(lengths, T)
 
-    out = _alloc((T, O), dtype=d_sums.dtype, zeros=True)
+    out = _alloc((T, O), dtype=d_sums.dtype, zeros=False)
 
     if d_sums.dtype == "float32":
         backprop_reduce_sum_kernel_float(
@@ -555,7 +555,7 @@ def backprop_reduce_mean(d_means, lengths, *, threads_per_block=128, num_blocks=
     O = d_means.shape[1]
     _check_lengths(lengths, T)
 
-    out = _alloc((T, O), dtype=d_means.dtype, zeros=True)
+    out = _alloc((T, O), dtype=d_means.dtype, zeros=False)
 
     if d_means.dtype == "float32":
         backprop_reduce_mean_kernel_float(
--- a/thinc/backends/_custom_kernels.py
+++ b/thinc/backends/_custom_kernels.py
@@ -204,7 +204,7 @@ def seq2col(seq, nW, *, lengths=None, threads_per_block=128, num_blocks=128):
     lengths = check_seq2col_lengths(lengths, B)
     nL = lengths.shape[0]
 
-    out = _alloc((B, I * nF), dtype=seq.dtype)
+    out = _alloc((B, I * nF), dtype=seq.dtype, zeros=True)
 
     if seq.size != 0 and lengths.size != 0:
         if seq.dtype == "float32":
@@ -268,7 +268,7 @@ def reduce_sum(X, lengths, *, threads_per_block=128, num_blocks=128):
 
     _check_lengths(lengths, T)
 
-    out = _alloc((B, O), dtype=X.dtype)
+    out = _alloc((B, O), dtype=X.dtype, zeros=True)
 
     if X.dtype == "float32":
         reduce_sum_kernel_float(
@@ -291,7 +291,7 @@ def reduce_mean(X, lengths, *, threads_per_block=128, num_blocks=128):
 
     _check_lengths(lengths, T)
 
-    out = _alloc((B, O), dtype=X.dtype)
+    out = _alloc((B, O), dtype=X.dtype, zeros=True)
 
     if X.dtype == "float32":
         reduce_sum_kernel_float(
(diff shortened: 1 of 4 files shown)


svlandeg (commented):
Great idea! 

Not sure about the naming - `uninitialized` is not the most straightforward to type, but maybe just `empty` is too cryptic?

shadeMe:
> Great idea!
> 
> Not sure about the naming - `uninitialized` is not the most straightforward to type, but maybe just `empty` is too cryptic?

Yeah, I thought `empty` was wasn't explicit enough, especially given that the incorrect usage of this flag will almost certainly lead to memory corruption. I also considered flipping the logic and calling the param `zero_init` but wasn't convinced that it was better.

I'm open to suggestions/switching to `zero_init` if you find that it's a better alternative.

svlandeg (commented):
Looks good to me!

Could you also update the documentation in `/website/docs` ?

website/docs/api-backends.md
```diff
@@ -426,7 +427,8 @@ Y = model.ops.alloc1i(4)  # Ints1d
 | `*shape`       | <tt>int</tt>                              | The shape, one positional argument per dimension.                          |
 | _keyword-only_ |                                           |                                                                            |
 | `dtype`        | <tt>DTypesInt</tt> / <tt>DTypesFloat</tt> | The data type (float type for float methods and int type for int methods). |
-| **RETURNS**    | <tt>ArrayXd</tt>                          | An array of the correct shape and data type, filled with zeros.            |
+| `zeros`        | <tt>bool</tt>                             | Fill the array with zeros (default: `True`).                               |
```
danieldk, line 430: Could you reformat the markdown with [prettier](https://prettier.io)?

danieldk:
There are some allocations in `CupyOps`/`_custom_kernels` that could be updated in a similar way. E.g. the `maxout` kernel wrapper unnecessarily zero-initializes an array.

shadeMe:
`CupyOps` had a fair amount of zero-allocs that could be replaced. These changes pass the tests on the GPU.

danieldk:

thinc/backends/_custom_kernels.py
```diff
@@ -303,8 +317,8 @@ def reduce_max(X, lengths, *, threads_per_block=128, num_blocks=128):
     _check_lengths(lengths, T)
 
     out_shape = (B, O)
-    maxes = cupy.zeros(out_shape, dtype=X.dtype)
-    which = cupy.zeros(out_shape, dtype="i")
+    maxes = _alloc(out_shape, dtype=X.dtype, zeros=False)
+    which = _alloc(out_shape, dtype="i", zeros=False)
```
svlandeg, line 321: Are we sure about this one? `reduce_max` is retrieving values from `maxes` and `which`, no?

thinc/backends/_custom_kernels.py
```diff
@@ -519,7 +533,7 @@ def backprop_reduce_sum(d_sums, lengths, *, threads_per_block=128, num_blocks=12
     O = d_sums.shape[1]
     _check_lengths(lengths, T)
 
-    out = cupy.zeros((T, O), dtype=d_sums.dtype)
+    out = _alloc((T, O), dtype=d_sums.dtype)
```
svlandeg, line 536: I think we can set `zeros=False` here?

svlandeg (commented):
Marked a few cases where I was unsure, but an additional set of eyes would be good!

thinc/backends/_custom_kernels.py
```diff
@@ -190,7 +204,7 @@ def seq2col(seq, nW, *, lengths=None, threads_per_block=128, num_blocks=128):
     lengths = check_seq2col_lengths(lengths, B)
     nL = lengths.shape[0]
 
-    out = cupy.zeros((B, I * nF), dtype=seq.dtype)
+    out = _alloc((B, I * nF), dtype=seq.dtype)
```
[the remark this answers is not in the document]
shadeMe: Can't disagree with that, will do.

thinc/backends/_custom_kernels.py
danieldk: Yes, `reduce_max` should fill all elements (like `maxout`), so `zeros=False` should be correct.

thinc/backends/_custom_kernels.py
```diff
@@ -519,7 +533,7 @@ def backprop_reduce_sum(d_sums, lengths, *, threads_per_block=128, num_blocks=12
     O = d_sums.shape[1]
     _check_lengths(lengths, T)
 
-    out = cupy.zeros((T, O), dtype=d_sums.dtype)
+    out = _alloc((T, O), dtype=d_sums.dtype, zeros=True)
```
danieldk, line 536: `backprop_reduce_mean` should backprop to every sequence item, so I think this can be `zeros=False`.

thinc/backends/_custom_kernels.py
```diff
@@ -541,7 +555,7 @@ def backprop_reduce_mean(d_means, lengths, *, threads_per_block=128, num_blocks=
     O = d_means.shape[1]
     _check_lengths(lengths, T)
 
-    out = cupy.zeros((T, O), dtype=d_means.dtype)
+    out = _alloc((T, O), dtype=d_means.dtype, zeros=True)
```
danieldk, line 558: `backprop_reduce_mean` should backprop to every sequence item, so I think this can be `zeros=False`.

thinc/backends/_custom_kernels.py
svlandeg: Ok cool, thanks for double checking!

danieldk:

danieldk (approved):
Looks great to me! But I'll defer to @svlandeg for a second opinion :).

thinc/backends/_custom_kernels.py
svlandeg: No, looks like you're right, thanks for double checking!

shadeMe:
Whoops! Clicked on the wrong "Re-request review" button. Only mean to do so for @svlandeg.

svlandeg:
Merged!

PR #634, opened 2022-04-14
Doc fixes

shadeMe:
Indicate support for `DLPack` when using TF models ([already implemented](https://github.com/explosion/thinc/blob/master/thinc/util.py#L367)).

Remove the `Distributed Training` section due to poor support for Ray and deleted Colab notebook.


website/docs/usage-training.md
```diff
@@ -293,15 +293,3 @@ values = [0.001, 0.01, 0.1]
...
-
-We expect to recommend [Ray](https://ray.io/) for distributed training. Ray
-offers a clean and simple API that fits well with Thinc's model design. While
-full support is still under development, you can find an example here:
-
-<tutorials header="false">
-
-- parallel_training_ray
-
-</tutorials>
```
svlandeg, line 307: Perhaps we can keep the section and just say "Full support is still under development." and end it at that? We do want to pick this up again some time "soon".

shadeMe: That's indeed a better idea; will fix.

svlandeg:
Thanks!

PR #637, opened 2022-04-19
Fix reductions when applied to zero-length sequences

danieldk:
* A bug was introduced in `NumpyOps` that caused the output pointer not to be incremented for zero-length sequences.
* When using uninitialized arrays, the sum or mean for a zero-length array was not correctly set to zero in `Ops`.

We should also do something about zero-length sequences for `reduce_max`. However, it's unclear what the corresponding `which` should be set to (since `0` is not a valid index for a zero-sized sequence). Maybe we should throw an exception when trying to apply `reduce_max` to a zero-length sequence?

### DIFF
--- a/thinc/backends/cpu_kernels.hh
+++ b/thinc/backends/cpu_kernels.hh
@@ -83,9 +83,7 @@ void cpu_reduce_max(A* maxes__bo, L* which__bo, const A* X__to,
     static_assert(std::is_integral<L>::value, "Array length should be integral");
 
     for (const L* length = lengths__b; length < lengths__b + B; ++length) {
-        if (*length == 0)
-            continue;
-        else if (*length < 0)
+        if (*length <= 0)
             throw std::invalid_argument(std::string("all sequence lengths must be >= 0, was: ") + std::to_string(*length));
         else if (*length > T) {
             throw std::out_of_range("lengths must sum up to the number of rows");
--- a/thinc/backends/cpu_kernels.hh
+++ b/thinc/backends/cpu_kernels.hh
@@ -143,9 +143,7 @@ void cpu_reduce_mean(A* means__bo, const A* X__to, const L* lengths__b,
     static_assert(std::is_integral<L>::value, "Array length should be integral");
 
     for (const L* length = lengths__b; length < lengths__b + B; ++length) {
-        if (*length == 0)
-            continue;
-        else if (*length < 0)
+        if (*length < 0)
             throw std::invalid_argument(std::string("all sequence lengths must be >= 0, was: ") + std::to_string(*length));
         else if (*length > T) {
             throw std::out_of_range("lengths must sum up to the number of rows");
@@ -226,9 +224,7 @@ void cpu_reduce_sum(A* sums__bo, const A* X__to, const L* lengths__b,
     static_assert(std::is_integral<L>::value, "Array length should be integral");
 
     for (const L* length = lengths__b; length < lengths__b + B; ++length) {
-        if (*length == 0)
-            continue;
-        else if (*length < 0)
+        if (*length < 0)
             throw std::invalid_argument(std::string("all sequence lengths must be >= 0, was: ") + std::to_string(*length));
         else if (*length > T) {
             throw std::out_of_range("lengths must sum up to the number of rows");
--- a/thinc/backends/cpu_kernels.hh
+++ b/thinc/backends/cpu_kernels.hh
@@ -141,8 +141,13 @@ void cpu_reduce_mean(A* means__bo, const A* X__to, const L* lengths__b,
     static_assert(std::is_integral<L>::value, "Array length should be integral");
 
     for (const L* length = lengths__b; length < lengths__b + B; ++length) {
-        if (*length < 0)
+        if (*length < 0) {
             throw std::invalid_argument(std::string("all sequence lengths must be >= 0, was: ") + std::to_string(*length));
+        }
+        else if (length == 0) {
+            means__bo += O;
+            continue;
+        }
         else if (*length > T) {
             throw std::out_of_range("lengths must sum up to the number of rows");
         }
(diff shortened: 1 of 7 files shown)


shadeMe:
Are there any contexts where zero-length sequences can be considered valid input? Also, since the `which` array is signed, using `-1` to signify zero-length input sequences might be one possible way to go.

danieldk:
> Are there any contexts where zero-length sequences can be considered valid input?

The floret support in spaCy relies on the mean of a zero-length sequence to be zero.

> Also, since the `which` array is signed, using `-1` to signify zero-length input sequences might be one possible way to go.

Maybe. I think the nice thing about `which` is that you can now use it with e.g. `take_along_axis`. However, `-1` would silently take the last element along a given axis.

danieldk:
@explosion-bot please test_gpu --spacy-branch master

website/docs/api-backends.md
```diff
@@ -1275,7 +1277,9 @@ Backpropagate the `reduce_mean` operation.
 
 </inline-list>
 
-Perform sequence-wise max pooling for data in the ragged format.
+Perform sequence-wise max pooling for data in the ragged format. Zero-length
+sequences are not allowed. A `ValueError` is raised if one any element in
```
shadeMe, line 1281:
```suggestion
sequences are not allowed. A `ValueError` is raised if any element in
```

`...raised if any element...` sounds better, I think.

website/docs/api-backends.md
```diff
@@ -1293,7 +1297,8 @@ Perform sequence-wise max pooling for data in the ragged format.
 
 </inline-list>
 
-Backpropagate the `reduce_max` operation.
+Backpropagate the `reduce_max` operation. A `ValueError` is raised if one any
```
shadeMe, line 1300:
```suggestion
Backpropagate the `reduce_max` operation. A `ValueError` is raised if any
```
Same as above.

website/docs/api-layers.md
```diff
@@ -863,7 +863,8 @@ https://github.com/explosion/thinc/blob/master/thinc/layers/reduce_last.py
 </inline-list>
 
 Pooling layer that reduces the dimensions of the data by selecting the maximum
-value for each feature.
+value for each feature. A `ValueError` is raised if one any element in `lengths`
```
shadeMe, line 866:
```suggestion
value for each feature. A `ValueError` is raised if any element in `lengths`
```

thinc/backends/cpu_kernels.hh
```diff
@@ -83,9 +83,7 @@ void cpu_reduce_max(A* maxes__bo, L* which__bo, const A* X__to,
     static_assert(std::is_integral<L>::value, "Array length should be integral");
 
     for (const L* length = lengths__b; length < lengths__b + B; ++length) {
-        if (*length == 0)
-            continue;
-        else if (*length < 0)
+        if (*length <= 0)
             throw std::invalid_argument(std::string("all sequence lengths must be >= 0, was: ") + std::to_string(*length));
```
shadeMe, line 87:
```suggestion
            throw std::invalid_argument(std::string("all sequence lengths must be > 0, was: ") + std::to_string(*length));
```

Think the message should be updated to `> 0` here.

thinc/backends/ops.py
```diff
@@ -1068,7 +1072,7 @@ def reduce_max(self, X: Floats2d, lengths: Ints1d) -> Tuple[Floats2d, Ints2d]:
         which = self.alloc2i(lengths.shape[0], X.shape[1], zeros=False)
         start = 0
         for i, length in enumerate(lengths):
-            if length < 0:
+            if length <= 0:
                 raise ValueError(f"all sequence lengths must be >= 0, got {length}")
```
shadeMe, line 1076:
```suggestion
                raise ValueError(f"all sequence lengths must be > 0, got {length}")
```

thinc/backends/ops.py
```diff
@@ -1108,6 +1112,9 @@ def backprop_reduce_max(
         dX = self.alloc2f(lengths.sum(), d_maxes.shape[1], dtype=d_maxes.dtype)
         start = 0
         for i, length in enumerate(lengths):
+            if length <= 0:
+                raise ValueError(f"all sequence lengths must be >= 0, got {length}")
```
shadeMe, line 1116:
```suggestion
                raise ValueError(f"all sequence lengths must be > 0, got {length}")
```

svlandeg:
I think Ákos is better positioned to review this one :-)

thinc/backends/cpu_kernels.hh
```diff
@@ -143,9 +141,7 @@ void cpu_reduce_mean(A* means__bo, const A* X__to, const L* lengths__b,
     static_assert(std::is_integral<L>::value, "Array length should be integral");
 
     for (const L* length = lengths__b; length < lengths__b + B; ++length) {
-        if (*length == 0)
-            continue;
-        else if (*length < 0)
+        if (*length < 0)
```
kadarakos, line 144: Not sure if it makes sense, but it could also have a `length == 0` and then it just returns `0` for clarity?

thinc/backends/cpu_kernels.hh
```diff
@@ -226,9 +222,7 @@ void cpu_reduce_sum(A* sums__bo, const A* X__to, const L* lengths__b,
...
```
kadarakos, line 225: Same here with the `length == 0` special case. I was just thinking that because it works differently from reduce max it might be helpful for interpretation, but maybe not.

website/docs/api-backends.md
```diff
@@ -1203,7 +1203,8 @@ Backpropagate the hard Swish MobileNet activation.
 
 </inline-list>
 
-Perform sequence-wise summation for data in the ragged format.
+Perform sequence-wise summation for data in the ragged format. Zero-length
+sequences are reduced to a zero vector.
```
kadarakos, line 1207: to "the" zero vector?

kadarakos (commented):
Only had very minor comments. One extra thing I was wondering about is whether I see it correctly that this throws an uncaught zero-division error when length IS 0 https://github.com/danieldk/thinc/blob/fix-reductions-zero-length/thinc/backends/ops.py#L1105.

thinc/backends/cpu_kernels.hh
danieldk: So, I think you are proposing:

```
else if (length == 0) {
    means__bo += O;
    continue;
}
```

?

kadarakos: yes

kadarakos (approved):

danieldk:
The tensorflow errors are unrelated, see https://github.com/explosion/thinc/pull/653.

PR #639, opened 2022-04-19
Fix PyTorch tensor handling in `CupyOps.asarray`

### LINKED ISSUE #564: Cupy TypeError when using asarray with torch tensors
Hi,

I'm chaining a custom feature extraction model with a PyTorchWrapper model. In the feature extraction part, I'm using asarray function on a torch tensor but as you can see below, cupy raises an error when MemoryPointer is called. Do you have any idea why that line raises an exception?

Thanks,

```
Traceback (most recent call last): 
...
  File "/storage/myuser/myproject/main.py", line 180, in forward
    features.append(model.ops.asarray2f(attrs, dtype="float32"))
  File "/storage/miniconda/envs/safirx-ml/lib/python3.9/site-packages/thinc/backends/ops.py", line 494, in asarray2f
    return cast(Floats2d, self.asarray(data, dtype=dtype))
  File "/storage/miniconda/envs/safirx-ml/lib/python3.9/site-packages/thinc/backends/cupy_ops.py", line 67, in asarray
    pointer = cupy.cuda.MemoryPointer(data.data_ptr())
  File "cupy/cuda/memory.pyx", line 343, in cupy.cuda.memory.MemoryPointer.__init__
TypeError: __init__() takes exactly 2 positional arguments (1 given)```

### PR DESCRIPTION
**Pull request by shadeMe**
This PR fixes the following and adds a regression test:

- The original invocation of `cupy.cuda.MemoryPointer` was missing the mandatory `offset` argument. This should be `0`.
- `cupy.cuda.MemoryPointer` expects a wrapper to a `cupy`-owned CUDA memory and not a raw pointer to it. So, CUDA memory allocated by external libraries needs to be wrapped with `UnownedMemory` and then passed to the array c'tor [[1](https://github.com/cupy/cupy/issues/4644#issuecomment-775480279)].
- We weren't testing that the incoming array-like obj was owned by a CUDA device.
- We were incorrectly passing the array stride as its shape.

Should fix #564.

### DIFF
--- a/thinc/backends/cupy_ops.py
+++ b/thinc/backends/cupy_ops.py
@@ -75,10 +75,23 @@ class CupyOps(Ops):
         dtype = {"dtype": dtype} if dtype is not None else {}
         if isinstance(data, cupy.ndarray):
             return self.xp.asarray(data, **dtype)
-        elif hasattr(data, "data_ptr"):
+        elif (
+            hasattr(data, "data_ptr")
+            and hasattr(data, "storage")
+            and hasattr(data, "device")
+        ):
             # Handles PyTorch Tensors
-            pointer = cupy.cuda.MemoryPointer(data.data_ptr())
-            shape = data.stride()
+            assert data.device.type == "cuda", "Non-CUDA tensor passed to CupyOps"
+
+            ext_pointer_wrapper = cupy.cuda.UnownedMemory(
+                data.data_ptr(),
+                data.storage().size(),
+                data.storage(),
+                data.device.index,
+            )
+
+            pointer = cupy.cuda.MemoryPointer(ext_pointer_wrapper, 0)
+            shape = data.shape
             array = self.xp.ndarray(shape, memptr=pointer, **dtype)
             return array
         else:
--- a/thinc/backends/cupy_ops.py
+++ b/thinc/backends/cupy_ops.py
@@ -19,6 +19,14 @@ from .ops import Ops
 from .numpy_ops import NumpyOps
 from . import _custom_kernels
 from ..types import DeviceTypes
+from ..util import (
+    torch2xp,
+    tensorflow2xp,
+    is_mxnet_array,
+    mxnet2xp,
+    is_torch_array,
+    is_tensorflow_array,
+)
 
 
 @registry.ops("CupyOps")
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by shadeMe** [comment]

**Comment by shadeMe** [comment]
One more time with feeling: @explosion-bot please test_gpu

**Comment by shadeMe** [comment]

File: thinc/backends/cupy_ops.py [hunk 1]
```diff
@@ -75,11 +75,23 @@ def asarray(self, data, dtype=None):
...
+        elif (
+            hasattr(data, "data_ptr")
+            and hasattr(data, "storage")
+            and hasattr(data, "device")
+        ):
             # Handles PyTorch Tensors
-            pointer = cupy.cuda.MemoryPointer(data.data_ptr())
-            shape = data.stride()
-            array = self.xp.ndarray(shape, memptr=pointer, **dtype)
+            assert data.device.type == "cuda", "Non-CUDA tensor passed to CupyOps"
```
danieldk, line 84: Since this method is part of the public API, it's probably better to throw a `ValueError` here?

thinc/backends/cupy_ops.py
```diff
@@ -75,11 +75,23 @@ def asarray(self, data, dtype=None):
...
+
+            ext_pointer_wrapper = cupy.cuda.UnownedMemory(
+                ptr=data.data_ptr(),
+                size=data.storage().size(),
+                owner=data,
+                device_id=data.device.index,
+            )
+
+            pointer = cupy.cuda.MemoryPointer(ext_pointer_wrapper, offset=0)
+            array = self.xp.ndarray(shape=data.shape, memptr=pointer, **dtype)
```
danieldk, line 94: The semantics of `dtype` are incorrect. If `dtype` is not specified, the data type of the original array must be used, otherwise:

```
>>> x = torch.tensor([42.0], device='cuda')
>>> y = ops.asarray(x)
>>> y
array([5.48372203e-315])
```

If `dtype` is specified and differs from the `dtype` of the original array, a copy should be made (converting to the specified `dtype`).

shadeMe:
Added support for zero-alloc conversions for Torch, TF and MXNet arrays and removed the error for CPU-allocated inputs - I think we'll want to just copy it to the GPU in such cases instead of raising a hard error.

shadeMe:

thinc/backends/cupy_ops.py
```diff
@@ -19,6 +19,14 @@
...
 from . import _custom_kernels
 from ..types import DeviceTypes
+from ..util import (
+    torch2xp,
+    tensorflow2xp,
+    is_mxnet_array,
+    mxnet2xp,
+    is_torch_array,
+    is_tensorflow_array,
+)
```
adrianeboyd, line 29: We usually reformat (manually) with multiple imports rather than using this kind of multi-line import:

```suggestion
from ..util import torch2xp, tensorflow2xp, is_mxnet_array, mxnet2xp
from ..util import is_torch_array, is_tensorflow_array
```

svlandeg (commented):
Looks good to me, but will defer the final review & merging to @danieldk if he's happy :-)

danieldk (commented):
Looks good to me, though I'll run my 'test all the PyTorch versions' script just to be sure.

PR #640, opened 2022-04-20
Bloom Embeddings Notebook

koaning:
Added a notebook to accompany https://github.com/explosion/explosion.ai/pull/48.


examples/bloom-embeddings.ipynb
```diff
@@ -0,0 +1,729 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## The Bloom embeddings algorithm\n",
+    "\n",
+    "In a normal embedding table, each word-string is mapped to a distinct ID.\n",
+    "Usually these IDs will be sequential, so if you have a vocabulary of 100 words,\n",
+    "your words will be mapped to numbers `range(100)`. The sequential IDs can then\n",
+    "adrianeboyd marked this conversation as resolved.\n",
```
kadarakos, line 39: this is a bug line :-)

examples/bloom-embeddings.ipynb
```diff
@@ -0,0 +1,729 @@
...
+    "be used as indices into an embedding table: if you have 100 words in your\n",
+    "adrianeboyd marked this conversation as resolved.\n",
```
kadarakos, line 41: bugline

examples/bloom-embeddings.ipynb
```diff
@@ -0,0 +1,729 @@
...
+    "    else:\n",
+    "        return number_vector + (word_id % number_oov)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This gives the model a little more resolution for the unknown words. If all\n",
+    "out-of-vocabulary words are assigned the same vector, then they'll all look\n",
```
kadarakos, line 87: In a sense I agree, but in another sense since the encoding is most of the time contextual, models learn to properly contextualize `UNK`. This is why for example a pointer mechanism is good at a copying unseen words.

examples/bloom-embeddings.ipynb
```diff
@@ -0,0 +1,729 @@
...
+    "    row1 = hash1 % table.shape[0]\n",
+    "    row2 = hash2 % table.shape[0]\n",
+    "    return table[row1] + table[row2]\n",
+    "\n",
+    "def update_vector(table, word, d_vector):\n",
+    "    hash1 = mmh3.hash(word, seed=0)\n",
+    "    hash2 = mmh3.hash(word, seed=1)\n",
+    "    row1 = hash1 % table.shape[0]\n",
+    "    row2 = hash2 % table.shape[0]\n",
+    "    table[row1] -= 0.001 * d_vector\n",
```
kadarakos, line 147: Is this `0.001` the learning-rate? Would it make sense to have a `learn_rate` parameter here?

examples/bloom-embeddings.ipynb
```diff
@@ -0,0 +1,729 @@
...
+    "    random.shuffle(examples)\n",
+    "    loss=0.\n",
+    "    for word, truth in examples:\n",
+    "        key1 = mmh3.hash(word, 0) % nr_hash_vector\n",
+    "        key2 = mmh3.hash(word, 1) % nr_hash_vector\n",
+    "        hash_vector = hash_vectors[key1] + hash_vectors[key2]\n",
+    "        diff = hash_vector - truth\n",
+    "        hash_vectors[key1] -= learn_rate * diff\n",
+    "        hash_vectors[key2] -= learn_rate * diff\n",
+    "        loss += (diff**2).sum()\n",
```
kadarakos, line 504: I'm always very confused with derivatives, but here the loss is the `(x - y)^2` and the derivative is `x - y`. It doesn't really matter I guess but the if the derivative is `x - y` then the loss is `1/2 * (x - y)^2` right?

koaning:
Many of the comments here seem valid, but these are excerpts from the blogpost written [here](https://github.com/explosion/explosion.ai/pull/48). I wouldn't mind addressing the items that @kadarakos mentions here, but any changes should also be reflected in the blog post, which is why I'm hesitant to make changes to the code examples. 

@adrianeboyd, looping you in because you're also working on the blogpost.

examples/bloom-embeddings.ipynb
adrianeboyd: I don't think this really matters, and as far as I can tell it's what's currently in e.g. `CategoricalCrossentropy`.

examples/bloom-embeddings.ipynb
adrianeboyd: I don't think this is needed for this example.

examples/bloom-embeddings.ipynb
adrianeboyd: I think it's okay that this explanation is only focused on the embed part.

examples/bloom-embeddings.ipynb
kadarakos: but this is the sum of squared error loss not cross-entropy. The gradient is the same.

kadarakos: Also I believe that the loss shown in the categorical cross entropy is wrong and the gradient is correct.

adrianeboyd: Ah, I see what you're saying. Since it doesn't make a practical difference for this demo, I think I would just leave it the way Matt wrote it for now, but we'll know what they're referring to if anyone brings it up.

svlandeg (approved):
Great idea to have an accompanying notebook! I'll leave the details for Adriane to review & coordinate with the blog post :-)

adrianeboyd:
Let me go ahead and merge this so we can test it in the blog post draft. We can always update details in the future...

PR #643, opened 2022-04-22
NumpyOps: Add a method to get a table of C BLAS functions

danieldk:
This table can be used for downstream `cdef nogil` functions that need to use a BLAS function from the BLAS implementation used by an Ops subclass.

In draft because this change requires https://github.com/explosion/cython-blis/pull/70.


shadeMe (approved):

svlandeg (commented):
We're retiring `develop` - can you target `master` instead? 🙏

danieldk:
> We're retiring `develop` - can you target `master` instead? 🙏

Updated to `master`. Now that blis 0.9.0 is released, I'll also put hits out of draft.

thinc/backends/numpy_ops.pyx
```diff
@@ -82,6 +83,9 @@ class NumpyOps(Ops):
         else:
             return self.xp.empty(shape, dtype=dtype)
 
+    def cblas(self) -> CBlas:
+        return CBlas()
```
svlandeg, line 87: Do we want to create a new instance here every time this is called?

danieldk: I changed this to use a singleton. Unfortunately, it needs to be a global variable, otherwise `NumpyOps` cannot be pickled, since `CBlas` contains an unserializable C pointer.

website/docs/api-backends.md
```diff
@@ -382,6 +382,27 @@ the inputs and outputs.
...
+</inline-list>
+
+Get a table of C BLAS functions usable in Cython `cdef nogil` functions. This
+method does not take any arguments.
+
+<infobox variant="warning">
+
+This method is only supported by `NumpyOps` and subclasses of `NumpyOps`. A
+`NotImplementedError` exception is raised when calling this method on `Ops` or
+`CupyOps`.
```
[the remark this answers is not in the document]
danieldk: Fixed.

PR #646, opened 2022-04-25
Sanity checks for GPU tests, 3rd party framework tensor conversion changes

shadeMe:
* Add sanity checks to handle cases of hidden GPU devices
* Allow `...2xp` utility methods to accept a target `Ops` object for conversions


svlandeg:
> First of two stacked PRs (to be merged with a merge commit; squashing/rebasing will break the 2nd, in-progress PR).

I'm not sure I understand this? If we squash this one, can't you rebase the next PR or merge in `develop`? (not sure what you mean by "will break")

shadeMe:
Actually, please ignore that last bit - Looks like I misunderstood how stacked PRs work. The original intention was to work on a PR that was dependent on these changes whilst waiting for it to be merged. I was hoping to stack the 2nd PR on this in such a way that I wouldn't have to do any additional merges (i.e., from `explosion:develop` after this PR is merged), but it seems like it doesn't work when attempting to merge branches _between_ forks.

So, this PR can be squashed/rebased into `explosion:develop`. I'll just merge that into my local repo and rebase by PR as you suggested.

thinc/util.py
```diff
@@ -20,8 +20,16 @@
     import cupy
 
     has_cupy = True
+    cupy_version = Version(cupy.__version__)
+
+    if cupy_version.major >= 10:
+        # fromDlpack was deprecated in v10.0.0.
+        cupy_from_dlpack = cupy.from_dlpack
+    else:
+        cupy_from_dlpack = cupy.fromDlpack
```
svlandeg, line 29: Alternatively, we could have a module like spaCy's `spacy/compat.py` to deal with this deprecation, I think it's nice to have it in one place.

svlandeg (approved):
All of these changes feel pretty reasonable to me! 

Maybe the only thing I was wondering about is whether the `ops` argument should be keyword-only. Then again I don't think we'll likely be expanding these methods with many more arguments in the future?

From a practical point of view, I'd love it if Daniël could give this a quick review as well. Considering he's out at the moment, we can either just merge this already (it's `develop` anyway) or wait for him to get back - do you have a preference, especially considering the follow-up PR?

thinc/util.py
shadeMe: Good idea! I'll move it to a separate module.

shadeMe:
Re. `ops` - Yeah, it can be keyword-only, and I too doubt that we'll be adding many more arguments - if at all any - in the future. Will change that.

Re. merging - I'm fine with leaving this open until Daniël gets a chance to take a look at it. The other PR doesn't include any urgent fixes, and we still need to fix the Buildkite GPU build environment to actually test the changes in CI. So, it can wait.

thinc/util.py
shadeMe: On second thought, I'll leave that for a fresh PR as it'll touch a bunch of other modules, etc.

thinc/util.py
```diff
@@ -124,6 +137,15 @@ def is_torch_array(obj: Any) -> bool:  # pragma: no cover
         return False
 
 
+def is_torch_gpu_array(obj: Any) -> bool:  # pragma: no cover
+    if not is_torch_array(obj):
+        return False
+    elif obj.device.type == "cuda":
```
danieldk, line 143: I think

```suggestion
    elif obj.is_cuda:
```

should work here.

thinc/util.py
```diff
@@ -124,6 +137,15 @@ def is_torch_array(obj: Any) -> bool:  # pragma: no cover
...
+        return True
+    else:
+        return False
```
danieldk, line 146: PyTorch tensors should have `is_cuda` and we can short-circuit here:

```suggestion
    return is_torch_array(obj) and obj.is_cuda
```

thinc/util.py
```diff
@@ -133,6 +155,15 @@ def is_tensorflow_array(obj: Any) -> bool:  # pragma: no cover
         return False
 
 
+def is_tensorflow_gpu_array(obj: Any) -> bool:  # pragma: no cover
+    if not is_tensorflow_array(obj):
+        return False
+    elif "GPU:" in obj.device:
+        return True
+    else:
+        return False
```
danieldk, line 164:
```suggestion
    return is_tensorflow_array(obj) and "GPU:" in obj.device:
```

thinc/util.py
```diff
@@ -142,6 +173,15 @@ def is_mxnet_array(obj: Any) -> bool:  # pragma: no cover
         return False
 
 
+def is_mxnet_gpu_array(obj: Any) -> bool:  # pragma: no cover
+    if not is_mxnet_array(obj):
+        return False
+    elif obj.context.device_type != "cpu":
+        return True
+    else:
+        return False
```
danieldk, line 182:
```suggestion
    return is_mxnet_array(obj) and obj.context.device_type != "cpu"
```

thinc/util.py
```diff
@@ -351,12 +392,25 @@ def xp2torch(
     return torch_tensor
 
 
-def torch2xp(torch_tensor: "torch.Tensor") -> ArrayXd:  # pragma: no cover
-    """Convert a torch tensor to a numpy or cupy tensor."""
-    if torch_tensor.is_cuda:
-        return cupy.fromDlpack(torch.utils.dlpack.to_dlpack(torch_tensor))
+def torch2xp(
+    torch_tensor: "torch.Tensor", *, ops: Any = None
```
danieldk, line 396:
```suggestion
    torch_tensor: "torch.Tensor", *, ops: Ops = None
```

or if that leads to a recursive import (probably does), then:

```suggestion
    torch_tensor: "torch.Tensor", *, ops: "Ops" = None
```

(I think)

thinc/util.py
```diff
@@ -382,24 +436,33 @@ def xp2tensorflow(
     return tf_tensor
 
 
-def tensorflow2xp(tf_tensor: "tf.Tensor") -> ArrayXd:  # pragma: no cover
-    """Convert a Tensorflow tensor to numpy or cupy tensor."""
+def tensorflow2xp(
+    tf_tensor: "tf.Tensor", *, ops: Any = None
```
danieldk, line 440: The `ops` type should probably be added here as well.

thinc/backends/_custom_kernels.py
```diff
@@ -5,7 +5,12 @@
 
 try:
     import cupy
-except ImportError:
+
+    try:
+        cupy.cuda.runtime.getDeviceCount()
+    except cupy.cuda.runtime.CUDARuntimeError:
+        cupy = None
```
danieldk, line 12: I am not sure I understand this check. If there are no actual devices, wouldn't/shouldn't we fail in `CupyOps` already?

shadeMe: The `CupyOps` module imports dependencies from `_custom_kernels`, and none of the `cupy`-related imports in the former raise an exception when there are no actual devices. The first-chance exception gets raised only when an actual `cupy` function gets called, i.e., when the `KERNELS.get_function(name)` function from the latter module gets called.

To be fair, perhaps the above method isn't the best way to catch these errors after all. We have a bunch of modules that perform their own flavour of compatibility checks, leading to multiple sources of truth. I think it would be better to do all the compatibility-related checks in a single module (`compat.py` as Sofie suggested) and then use that as the source of truth in all other modules. If we wish to retain backward-compatibility, we can also re-export the necessary variables (`has_cupy`, etc) from those modules where we're currently doing this. Although in my opinion, we probably want to avoid re-exports as it unnecessarily diffuses/increases the public API's contract/surface area.

Does that sound like a reasonable approach?

thinc/util.py
shadeMe: Done.

shadeMe:
Hmm, looks like the `mypy` checks fail when `ops: "Ops"` is used as the type annotation. Using a fully/partially qualified identifier (`thinc.api,Ops` or `.api.Ops`) doesn't fix it either. Any suggestions?

thinc/backends/_custom_kernels.py
danieldk: Sounds good!

danieldk (approved):

danieldk:
Merging, since @svlandeg also approved, and this is to `develop`, so we can still fix any issues that might pop up.

PR #647, opened 2022-04-28
Cross entropy fix

kadarakos:
This PR makes the `CategoricalCrossentropy` loss more strict only allowing `guesses` and `truths` that represent exclusive classes and it fixes the computation of the loss value. It also changes all the tests that involve the `CategoricalCrossentropy` and `SequenceCategoricalCrossentropy`. 

It also separates `CategoricalCrossentropy` from `SparseCategoricalCrossentropy` and `CategoricalCrossentropy` takes only `Floats2d` as `truths` whereas `SparseCategoricalCrossentropy` takes `Union[Ints1d, Sequence[int], Sequence[str]]`. 

 It also adds a way of handling `legacy` modules in `thinc`.


thinc/loss.py
```diff
@@ -98,21 +97,38 @@ def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]
...
+            # Convert 1d truths to 2d and apply smoothing.
+            if truths.ndim == 1:
+                truths = to_categorical(
+                    cast(Ints1d, truths),
+                    n_classes=guesses.shape[-1],
+                    label_smoothing=self.label_smoothing,
+                )
+            # Validate 2d truths and apply smoothing if its one-hot.
+            elif truths.ndim == 2:
+                if not xp.all(truths.sum(axis=1) == 1):
```
danieldk, line 112: Doesn't this also pass for `truths` where the sum of the probabilities is one?

danieldk: Never mind, I was already thinking about the smoothing :).

thinc/loss.py
```diff
@@ -98,21 +97,38 @@ def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]
...
+                    raise ValueError(
+                        "Cannot calcuate CategoricalCrossentropy. "
+                        "All rows of 'truths' have to be a "
+                        "valid categorical distribution (sum to 1)."
+                    )
+                if self.label_smoothing:
+                    # Check if one-hot
+                    if xp.all(truths.sum(axis=0) == 1):
```
danieldk, line 120: Isn't this false when e.g. the first label is not used in the current batch?

thinc/loss.py
```diff
@@ -98,21 +97,38 @@ def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]
...
+                        truths = smooth_one_hot(truths, self.label_smoothing)
+                    else:
+                        raise ValueError(
+                            "Can only apply label-smoothing to one-hot target."
+                        )
+            # Something went wrong.
+            else:
```
danieldk, line 127: Maybe with these changes, we can also specify the type in the signature for `truths`?

thinc/loss.py
```diff
@@ -141,19 +166,26 @@ def get_grad(self, guesses: Floats2d, truths: IntsOrFloatsOrStrs) -> Floats2d:
...
-        return self._get_loss_from_grad(d_truth)
-
-    def _get_loss_from_grad(self, d_truth: Floats2d) -> float:
-        # TODO: Add overload for axis=None case to sum
-        return (d_truth**2).sum()  # type: ignore
+        xp = get_array_module(guesses)
+        target, mask = self.convert_truths(truths, guesses)
+        self._validate_input(guesses, target)
+        target *= mask
+        logprobs = xp.log(guesses + 1e-9)
```
danieldk, line 184: 🎉

thinc/loss.py
```diff
@@ -126,12 +142,21 @@ def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]
     def __call__(
         self, guesses: Floats2d, truths: IntsOrFloatsOrStrs
     ) -> Tuple[Floats2d, float]:
+        # XXX not ideal to run convert_input and _validate_input twice.
```
danieldk, line 145: Maybe it's useful to check the cost. If it's too much, I guess an alternative could be to have `_get_grad` and `_get_loss` private methods that don't do the validation and conversion, with would then be called by the entry points `__call__`, `get_grad` and `get_loss`, so we'd only have to do validation once for `__call__`.

thinc/loss.py
kadarakos: Yeah you are right this is wrong!

thinc/loss.py
kadarakos: okok good point!

kadarakos: Ah right, I tried it before, its just really hard to get `mypy` to be happy, but I'll try!

svlandeg (commented):
We're retiring `develop` - can you target `master` instead and resolve the conflicts carefully? 🙏

thinc/loss.py
```diff
@@ -96,26 +99,44 @@ def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]
...
+                    n_classes=guesses.shape[-1],
+                    label_smoothing=self.label_smoothing,
+                ),
             )
-        else:
-            if self.label_smoothing:
+        # Check if 2D array is valid and maybe apply smoothing.
+        elif truths.ndim == 2:
+            # Validate if each row sums to 1.
+            if not xp.all(truths.sum(axis=1) == 1):
```
danieldk, line 123: I am a bit worried about this part. What if it doesn't sum up to exactly one due to rounding errors that where incurred while computing the truths table?

thinc/loss.py
```diff
@@ -96,26 +99,44 @@ def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]
...
                 raise ValueError(
-                    "Label smoothing is only applied, when truths have type "
-                    "List[str], List[int] or Ints1d, but it seems like Floats2d "
-                    "was provided."
+                    "Cannot calculate CategoricalCrossentropy. "
+                    "All rows of 'truths' have to be a "
+                    "valid categorical distribution (sum to 1)."
                 )
+            if self.label_smoothing:
```
danieldk, line 129: Do we want to support this case? I think it is reasonable expect that if people want one-hot semantics that they pass in the gold label representation. `convert_truths` is already a large enumeration of cases. I worry that we end up supporting the cartesian product of all representations/options.

thinc/loss.py
```diff
@@ -126,34 +147,63 @@ def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]
...
-    def get_grad(self, guesses: Floats2d, truths: IntsOrFloatsOrStrs) -> Floats2d:
         target, mask = self.convert_truths(truths, guesses)
+        self._validate_input(guesses, target)
+        d_truth = self._get_grad(guesses, target, mask)
+        loss = self._get_loss(guesses, target, mask)
+        return (d_truth, loss)
+
+    def _validate_input(self, guesses: Floats2d, target: Floats2d) -> None:
         xp = get_array_module(target)
+        if not xp.allclose(guesses.sum(axis=1), 1.0):
```
danieldk, line 158: I think it's good to do input validation, but do you have an idea of the impact in CPU and GPU training?

thinc/loss.py
```diff
@@ -225,7 +275,7 @@ def __call__(
         self, guesses: Sequence[Floats2d], truths: Sequence[IntsOrFloatsOrStrs]
     ) -> Tuple[List[Floats2d], float]:
         grads = self.get_grad(guesses, truths)
-        loss = self._get_loss_from_grad(grads)
+        loss = self.get_loss(guesses, truths)
```
danieldk, line 278: Do we want `_get_loss` here? I think the input validation is already done through `get_grad`?

danieldk: Oh, nevermind.

thinc/loss.py
kadarakos: this is a mistake indeed it should just check if its close.

thinc/loss.py
kadarakos: Not sure actually. Since we do allow users to give different kinds of representations of "one correct label per sample" I think it would be more counter intuitive then not do smoothing when its not in one of the 3 options (List[str], List[int], Floats2d(onehot)). 

With the new version of the loss we actually support less kinds of things than before: we only accept rows that correspond to categorical distributions. I think we enumerated already all the options in that realm with the current code.

thinc/loss.py
kadarakos: Not sure honestly yet, but I think this is really important, because if the rows of `truths` and `guesses` don't represent categorical distributions the loss value doesn't make sense.

thinc/loss.py
```diff
@@ -58,30 +58,33 @@ def __init__(
         else:
             self._name_to_i = {}
 
-    def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]:
+    def convert_truths(
+        self, truths: IntsOrFloatsOrStrs, guesses: Floats2d
```
shadeMe, line 62: This doesn't necessarily belong to this PR, but I was at first confused by the `Floats` in `IntsOrFloatsOrStrs` given that we weren't explicitly handling floats in the method below. 

The definition of that type as it stands currently: `IntsOrFloatsOrStrs = Union[Ints1d, Floats2d, Sequence[int], Sequence[str]]`

IMO, the type alias doesn't sufficiently describe the underlying type annotation. I think it would be less confusing to use the latter or rename the former.

thinc/loss.py
```diff
@@ -58,30 +58,33 @@ def __init__(
...
                     if self.names is None:
-                        msg = (
+                        raise ValueError(
                             "Cannot calculate loss from list of strings without names. "
                             "You can pass the names as a keyword argument when you "
                             "create the loss object, "
                             "e.g. CategoricalCrossentropy(names=['dog', 'cat'])"
                         )
-                        raise ValueError(msg)
+                    # Convert List[str] to List[int]
```
svlandeg, line 85: lets add an 
```
assert(truths[0], str)
```
here for clarity & robustness?

thinc/loss.py
```diff
@@ -96,26 +99,44 @@ def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]
                             negatives_mask[i] = 0  # type: ignore
                             negatives_mask[i][neg_index] = -1  # type: ignore
                     truths = [self._name_to_i[name] for name in truths]
+            # cast truths as 1d array
             truths = xp.asarray(truths, dtype="i")
+            # get mask for 1d array
             mask = _make_mask(guesses, missing)
         else:
+            # get mask for 2d array
```
svlandeg, line 107: So `truths` is of type ``IntsOrFloatsOrStrs`, so what happens when it's `Ints1d`?

thinc/util.py
```diff
@@ -19,7 +19,8 @@
 
 DATA_VALIDATION: ContextVar[bool] = ContextVar("DATA_VALIDATION", default=False)
 
-from .types import ArrayXd, ArgsKwargs, Ragged, Padded, FloatsXd, IntsXd  # noqa: E402
+
+from .types import ArrayXd, ArgsKwargs, Ragged, Padded, FloatsXd, IntsXd, Floats2d # noqa: E402
```
svlandeg, line 23: I don't know what happened here historically, but let's move the `DATA_VALIDATION` definition to below all the imports, and move the very last `from typing` import to the second line.

(this is outside your PR but just happened to notice this, so we might as well clean it up)

thinc/loss.py
```diff
@@ -246,12 +296,12 @@ def get_grad(
     def get_loss(
         self, guesses: Sequence[Floats2d], truths: Sequence[IntsOrFloatsOrStrs]
     ) -> float:
-        return self._get_loss_from_grad(self.get_grad(guesses, truths))
-
-    def _get_loss_from_grad(self, grads: Sequence[Floats2d]) -> float:
+        err = "Cannot calculate SequenceCategoricalCrossentropy loss: guesses and truths must be same length"
+        if len(guesses) != len(truths):  # pragma: no cover
+            raise ValueError(err)
```
shadeMe, line 301: The error string can be passed to `ValueError` directly, I think. Same in the `get_grad` method above.

thinc/loss.py
kadarakos: good point, its super complicated with the handling of the different types of input for me too!

thinc/loss.py
kadarakos: I completely agree and I agree with @svlandeg as well that the `if-else` forest is a mess, but this PR is trying to make the `CategoricalCrossentropy` more correct than before. And I do feel like there will be more refactor followups.

thinc/loss.py
kadarakos: If its `Ints1d` then it goes into `_make_mask_by_value` which handles both 1d and 2d and my comment #get mask for 2d array is lying.

danieldk: > (all these if-else branches are so confusing!)

Maybe it makes sense to break this up into multiple functions. Then `convert_truths` could become more readable. E.g.:

```
if isinstance(truths, list):
  truths = truths.truths_list_to_array(list) # Now truths is Union[Ints1d, Floats2d]

if truths.ndim == 1:
  self.truths2onehot(truths, n_classes=guesses.shape[-1]) # Not truths is Floats2d
else:
  # Apply label smoothing, etc.
```

(This is more schematic than actual code. ;) )

thinc/loss.py
danieldk: Agreed. I think that applies to my comment breaking this up into functions as well.

thinc/loss.py
danieldk: Sorry, I should add that this is not necessarily for this PR. But I think we should break this up in a follow-up PR.

svlandeg: > Maybe it makes sense to break this up into multiple functions. 

Yes - I was thinking the exact same thing. And yes, follow-up PR is fine :-)

kadarakos: I do feel that there is this cross-entropy refactor lurking in my life. I'll add a ticket.

kadarakos: Do we agree that its a next PR?

svlandeg: >and my comment #get mask for 2d array is lying

This was my suspicion 😛

svlandeg: I'll resolve here for now!

thinc/loss.py
kadarakos: What do you think?

thinc/loss.py
kadarakos: What do you think?

thinc/loss.py
```diff
@@ -58,30 +58,34 @@ def __init__(
...
-                        msg = (
+                        raise ValueError(
                             "Cannot calculate loss from list of strings without names. "
                             "You can pass the names as a keyword argument when you "
                             "create the loss object, "
                             "e.g. CategoricalCrossentropy(names=['dog', 'cat'])"
                         )
-                        raise ValueError(msg)
+                    # Make sure "truths" is List[str]
+                    assert isinstance(truths[0], str)
```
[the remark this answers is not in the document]
kadarakos: This is not part of the user input validation. In `convert_truths` there is not much input validation because now the typing is "fixed" in the sense that the `IntsOrFloatsOrStrs` is the type of `truths` so if the user provides the right type then this works. The `assert` is there for clarity that if it wasn't `List[int]` then at this point of the code it must be a `List[str]`.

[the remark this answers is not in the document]
danieldk: But you'll only capture it if you are using MyPy? I think many downstream users don't use type checking, so their code will happily run until they get an assertion and it's not really clear from that what they did wrong.

svlandeg (commented):
So the way we can avoid this to be breaking, is to ensure the legacy versions keep doing the same thing. The tests could query the registry for a specific version and test the expected output of that version, even if it means it's testing for the MSE value :trollface: 

We'd then basically introduce `CategoricalCrossentropy.v4` which will use the new code and will have the actual correct loss value, and a new set of tests that test against that specifically.

This will be a bit of a hassle because it'll require copying some things over into something like `thinc.legacy`, but considering the versioning we have in place, it feels like the best option?

svlandeg:
> This PR makes the `CategoricalCrossentropy` loss more strict only allowing `guesses` and `truths` that represent exclusive classes 

Why can't / shouldn't it (also) support multi-label classification?

kadarakos:
> > This PR makes the `CategoricalCrossentropy` loss more strict only allowing `guesses` and `truths` that represent exclusive classes
> 
> Why can't / shouldn't it (also) support multi-label classification?

I was thinking that I would have a separate loss for that for two main reasons.

The simple one was just that the loss computation itself is different. So if the rows add up to 1 then its the way its done now, otherwise it would have to compute the binary cross-entropy. The gradient is the same though that's true.

The other main reason was just the confusing usage pattern from my perspective. So it takes `Ints1d`, `List[str]`, `List[int]` or `Floats2d` as `truths` and it also does label-smoothing. The first three types of input and label-smoothing doesn't make sense for the multi-label binary cross entropy case. The only way to make it do binary cross entropy is to give it `Floats2d` and label-smoothing = False and in all other configurations it does categorical. 

I just thought its nicer to implement a binary cross-entropy separately that can take different kinds of `truths` that's more appropriate for multiple labels per sample and has its own docs plus it always computes that same loss.

svlandeg:
Thanks for the additional explanation, and agreed!

danieldk:
> The simple one was just that the loss computation itself is different. So if the rows add up to 1 then its the way its done now, otherwise it would have to compute the binary cross-entropy. The gradient is the same though that's true.

But is `CategoricalCrossEntropy` currently used anywhere for multi-class classification? If so, this change would break those cases? I am not sure, I haven't really looked much yet at multi-class classification in Thinc/spaCy.

svlandeg:
I don't think it's being used in spaCy right now, but it should. And when we do, I'm proposing we work via specific versions, e.g. 
```
loss_function = registry.get("losses", "CategoricalCrossentropy.v3")
```
This will make it easier to keep the same behaviour if people want it, but to allow others to upgrade if they chose so.

It might still be that other users of Thinc are using the current `CategoricalCrossentropy` and would get bitten by a change in numerical values when they're logging results across training runs. Then we'd be able to present the above solution to them for making sure they have access to the same functionality.

The fact that we need to change the tests here, is a huge red flag that this can be very breaking - which is why I'm proposing we lean into the framework of versioning that we've already set up anyway. Then we can keep the tests largely the same for the old functionality but write new ones for the new functionality.

kadarakos:
> I don't think it's being used in spaCy right now, but it should. And when we do, I'm proposing we work via specific versions, e.g.
> 
> ```
> loss_function = registry.get("losses", "CategoricalCrossentropy.v3")
> ```
> 
> Instead of a direct import. This will make it easier to keep the same behaviour if people want it, but to allow others to upgrade if they chose so.
> 
> It might still be that other users of Thinc are using the current `CategoricalCrossentropy` and would get bitten by a change in numerical values when they're logging results across training runs, or get bitten because the functionality is now more strict as Daniël mentions. Then we'd be able to present the above solution to them for making sure they have access to the same functionality.
> 
> The fact that we need to change the tests here, is a huge red flag that this can be very breaking - which is why I'm proposing we lean into the framework of versioning that we've already set up anyway. Then we can keep the tests largely the same for the old functionality but write new ones for the new functionality.

Added a suggestion for handling legacy stuff in general, but I didn't exactly follow your suggestion, but let me know if you'd like that better.

The idea is that there is the `thinc.legacy` as you've said. So `thinc.legacy` at the moment only has `loss.py` meaning that the `thinc.loss` should import from `thinc.legacy.loss` if it would like to run older versions. This is what it does atm. 

For tests I've made a separate module `thinc.tests.legacy`, which again replicates the corresponding tests and imports from `thinc.legacy`. In this case its just a copy of the old `test_loss.py` removing the tests for, which legacy versions do not exist.

thinc/legacy/loss.py
```diff
@@ -0,0 +1,230 @@
+from ..util import to_categorical, get_array_module
+from typing import Optional, Sequence, Dict, Union, Tuple
+from typing import TypeVar, Generic, Any, cast, List
+from ..types import Floats2d, Ints1d
```
svlandeg, line 4:
```suggestion
from typing import Optional, Sequence, Dict, Union, Tuple
from typing import TypeVar, Generic, Any, cast, List
from ..types import Floats2d, Ints1d
from ..util import to_categorical, get_array_module
```
(just reordering)

thinc/legacy/loss.py
```diff
@@ -0,0 +1,230 @@
...
+
+LossT = TypeVar("LossT")
+GradT = TypeVar("GradT")
+GuessT = TypeVar("GuessT")
+TruthT = TypeVar("TruthT")
+IntsOrFloats = Union[Ints1d, Floats2d]
+IntsOrFloatsOrStrs = Union[Ints1d, Floats2d, Sequence[int], Sequence[str]]
+
+
+class Loss(Generic[GuessT, TruthT, GradT, LossT]):  # pragma: no cover
```
svlandeg, line 15: I don't think we have to repeat this class definition, do we? I would put in `legacy` only the parts that have changed in newer versions

thinc/legacy/loss.py
```diff
@@ -0,0 +1,230 @@
...
+        return self.get_grad(guesses, truths), self.get_loss(guesses, truths)
+
+    def get_grad(self, guesses: GuessT, truths: TruthT) -> GradT:
+        ...
+
+    def get_loss(self, guesses: GuessT, truths: TruthT) -> LossT:
+        ...
+
+
+class CategoricalCrossentropy(Loss):
```
svlandeg, line 35: You import this later on as `LegacyCrossEntropy` to distinguish the legacy class from the normal class in the type annotations. I hadn't really considered the typing when suggesting this legacy proposal. Now I'm thinking, could we make a class `LegacyCrossEntropy` here that inherits from the normal `CategoricalCrossentropy` but overrides the methods? The advantage would be that we wouldn't have to change the typing in the normal `loss.py` file - there we can just call the `LegacyCrossEntropy` constructor and say the returned type is a `CategoricalCrossentropy` so the public interface remains the same.

Also maybe we should call it `LegacyCategoricalCrossEntropy` and `LegacySequenceCrossEntropy` for clarity.

thinc/tests/legacy/test_loss.py
```diff
@@ -0,0 +1,252 @@
+import pytest
```
svlandeg, line 1: I would prefer to keep all the tests in the same file, that'll also help with tracking the changes in this PR. So basically where we're testing the old behaviour, we can rename the function to something with `legacy` in it and retrieve the correct class from the registry instead of importing directly. The rest should then remain the same, which will clearly showcase that the functionality can be non-breaking if you ajdust the script slightly to fetch the older version from registry.

svlandeg: And then we'd have entirely new testing methods for the new functionality.

svlandeg (commented):
This is going into the right direction IMO! Left some comments on general repo organization to look into that first.

thinc/legacy/loss.py
kadarakos: You mean to have the `LegacyCategoricalCrossEntropy` in the same `loss.py` as `CategoricalCrossEntropy` ?

thinc/legacy/loss.py
kadarakos: Oh, I've put stuff here only because `loss.py` imports from `legacy/loss.py` and then `legacy/loss.py` would import from `loss.py` and that would create a circular import.

thinc/legacy/loss.py
kadarakos: I guess this design was motivated by avoiding the circular imports. Like `legacy` would need to import from `loss.py` but then `loss.py` would need to import from `legacy/loss.py` to make this work I think. 

Here it would be `from thinc.loss import CategoricalCrossentropy` for the inheritance and then in `loss.py` it would be `from thinc.legacy.loss import LegacyCategoricalCrossentropy`.

thinc/legacy/loss.py
svlandeg: Right, but I think we can avoid the import from `legacy/loss.py` in `loss.py` and basically define the v1 etc directly in  `legacy/loss.py`?

thinc/legacy/loss.py
kadarakos: oh okay right!

thinc/legacy/loss.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+@registry.losses("CategoricalCrossentropy.v1")
+def configure_CategoricalCrossentropy_v1(
+    *,
+    normalize: bool = True,
+    names: Optional[Sequence[str]] = None,
+    missing_value: Optional[Union[str, int]] = None,
+) -> LegacyCategoricalCrossentropy:
+    return LegacyCategoricalCrossentropy(
```
svlandeg, line 193: So the hope is that now, we can have
``` 
 -> CategoricalCrossentropy:
```
as return type here (and for the other older versions)

kadarakos: ah right it does work i think!

thinc/loss.py
danieldk: Measure :).

danieldk:
We discussed this a while back, but I think this PR would be a good opportunity to replace the overloaded `CategoricalCrossEntropy` class by splitting the non-legacy classes into two variants: `SparseCategoricalCrossEntropy` and `CategoricalCrossEntropy` (and their sequence counterparts) like some other libraries (e.g. Torch) do. This simplifies the internals a lot, since you don't need a big decision tree for 1D vs. 2D input.

Also, some features can be specific to one variant. I still think label smoothing for 2D arrays has possible edge cases and users should just use sparse label representations for label smoothing [1]. If we have a separate class, we can avoid that users accidentally enable label smoothing by just not provideing that kwarg.

[1] I think label smoothing might be useful for 2D arrays, but we shouldn't just restrict it to one-hot arrays. I can imagine scenarios where you say the probability is 0.6 that it is this label and 0.4 that it is another label. Then label smoothing would be a method to increase the entropy of the distribution.

thinc/legacy/loss.py
```diff
@@ -0,0 +1,268 @@
...
+from typing import cast, List
+from ..types import Floats2d, Ints1d
+from ..config import registry
+from ..util import to_categorical, get_array_module
+from ..loss import CategoricalCrossentropy, IntsOrFloatsOrStrs
+from ..loss import _make_mask, _make_mask_by_value
+from ..loss import SequenceCategoricalCrossentropy
+
+
+class LegacyCategoricalCrossentropy(CategoricalCrossentropy):
```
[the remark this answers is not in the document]
kadarakos: Was wondering about that, but if the new version changes then that would mean that we have to remember to change the `Legacy` and I thought that's not really ideal.

kadarakos:
> We discussed this a while back, but I think this PR would be a good opportunity to replace the overloaded `CategoricalCrossEntropy` class by splitting the non-legacy classes into two variants: `SparseCategoricalCrossEntropy` and `CategoricalCrossEntropy` (and their sequence counterparts) like some other libraries (e.g. Torch) do. This simplifies the internals a lot, since you don't need a big decision tree for 1D vs. 2D input.
> 
> Also, some features can be specific to one variant. I still think one-hot label smoothing for 2D arrays has possible edge cases and users should just use sparse label representations for one-hot label smoothing [1]. If we have a separate class, we can avoid that users accidentally enable label smoothing by just not provideing that kwarg.
> 
> [1] I think label smoothing might be useful for 2D arrays, but we shouldn't just restrict it to one-hot arrays. I can imagine scenarios where the annotation say that with probability 0.6 that it is label A and 0.4 that it is label B. Then label smoothing would be a some method to increase the entropy of the distribution (where the one-hot case is just a specialization of the general method).

I used to be against the idea of having this added to the current PR, but since -- as you are saying -- we are already working out the legacy and the PR is huge to begin with, I think it does make sense to attempt the `Sparse`.

thinc/legacy/__init__.py
```diff
@@ -0,0 +1,8 @@
+from .loss import CategoricalCrossentropy
+from .loss import SequenceCategoricalCrossentropy
+
+
+__all__ = [
+    "CategoricalCrossentropy",
+    "SequenceCategoricalCrossentropy"
+]
```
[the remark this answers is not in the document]
kadarakos: sorry about the terrible names :(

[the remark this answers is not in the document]
svlandeg: Not your fault ;-)

thinc/legacy/loss.py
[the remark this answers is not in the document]
svlandeg: Ok, sure!

kadarakos:
@danieldk I was looking at the new version of the cross-entropy that handles the `Ragged` as input. I think it makes a lot of sense, but I'm wondering if its a good idea to support sequence types then like `Ragged`, `Padded`, `List[Floats2d]`, because I'm confused about why would then the truths always be `Floats2d`?

thinc/util.py
```diff
@@ -18,15 +19,14 @@
 from .compat import has_torch_mps_gpu
```
shadeMe, line 19:
```suggestion
```
Unused import.

thinc/loss.py
```diff
@@ -34,7 +36,118 @@ def get_loss(self, guesses: GuessT, truths: TruthT) -> LossT:
...
+    def _validate_input(self, guesses: FloatsOrRaggedT, target: Floats2d) -> None:
+        guesses_f2d = _to_array(guesses)
+        xp = get_array_module(target)
+        if not xp.allclose(guesses_f2d.sum(axis=1), 1.0):
+            raise ValueError(
+                "Cannot calculate CategoricalCrossentropy if "
+                "some rows of 'guesses' are not "
+                "valid categorical distributions (do not sum to 1)."
+            )
+        if guesses_f2d.shape != target.shape:  # pragma: no cover
```
shadeMe, line 51:
```suggestion
        elif guesses_f2d.shape != target.shape:  # pragma: no cover
```

thinc/loss.py
```diff
@@ -34,7 +36,118 @@ def get_loss(self, guesses: GuessT, truths: TruthT) -> LossT:
...
+            raise ValueError(
+                "Cannot calculate CategoricalCrossentropy loss: "
+                f"mismatched shapes: {guesses_f2d.shape} vs {target.shape}."
+            )
+        if xp.any(guesses_f2d > 1) or xp.any(guesses_f2d < 0):  # pragma: no cover
```
shadeMe, line 56:
```suggestion
        elif xp.any(guesses_f2d > 1) or xp.any(guesses_f2d < 0):  # pragma: no cover
```

thinc/loss.py
```diff
@@ -34,7 +36,118 @@ def get_loss(self, guesses: GuessT, truths: TruthT) -> LossT:
...
+            raise ValueError(
+                "Cannot calculate CategoricalCrossentropy loss "
+                "with guesses outside the [0,1] interval."
+            )
+        if xp.any(target > 1) or xp.any(target < 0):  # pragma: no cover
```
shadeMe, line 61:
```suggestion
        elif xp.any(target > 1) or xp.any(target < 0):  # pragma: no cover
```

thinc/loss.py
```diff
@@ -58,142 +171,190 @@ def __init__(
...
+    def __call__(
+        self, guesses: Floats2d, truths: Union[Sequence[int], Sequence[str]]
+    ) -> Tuple[Floats2d, float]:
+        target, mask = self.convert_truths(truths, guesses)
+        self._validate_input(guesses, target)
+        d_truth = self._get_grad(guesses, target, mask)
+        loss = self._get_loss(guesses, target, mask)
+        return (d_truth, loss)
+
+    def _check_ints1d(self, arr):
```
shadeMe, line 183: We should add a type hint for `arr`. According to the one call to this function in this file, it should be `Categories1d`?

thinc/loss.py
```diff
@@ -58,142 +171,190 @@ def __init__(
...
-                cast(Ints1d, truths),
-                n_classes=guesses.shape[-1],
+        if self.names is None:
+            raise ValueError(
+                "Cannot calculate loss from Sequence[str] without names. "
+                "You can pass the names as a keyword argument when you "
+                "create the loss object, "
+                "e.g. CategoricalCrossentropy(names=['dog', 'cat'])"
+            )
+        if missing_value is not None and not isinstance(missing_value, str):
```
shadeMe, line 244:
```suggestion
        elif missing_value is not None and not isinstance(missing_value, str):
```

thinc/loss.py
```diff
@@ -58,142 +171,190 @@ def __init__(
...
+            )
         missing = []
-        negatives_mask = None
-        if self.names:
-            negatives_mask = xp.ones((len(truths), len(self.names)), dtype="f")
+        for i, value in enumerate(truths):
+            if not isinstance(value, int):
+                raise ValueError(
+                    "The first value of the truths was of type "
+                    f"integer, but found {type(value)} during iteration."
```
shadeMe, line 216:
```suggestion
                    "The first value of `truths` was of type "
                    f"integer, but found {type(value)} during iteration."
```

thinc/loss.py
```diff
@@ -58,142 +171,190 @@ def __init__(
...
-            # transform categorical values to one-hot encoding
-            truths = to_categorical(
-                cast(Ints1d, truths),
-                n_classes=guesses.shape[-1],
+        if self.names is None:
+            raise ValueError(
+                "Cannot calculate loss from Sequence[str] without names. "
+                "You can pass the names as a keyword argument when you "
+                "create the loss object, "
+                "e.g. CategoricalCrossentropy(names=['dog', 'cat'])"
```
shadeMe, line 242: Seems a bit too verbose to me to include an example. Think we can trim it down to the following?

```suggestion
                "create the loss object"
```

thinc/loss.py
```diff
@@ -58,142 +171,190 @@ def __init__(
...
+            else:
                 raise ValueError(
-                    "Label smoothing is only applied, when truths have type "
-                    "List[str], List[int] or Ints1d, but it seems like Floats2d "
-                    "was provided."
+                    "When truths to SparseCE is provided "
+                    "in Sequence format, elements need to be "
+                    "of type str or int, but first element "
+                    f"was found to be {type(truths[0])}."
                 )
```
shadeMe, line 313: We should check all elements of `truths` like we do in `_convert_strs`.

thinc/util.py
```diff
@@ -254,6 +254,21 @@ def to_categorical(
     return label_distr[Y]
 
 
+def smooth_one_hot(X: Floats2d, label_smoothing: float):
```
shadeMe, line 257:
```suggestion
def smooth_one_hot(X: Floats2d, label_smoothing: float) -> Floats2d:
```

shadeMe (commented):
Just some minor stuff w.r.t type hints, etc.

thinc/legacy/loss.py
```diff
@@ -0,0 +1,266 @@
...
+    def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]:
+        xp = get_array_module(guesses)
+        missing = []
+        negatives_mask = None
+        if self.names:
+            negatives_mask = xp.ones((len(truths), len(self.names)), dtype="f")
+        missing_value = self.missing_value
+        # Convert list of ints or list of strings
+        if isinstance(truths, list):
+            truths = list(truths)
```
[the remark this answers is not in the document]
kadarakos: its the `legacy` so i just copied it verbatim.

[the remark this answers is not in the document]
kadarakos: should i add changes to the `legacy`?

thinc/loss.py
kadarakos: In this block we only check the first element and then `_convert_ints` and `_convert_strs` check all when they run and complain that the first one was of some type but now i ran into something else. So I think all elements get eventually checked here. No?

shadeMe: Ah, good point. Didn't notice that.

thinc/legacy/loss.py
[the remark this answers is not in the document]
kadarakos: But why are we changing the `legacy` version? I thought it should be just a snapshot of something we had before the new one.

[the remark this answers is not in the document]
shadeMe: I mean, I don't see any reason to make the above copy in either case. It's not a bug per-se, but if we expect the legacy version to be consumed by users, it makes sense to fix it there too. It's an implementation detail - we aren't changing any user-facing API contracts, so I'd say it's fine.

thinc/legacy/loss.py
```diff
@@ -0,0 +1,266 @@
...
+        self.names = names
+        self.missing_value = missing_value
+        self.neg_prefix = neg_prefix
+        self.label_smoothing = label_smoothing
+        if names is not None:
+            self._name_to_i = {name: i for i, name in enumerate(names)}
+        else:
+            self._name_to_i = {}
+
+    def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]:
```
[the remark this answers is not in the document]
kadarakos: The typing information for `truths` was left out originally I think because it was a bit tricky to get `mypy` to typecheck with some many different possibilities. But I'll try to add it, i think the correct type in the `legacy` loss is `Union[List[int], List[str], Ints1d, Floats2d]` for `truths`.

thinc/legacy/loss.py
```diff
@@ -0,0 +1,266 @@
...
+        missing_value = self.missing_value
+        # Convert list of ints or list of strings
+        if isinstance(truths, list):
+            truths = list(truths)
+            if len(truths):
+                if isinstance(truths[0], int):
+                    for i, value in enumerate(truths):
+                        if value == missing_value:
+                            missing.append(i)
+                else:
```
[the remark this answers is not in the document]
kadarakos: You mean `int`s here and then `str`s in the latter block?

thinc/legacy/loss.py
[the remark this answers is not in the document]
kadarakos: added the type to `truths` along with a small forest of `cast`s to type-check with `mypy`.

thinc/legacy/loss.py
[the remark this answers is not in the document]
kadarakos: I interleaved the type-checking with the conversion as in the new version of the cross-entropy so that we don't loop over all values twice.

thinc/legacy/loss.py
```diff
@@ -55,6 +65,12 @@ def convert_truths(self, truths, guesses: Floats2d) -> Tuple[Floats2d, Floats2d]
                         )
                         raise ValueError(msg)
                     for i, value in enumerate(truths):
+                        if not isinstance(value, str):
+                            raise ValueError(
+                                "All values in the truths list have to "
+                                "have the same time. The first value was "
+                                f"detected to be integer, but found {type(value)}."
```
shadeMe, line 72:
```suggestion
                                f"detected to be string, but found {type(value)}."
```

thinc/legacy/loss.py
```diff
@@ -0,0 +1,266 @@
...
+                n_classes=guesses.shape[-1],
+                label_smoothing=self.label_smoothing,
+            )
+        else:
+            if self.label_smoothing:
+                raise ValueError(
+                    "Label smoothing is only applied, when truths have type "
+                    "List[str], List[int] or Ints1d, but it seems like Floats2d "
+                    "was provided."
+                )
```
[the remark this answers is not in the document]
kadarakos: First I changed it but i reverted it back. The `else` refers to the situation where the `truths` are 2D meaning it has to be `Floats2d`. In this particular case the new typed code has to create the `truths_2d` variable and also check whether `label_smoothing` is needed. If it is needed, but the input was provided already in the `Floats2d` format we refuse to apply it.

thinc/tests/test_loss.py
```diff
@@ -1,233 +1,602 @@
...
+from functools import partial
 from thinc.api import CategoricalCrossentropy, SequenceCategoricalCrossentropy
-from thinc.api import L2Distance, CosineDistance
+from thinc.api import L2Distance, CosineDistance, softmax_activation
+from thinc.api import Ragged
 from thinc import registry
+from thinc.util import has_torch, to_categorical
+from hypothesis import given, settings
+from hypothesis.strategies import integers, floats
+from thinc.legacy import loss
```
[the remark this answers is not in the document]
kadarakos: Without this import the `registry` cannot resolve the losses.

thinc/legacy/loss.py
```diff
@@ -0,0 +1,266 @@
...
+        missing = []
+        negatives_mask = None
+        if self.names:
+            negatives_mask = xp.ones((len(truths), len(self.names)), dtype="f")
+        missing_value = self.missing_value
+        # Convert list of ints or list of strings
+        if isinstance(truths, list):
+            truths = list(truths)
+            if len(truths):
+                if isinstance(truths[0], int):
```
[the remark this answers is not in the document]
shadeMe: Yeah, looks good!

thinc/loss.py
kadarakos: I moved it out of the loss and then it can have its more true `arr: ArrayXd` signature.

thinc/loss.py
kadarakos: I think this is addressed by now right @danieldk ? in the new `_validate_input` of the `CategoricalCrossentropyBase` ?

shadeMe:
Adding this note here so that we don't forget - this PR should be rebased on/merged into `v9`.

danieldk (approved):

PR #650, opened 2022-05-02
Check Ops.sigmoid against PyTorch

danieldk:
A bit of a follow up to @shadeMe's sigmoid PR: let's check the output against Torch.


danieldk:

shadeMe (approved):

PR #652, opened 2022-05-03
Move compatiblity-related code into a separate `compat` module

shadeMe:
There are multiple places in the codebase where we attempt to import libraries inside `try...expect` clauses to safely determine their availability. This PR is primarily meant to replace the individual imports with code that calls into a central `compat` module that maintains the imports for all optional libraries.

Additional changes:
* Removal of the `has_cupy` exports from the `cupy_ops`~, `backends` and `api` modules (these were not explicitly exposed to the public API but used internally)~.
* Handle the deprecation of `cupy.fromDlpack` in `cupy >= 10.0.0`.
* ~`util.set_active_gpu` returns `None` when `cupy`/GPU support is unavailable. Prior to this change, the function could potentially raise an unhandled exception if the `cupy` library was not installed.~ `util.set_active_gpu` now raises an exception if no GPU support is detected.

### DIFF
--- a/examples/00_intro_to_thinc.ipynb
+++ b/examples/00_intro_to_thinc.ipynb
@@ -44,9 +44,9 @@
    },
    "outputs": [],
    "source": [
-    "import thinc.util\n",
+    "import thinc.compat\n",
     "# If you want to run this notebook on GPU, you'll need to install cupy.\n",
-    "if not thinc.util.has_cupy:\n",
+    "if not thinc.compat.has_cupy:\n",
     "    !pip install \"cupy-cuda101\""
    ]
   },
@@ -59,11 +59,11 @@
    },
    "outputs": [],
    "source": [
-    "import thinc.util\n",
+    "import thinc.compat\n",
     "# If you want to try out the tensorflow integration, you'll need to install that.\n",
     "# You'll either need to do tensorflow or tensorflow-gpu, depending on your\n",
     "# requirements.\n",
-    "if not thinc.util.has_tensorflow:\n",
+    "if not thinc.compat.has_tensorflow:\n",
     "    !pip install \"tensorflow-gpu>=2\""
    ]
   },
@@ -75,9 +75,9 @@
    },
    "outputs": [],
    "source": [
-    "import thinc.util\n",
+    "import thinc.compat\n",
     "# If you want to try out the PyTorch integration, you'll need to install it.\n",
-    "if not thinc.util.has_torch:\n",
+    "if not thinc.compat.has_torch:\n",
     "    !pip install \"torch\""
    ]
   },
@@ -89,9 +89,9 @@
    },
    "outputs": [],
    "source": [
-    "import thinc.util\n",
+    "import thinc.compat\n",
     "# If you want to try out the MxNet integration, you'll need to install it.\n",
-    "if not thinc.util.has_mxnet:\n",
+    "if not thinc.compat.has_mxnet:\n",
     "    !pip install \"mxnet>=1.5.1,<1.6.0\""
    ]
   },
@@ -117,10 +117,10 @@
    "outputs": [],
    "source": [
     "from thinc.api import prefer_gpu\n",
-    "import thinc.util\n",
+    "import thinc.compat\n",
     "print(\"Thinc GPU?\", prefer_gpu())\n",
     "\n",
-    "if thinc.util.has_tensorflow:\n",
+    "if thinc.compat.has_tensorflow:\n",
     "    import tensorflow as tf\n",
     "    print(\"Tensorflow GPU?\", bool(tf.config.experimental.list_physical_devices('GPU')))"
    ]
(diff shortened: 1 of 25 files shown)


shadeMe:

website/docs/api-util.md
```diff
@@ -162,10 +163,10 @@ Convert a PyTorch tensor to a `numpy` or `cupy` tensor.
 Convert a `numpy` or `cupy` tensor to a TensorFlow tensor.
 
 | Argument        | Type                       | Description                                           |
-| --------------- | -------------------------- | ----------------------------------------------------- |
+| --------------- | -------------------------- | ----------------------------------------------------- | --- |
 | `xp_tensor`     | <tt>ArrayXd</tt>           | The tensor to convert.                                |
 | `requires_grad` | <tt>bool</tt>              | Whether to backpropagate through the variable.        |
-| `as_variable`   | <tt>bool</tt>              | Convert the result to a `tensorflow.Variable` object. |  |
+| `as_variable`   | <tt>bool</tt>              | Convert the result to a `tensorflow.Variable` object. |     |
```
svlandeg, line 169: It looks like the original file had a stray `|` you'll want to remove, as well as the newly added ` --- |` bit in the separator

thinc/api.py
```diff
@@ -17,7 +17,7 @@
 from .util import to_categorical, get_width, get_array_module, to_numpy
 from .util import torch2xp, xp2torch, tensorflow2xp, xp2tensorflow, mxnet2xp, xp2mxnet
 from .backends import get_ops, set_current_ops, get_current_ops, use_ops
-from .backends import Ops, CupyOps, NumpyOps, has_cupy, set_gpu_allocator
+from .backends import Ops, CupyOps, NumpyOps, set_gpu_allocator
```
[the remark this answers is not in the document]
shadeMe: The primary motivation is to reduce the surface area of the public API and have a single source of truth when it comes to presence flags. I can definitely sympathize with your apprehension about [breaking someone's code](https://xkcd.com/1172/) (especially in the case of Python, where literally everything is observable to the end user), but I don't think we are breaking any API guarantees by removing it.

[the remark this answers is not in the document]
adrianeboyd: I agree with Sofie that we should keep this. We usually consider everything that doesn't start with `_` to be part of the public API even if it isn't documented, and for this in particular there is very little cost to keeping it.

thinc/backends/_custom_kernels.py
```diff
@@ -55,7 +51,7 @@
     cupy.RawModule(
         code=KERNELS_SRC, options=("--std=c++11",), name_expressions=KERNELS_LIST
     )
-    if cupy is not None
+    if has_cupy_gpu
```
danieldk, line 54: `has_cupy` or `has_cupy_gpu` (I am not sure)? I think the kernels are only compiled at `__call__` time.

thinc/backends/__init__.py
```diff
@@ -174,5 +175,4 @@ def _create_thread_local(
     "Ops",
     "CupyOps",
     "NumpyOps",
-    "has_cupy",
```
danieldk, line 177: We should also keep this for backward compat.

examples/00_intro_to_thinc.ipynb
```diff
@@ -44,9 +44,9 @@
    },
    "outputs": [],
    "source": [
-    "import thinc.util\n",
+    "import thinc.compat\n",
```
adrianeboyd, line 47: This should remain backwards-compatible, so let's (make sure that's the case and) undo these changes to the examples.

thinc/backends/_custom_kernels.py
shadeMe: We need `has_cupy_gpu` here as `_get_kernel` gets called during module init, which in turn calls into `cupy`. The latter call will raise an exception if there are no visible GPU devices (even if `cupy` is installed, i.e, `has_cupy` returns `True`).

thinc/util.py
```diff
@@ -172,19 +123,18 @@ def to_numpy(data):  # pragma: no cover
...
-    try:
-        import torch
 
+    if has_torch_gpu:
         torch.cuda.set_device(gpu_id)
         torch.set_default_tensor_type("torch.cuda.FloatTensor")
-    except ImportError:
-        pass
+
     return device
```
svlandeg, line 138: I worry that this might be slightly breaking behaviour, too. So before we would throw an error if you'd request to set the active GPU but that failed, and now we'd just silently move on?

shadeMe: I guess it depends on how we want to view the original behaviour. Previously, we were importing `cupy` without the necessary sanity checks (like we do elsewhere), opening up the possibility of an unhandled exception. This wasn't specified in the function's documentation (and the "if available" phrase alludes to some amount of exception safety, I think), so I interpreted this lack of a sanity check as a bug.

That said, I now see that the  "if available" phrase could parsed as an adjunct to _just_ `torch`, in which case it would in fact make sense to not modify the function behaviour per-se but to explicitly raise an exception before we call into `cupy`; the `cupy` call will eventually throw under such circumstances anyway). I take that this would be preferable?

svlandeg: I read it as the "if available" being applicable only to the "torch" bit, but it's definitely ambiguous. 

Raising an exception would make sense to me. The docs provide little additional clarification (let's add some) but the current typing to `cupy.cuda.Device` (without `Optional`) signals to me that we'd always like this to work, and if we can't get a proper device, the function should raise an exception.

svlandeg (commented):
This looks good to merge to me - just needs to target `master` instead and fix any potential merge conflicts ;-)

shadeMe:
No conflicts! Such a good feeling :grin:

svlandeg:
Just a flaky test. Let's rerun it and see what happens ;-)

PR #656, opened 2022-05-09
NumpyOps: Better type-casting in `asarray`

shadeMe:
Handle type-casting in all cases.

### DIFF
--- a/thinc/backends/numpy_ops.pyx
+++ b/thinc/backends/numpy_ops.pyx
@@ -61,13 +61,11 @@ class NumpyOps(Ops):
             raise ValueError("BLIS support requires blis: pip install blis")
 
     def asarray(self, data, dtype=None):
-        dtype = {"dtype": dtype} if dtype is not None else {}
-
         array = None
         cast_array = True
 
         if isinstance(data, self.xp.ndarray):
-            array = self.xp.asarray(data, **dtype)
+            array = self.xp.asarray(data, dtype=dtype)
             cast_array = False
         elif hasattr(data, 'numpy'):
             # Handles PyTorch Tensor
@@ -77,8 +75,8 @@ class NumpyOps(Ops):
         else:
             array = self.xp.array(data)
 
-        if cast_array and dtype != {}:
-            array = array.astype(dtype["dtype"])
+        if cast_array and dtype is not None:
+            array = array.astype(dtype=dtype)
 
         return array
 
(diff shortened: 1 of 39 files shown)


thinc/backends/numpy_ops.pyx
```diff
@@ -61,20 +61,27 @@ class NumpyOps(Ops):
...
+
+        array = None
+        cast_array = True
+
         if isinstance(data, self.xp.ndarray):
-            if dtype is not None:
-                return self.xp.asarray(data, dtype=dtype)
-            else:
-                return self.xp.asarray(data)
+            array = self.xp.asarray(data, **dtype)
```
adrianeboyd, line 70: Can't we just keep this as `dtype=dtype` and not create a dict at all?

adrianeboyd: Oh wait, now I see what you're doing. Hmm...

adrianeboyd: I'm not sure this is the worth the tradeoff for `dtype != {}` and `dtype["dtype"]`.

shadeMe: Yeah, you're right - We should be able to pass `dtype=None` to `xp.asarray` without an additional check.

svlandeg (commented):
We're retiring `develop` - can you target `master` instead? 🙏

thinc/backends/numpy_ops.pyx
```diff
@@ -61,20 +61,25 @@ class NumpyOps(Ops):
             raise ValueError("BLIS support requires blis: pip install blis")
 
     def asarray(self, data, dtype=None):
+        array = None
```
adrianeboyd, line 64: Unless there's a related check of some sort for a failure, this seems unnecessary.

```suggestion
```

adrianeboyd:
Let's make the same change removing `cast_array` here, too?

thinc/backends/numpy_ops.pyx
```diff
@@ -62,19 +62,20 @@ class NumpyOps(Ops):
 
     def asarray(self, data, dtype=None):
         if isinstance(data, self.xp.ndarray):
-            if dtype is not None:
-                return self.xp.asarray(data, dtype=dtype)
-            else:
-                return self.xp.asarray(data)
+            array = self.xp.asarray(data)
```
danieldk, line 65: Is this now still needed, since data is already `xp.ndarray`? Also see `Ops.asarray` for the expected semantics.

danieldk (approved):

adrianeboyd (approved):

PR #661, opened 2022-05-11
`CupyOps`: Simplify `asarray`

shadeMe:


adrianeboyd:
Looking at the numpy/cupy docs again, I think you can avoid `cast_array` entirely by using:

```python
array = array.astype(dtype=dtype, copy=False)
```

danieldk (approved):

thinc/backends/cupy_ops.py
```diff
@@ -59,29 +59,20 @@ def gemm(self, x, y, out=None, trans1=False, trans2=False):
...
-        # This is sort of frustrating, but we can't easily otherwise pass
-        # forward "unset".
-        dtype = {"dtype": dtype} if dtype is not None else {}
-
         # We'll try to perform a zero-copy conversion if possible.
-        array = None
-        cast_array = False
         if is_cupy_array(data):
-            array = self.xp.asarray(data, **dtype)
+            array = self.xp.asarray(data)
```
danieldk, line 64: Is this `asarray` needed if the data is already a CuPy array?

danieldk (approved):

adrianeboyd (approved):

PR #663, opened 2022-05-11
Fix a unit test in the PyTorch wrapper

danieldk:
This test checked whether the allocator was set to the PyTorch allocator when the PyTorch shim is used. However, this is not the case when PyTorch is installed, but CuPy isn't, so the test would fail. Since this test relies on CuPy, disable it when CuPy is not available.


danieldk:
@adrianeboyd I think you added the pool assertion, so you are probably the best reviewer for this PR.

PR #664, opened 2022-05-13
Fix out-of-bounds writes in NumpyOps/CupyOps

danieldk:
- Using `{CupyOps,NumpyOps}.adam` with incompatible shapes for weights,
  gradients, or moments resulted in out-of-bound writes.
- Using `NumpyOps.adam` with non-float32 arrays resulted filling arrays
  with incorrect data.

### DIFF
--- a/thinc/backends/ops.py
+++ b/thinc/backends/ops.py
@@ -1112,13 +1112,19 @@ class Ops:
         learn_rate: float,
         mod_rate: float = 1.0,
     ) -> Tuple[Floats1d, Floats1d, Floats1d, Floats1d]:
+        _check_compatible_shape(weights, gradient)
+        _check_compatible_shape(weights, mom1)
+        _check_compatible_shape(weights, mom2)
+
         # Internals for optimizer
         mom1 *= beta1
         mom2 *= beta2
+        print(mom1.shape, mom2.shape, gradient.shape, weights.shape)
         mom1 += gradient * (1.0 - beta1)
         mom2 += gradient * gradient * (1.0 - beta2)
         # Here we assume learn rate is calculated by the caller.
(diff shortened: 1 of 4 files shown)


thinc/backends/ops.py
```diff
@@ -1112,13 +1112,19 @@ def adam(
...
         mod_rate: float = 1.0,
     ) -> Tuple[Floats1d, Floats1d, Floats1d, Floats1d]:
+        _check_compatible_shape(weights, gradient)
+        _check_compatible_shape(weights, mom1)
+        _check_compatible_shape(weights, mom2)
+
         # Internals for optimizer
         mom1 *= beta1
         mom2 *= beta2
+        print(mom1.shape, mom2.shape, gradient.shape, weights.shape)
```
adrianeboyd, line 1122:
```suggestion
```

thinc/backends/ops.py
```diff
@@ -1112,13 +1112,19 @@ def adam(
...
         mom1 += gradient * (1.0 - beta1)
         mom2 += gradient * gradient * (1.0 - beta2)
         # Here we assume learn rate is calculated by the caller.
         # cdef weight_t a_t = learn_rate * sqrt(1-beta2**hp.t) / (1-beta1**hp.t);
+        print(mom1.shape, mom2.shape, weights.shape)
```
adrianeboyd, line 1127:
```suggestion
```

svlandeg (approved):
Definitely makes sense to me to have this check!

This might be breaking upstream code, but then it probably should?

danieldk:
> This might be breaking upstream code, but then it probably should?

Yeah, we want it to break, since overwriting memory is worse. No issues in spaCy though.

PR #668, opened 2022-05-18
Backport #663 to v8.0.x

danieldk:
Backport #663 to v8.0.x:

This test checked whether the allocator was set to the PyTorch allocator
when the PyTorch shim is used. However, this is not the case when
PyTorch is installed, but CuPy isn't, so the test would fail. Since this
test relies on CuPy, disable it when CuPy is not available.


danieldk:

PR #674, opened 2022-05-20
Remove use of `torch.set_default_tensor_type`

danieldk:
**Initial remark:** this PR doesn't implement all the functionality that I am aiming at yet (e.g. support other than `cpu` and `cuda`) devices. I want to split this into multiple PRs to make sure that it is all reviewable. The `spacy-transformers` tests pass on the GPU and after this PR, training and evaluating a transformer works.

---

This PR removes use of `torch.set_default_tensor_type`. There are various reasons why we should probably move away from using this function:

- Upstream will deprecate and remove it: https://github.com/pytorch/pytorch/issues/53124
- We cannot use this mechanism for other devices than CPU/CUDA, such as Metal Performance Shaders.
- It offers little flexibility in allocating Torch models on different devices.

This PR makes `PyTorchWrapper`/`PyTorchShim` flexible in terms of the devices it can use. Both classes add a `device` argument to their constructors that takes a `torch.device` instance. The shim ensures that the model is on the given device. The wrapper ensures that input tensors are on the correct device, by calling `xp2torch` with the new `device` keyword argument.

Even though this approach offers more flexibility, as a default we want to use the `cpu` device when `NumpyOps` is used and `cuda:N` when CupyOps is used. In order to do so, this PR also adds a new function `get_torch_default_device` that returns the correct device for the currently active Ops. `PyTorchWrapper`/`PyTorchShim`/`xp2torch` use this function when `None` is given as the device to fall back on this default, mimicking the behavior from before this PR.

### DIFF
--- a/thinc/shims/pytorch.py
+++ b/thinc/shims/pytorch.py
@@ -5,6 +5,7 @@ import itertools
 import srsly
 
 from ..util import torch2xp, xp2torch, convert_recursive, iterate_recursive
+from ..util import get_torch_default_device
 from ..compat import torch
 from ..backends import get_current_ops, context_pools, CupyOps
 from ..backends import set_gpu_allocator
@@ -25,6 +26,10 @@ class PyTorchShim(Shim):
         The gradient scaler to use for mixed-precision training. If this
         argument is set to "None" and mixed precision is enabled, a gradient
         scaler with the default configuration is used.
+    device:
+        The PyTorch device to run the model on. When this argument is
+        set to "None", the default device for the currently active Thinc
+        ops is used.
     """
 
     def __init__(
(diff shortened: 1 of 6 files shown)


thinc/shims/pytorch.py
```diff
@@ -34,12 +39,20 @@ def __init__(
         optimizer: Any = None,
         mixed_precision: bool = False,
         grad_scaler: Optional[PyTorchGradScaler] = None,
+        device: Optional["torch.device"] = None,
     ):
         super().__init__(model, config, optimizer)
 
+        if device is None:
+            device = get_torch_default_device()
+        if model is not None:
```
shadeMe, line 48: Is this check necessary? The rest of the code appears to have an implicit pre-condition that `model` is always a valid object (of the expected type).

danieldk: I had to add it because it occurred. IIRC it's when an existing transformer is deserialized. The shim is first constructed with `model=None` and then the model is put in place by deserialization.

shadeMe (approved):
LGTM.

svlandeg (approved):
Looks good to me as well :-)

danieldk:
Cool, I'll merge it then :), it's on a branch anyway.

PR #675, opened 2022-05-23
Replace use of gpu_is_available with has_cupy_gpu

danieldk:
This PR is in preparation of better non-CUDA device support. Once we support non-CUDA GPUs, there may be GPUs available that are not 'CuPy GPUs'. In all places where we use `gpu_is_available` we actually mean: is 'CuPy available with a CUDA GPU'? So, this PR replaces uses of `gpu_is_available` by `has_cupy_gpu`. This allows us to use `gpu_is_available` in the future to check if any GPU is available.

In addition to that, some code had expressions like

```
has_cupy and gpu_is_available()
```

This PR simplify such conditions to `has_cupy_gpu`, since `has_cupy_gpu` implies `has_cupy`.

### DIFF
--- a/thinc/util.py
+++ b/thinc/util.py
@@ -126,7 +126,7 @@ def to_numpy(data):  # pragma: no cover
 def set_active_gpu(gpu_id: int) -> "cupy.cuda.Device":  # pragma: no cover
     """Set the current GPU device for cupy and torch (if available)."""
     if not has_cupy_gpu:
-        raise ValueError("No GPU devices detected")
+        raise ValueError("No CUDA GPU devices detected")
 
     device = cupy.cuda.device.Device(gpu_id)
     device.use()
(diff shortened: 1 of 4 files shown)


shadeMe (approved):
LGTM.

thinc/util.py
```diff
@@ -163,7 +161,7 @@ def prefer_gpu(gpu_id: int = 0) -> bool:  # pragma: no cover
 def require_gpu(gpu_id: int = 0) -> bool:  # pragma: no cover
     from .backends import set_current_ops, CupyOps
 
-    if not gpu_is_available():
+    if not has_cupy_gpu:
         raise ValueError("No GPU devices detected")
```
svlandeg, line 165: For now perhaps we want to make this error more specific to say "No (CUDA) GPU devices detected" ?

danieldk: Fixed, though I didn't use parentheses.

PR #677, opened 2022-05-23
`Shim`: Fix potential data race when allocated on different threads

shadeMe:
Moves both volatile reads of the `global_id` global variable into the critical section to prevent a write (from another thread) interleaving them and potentially resulting in multiple `Shim` instances receiving the same ID.


danieldk (approved):

danieldk:

PR #678, opened 2022-05-23
Prevent test failures on Azure CI hosts running Windows

shadeMe:
By default, `pytest` fails when it encounters unexpected variability in test execution times between consecutive runs (with different parameters, etc). This was leading to occasional build failures on Windows hosts.  
 
The actual cause for this variability (specifically on Windows) is currently unknown, partly owing to the lack of options when it comes to debugging such non-deterministic issues on the remote machine.


svlandeg:
It looks like this test has some [history](https://github.com/explosion/thinc/commit/6909185e9401a71f6014a007b102c4063aa90e5d
) :p

PR #679, opened 2022-05-24
Make the compat module private

danieldk:
As discussed, this module contains a lot of booleans that Thinc uses internally to keep track of what libraries and devices are available. However, we want to keep the freedom to rename or change the semantics of these booleans, so let's make the module private.


shadeMe (approved):
LGTM except for the merge conflicts.

danieldk:
> Can you change all the imports to import from `util` or `api` as much as possible rather than from `_compat`?

Do we really want this? To me all the `has_.*` are effectively private Thinc state that we use to keep track of what libraries/GPUs are available. Exposing them to the outside makes it harder to change/improve the semantics of these variables. E.g., currently we have `has_torch_gpu` which flags 'torch has access to a CUDA GPU', but should eventually mean 'torch has access to a GPU'. However, if we expose this officially through `thinc.util` we cannot give them the intended semantics, because someone may rely on `has_torch_gpu` to check whether there is a CUDA GPU.

The only use in spaCy is `has_cupy` in a condition

```
if has_cupy and gpu_is_available():
```

however, `gpu_is_available()` alone would suffice.

danieldk:
I have changed my view a bit on this module, and I think it's sometimes useful to probe whether Thinc has found CuPy, a GPU, etc. So I'll close this issue.

PR #680, opened 2022-05-25
`test_to_categorical`: Ensure that `label_smoothing < 0.5`

shadeMe:
`to_categorical` expects `0.0 <= label_smoothing < 0.5`.

### DIFF
--- a/thinc/tests/test_util.py
+++ b/thinc/tests/test_util.py
@@ -46,7 +46,7 @@ def test_array_module_cpu_gpu_helpers():
     assert not is_numpy_array((1, 2))
 
 
-@given(label_smoothing=strategies.floats(min_value=0.0, max_value=0.5))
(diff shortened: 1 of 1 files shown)


thinc/tests/test_util.py
```diff
@@ -46,7 +46,7 @@ def test_array_module_cpu_gpu_helpers():
     assert not is_numpy_array((1, 2))
 
 
-@given(label_smoothing=strategies.floats(min_value=0.0, max_value=0.5))
+@given(label_smoothing=strategies.floats(min_value=0.0, max_value=0.49))
```
danieldk, line 49: I think it's better to keep it at 0.5 and use `exclude_max`:

https://hypothesis.readthedocs.io/en/latest/data.html?highlight=floats#hypothesis.strategies.floats

danieldk:
The remaining failures are in relu, so let's merge this...

PR #681, opened 2022-05-27
test_ops: do not lower precision in conversion to Torch tensor

danieldk:
float64 test values close to zero were rounded by conversion to a float32 Torch tensor, resuling in mismatches between Thinc and Torch gradients. This change prevents the loss in precision.


danieldk:

kadarakos (approved):

kadarakos (approved):

PR #685, opened 2022-05-29
Add support for PyTorch Metal Performance Shaders

danieldk:
Nightly PyTorch versions add support for Metal Performance Shaders (MPS). Metal is a low-level graphics API for Apple platforms that also supports compute kernels (shaders). MPS is a framework of highly-optimized compute and graphics kernels, including kernels for neural networks. MPS is supported on both Apple Silicon, such as the M1 family of SoC, as well as a range of AMD GPUs used in Macs.

Since devices are handled in Thinc through a specific `Ops` implementation (e.g. `CupyOps` == CUDA GPUs), this change introduces the `MPSOps` class. This class is a subclass of `NumpyOps` or `AppleOps` (when available). `MPSOps` does not override any methods, but is used to signal to relevant code paths (e.g. `xp2torch`) that Torch tensors should be placed on the MPS device.

The mapping in the previously introduced `get_torch_default_device` function is updated to:

- `NumpyOps` -> `cpu`
- `CupyOps` -> `cuda:N`, where N is the selected CUDA device.
- `MPSOps` -> `mps`

to ensure placement of Torch tensors on the `mps` device when `MPSOps` is active.

Finally, the following booleans have been added to or changed in `compat`:

- `has_torch_mps` (new): PyTorch has MPS support
- `has_torch_mps_gpu` (new): PyTorch has MPS support and an
  MPS-capable GPU is available.
- `has_torch_cuda_gpu` (new): PyTorch has CUDA support and a
  CUDA-capable GPU is available.
- `has_torch_gpu` (changed): PyTorch has a GPU available (CUDA
  or MPS).

### DIFF
--- a/thinc/backends/mps_ops.py
+++ b/thinc/backends/mps_ops.py
@@ -1,14 +1,21 @@
+from typing import TYPE_CHECKING
 import numpy
 
 from .. import registry
-from . import NumpyOps
+from . import NumpyOps, Ops
 
-try:
-    from thinc_apple_ops import AppleOps
+if TYPE_CHECKING:
+    # Type checking does not work with dynamic base classes, since MyPy cannot
+    # determine against which base class to check. So, always derive from Ops
+    # during type checking.
+    _Ops = Ops
+else:
+    try:
+        from thinc_apple_ops import AppleOps
 
-    _Ops = AppleOps
-except:
-    _Ops = NumpyOps
+        _Ops = AppleOps
+    except:
+        _Ops = NumpyOps
 
 
 @registry.ops("MPSOps")
--- a/thinc/backends/mps_ops.py
+++ b/thinc/backends/mps_ops.py
@@ -14,7 +14,7 @@ else:
         from thinc_apple_ops import AppleOps
 
         _Ops = AppleOps
-    except:
+    except ImportError:
         _Ops = NumpyOps
 
 
--- a/thinc/backends/mps_ops.py
+++ b/thinc/backends/mps_ops.py
@@ -1,0 +1,19 @@
+import numpy
+
+from .. import registry
+from . import NumpyOps
+
+try:
+    from thinc_apple_ops import AppleOps
+
+    _Ops = AppleOps
+except:
+    _Ops = NumpyOps
+
+
+@registry.ops("MPSOps")
+class MPSOps(_Ops):
+    """Ops class for Metal Performance shaders."""
+
+    name = "mps"
+    xp = numpy
--- a/thinc/compat.py
+++ b/thinc/compat.py
@@ -38,7 +38,7 @@ try:  # pragma: no cover
         and not torch.cuda.amp.common.amp_definitely_not_available()
     )
 except ImportError:  # pragma: no cover
-    torch = None
+    torch = None  # type: ignore
     has_torch = False
     has_torch_gpu = False
     has_torch_amp = False
--- a/thinc/compat.py
+++ b/thinc/compat.py
@@ -37,7 +37,7 @@ try:  # pragma: no cover
         and torch.has_mps
         and torch.backends.mps.is_available()
     )
-    has_torch_gpu = has_torch_cuda_gpu or has_torch_mps_gpu
+    has_torch_gpu = has_torch_cuda_gpu
     torch_version = Version(str(torch.__version__))
     has_torch_amp = (
         torch_version >= Version("1.9.0")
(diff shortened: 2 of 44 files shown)


danieldk:
> The diff should get less messy after #692 is merged.

The history is clean again :).

thinc/backends/mps_ops.py
```diff
@@ -0,0 +1,26 @@
...
+    # determine against which base class to check. So, always derive from Ops
+    # during type checking.
+    _Ops = Ops
+else:
+    try:
+        from thinc_apple_ops import AppleOps
+
+        _Ops = AppleOps
+    except ImportError:
+        _Ops = NumpyOps
```
shadeMe, line 18: Do we want to move this to the `_compat` module (and add an additional `bool` presence flag)?

shadeMe (approved):
LGTM with just one comment.

thinc/backends/mps_ops.py
danieldk: That would be nice! I am a bit worried about introducing import cycles again between thinc and thinc-apple-ops. Let me look if it's possible.

danieldk: Would introduce another cycle. `AppleOps` requires `NumpyOps`. `NumpyOps` requires `util`, `util` requires `compat` and `compat` would import `thinc_apple_ops`.

shadeMe: Ah, shame.

thinc/compat.py
```diff
@@ -31,7 +31,10 @@
     import torch
 
     has_torch = True
-    has_torch_gpu = torch.cuda.device_count() != 0
+    has_torch_cuda_gpu = torch.cuda.device_count() != 0
+    has_torch_mps = hasattr(torch, "has_mps") and torch.has_mps
+    has_torch_mps_gpu = has_torch_mps and torch.backends.mps.is_available()
```
svlandeg, line 36: Is `has_torch_mps` by itself actually useful? (if `has_torch_mps_gpu` is `False`)
Asking to figure out whether we actually need to keep track of it?

thinc/compat.py
```diff
@@ -31,7 +31,10 @@
...
+    has_torch_gpu = has_torch_cuda_gpu or has_torch_mps_gpu
```
svlandeg, line 37: Will this redefinition trip people up if they're using code that checks this boolean? Because we're editing some of our own tests to now use `has_torch_cuda_gpu` instead of `has_torch_gpu`, it feels like this could be semantically breaking?

thinc/util.py
```diff
@@ -50,7 +54,7 @@ def get_array_module(arr):  # pragma: no cover
 
 
 def gpu_is_available():
-    return has_cupy_gpu
+    return has_cupy_gpu or has_torch_mps_gpu
```
svlandeg, line 57: Why don't we define a `has_gpu` var that incorporates this `or` logic? It would make the check in `prefer_gpu` less awkward, too.

svlandeg (commented):
This rocks! Just had a few small comments but otherwise good to merge!

thinc/compat.py
danieldk: I wasn't sure about that either. On the one hand yes, on the other hand why would anybody use this? If I want to know if Torch thinks there are GPUs I would just ask Torch?

I am also wondering if we should underscore-prefix the new (since Thinc 8.1) `has_` variables to avoid these compatibility issues in the future?

thinc/util.py
danieldk: Added and updated checks.

thinc/compat.py
danieldk: Removed `has_torch_mps`.

thinc/compat.py
danieldk: I have reverted this to `has_torch_gpu = has_torch_cuda_gpu`, since `has_torch_gpu = has_torch_cuda_gpu or has_torch_mps_gpu` is not needed for this PR. I think it's still something that we have to device before releasing 8.1.

thinc/util.py
```diff
@@ -176,17 +176,17 @@ def require_cpu() -> bool:  # pragma: no cover
...
 def prefer_gpu(gpu_id: int = 0) -> bool:  # pragma: no cover
     """Use GPU if it's available. Returns True if so, False otherwise."""
-    if not (has_cupy_gpu or has_torch_mps_gpu):
-        return False
-    else:
+    if has_gpu:
         require_gpu(gpu_id=gpu_id)
         return True
+    else:
+        return False
```
svlandeg, line 183:
```suggestion
    return has_gpu
```
(Feel free to ignore if you don't like it better - I don't feel strongly)

svlandeg (approved):

PR #688, opened 2022-06-01
`test_model_gpu`: Use TF memory pool if available, feature-gate test

shadeMe:


shadeMe:

thinc/tests/model/test_model.py
```diff
@@ -4,9 +4,10 @@
 import time
 from thinc.api import Adam, CupyOps, Dropout, Linear, Model, Relu
 from thinc.api import Shim, Softmax, chain, change_attr_values
-from thinc.api import concatenate, set_dropout_rate
-from thinc.api import use_ops, with_debug, wrap_model_recursive
-from thinc.compat import has_cupy_gpu
+from thinc.api import concatenate, set_dropout_rate, set_gpu_allocator
+from thinc.api import with_debug, wrap_model_recursive, use_ops
```
danieldk, line 8: Nitpick: why is this import list reordered? It was in alphabetic order?

thinc/tests/model/test_model.py
```diff
@@ -404,15 +405,17 @@ def get_model_id(id_list, index):
     assert len(list_of_ids) == len(list(set(list_of_ids)))
 
 
+@pytest.mark.skipif(not gpu_is_available(), reason="needs GPU")
```
danieldk, line 408: I'd expect

```suggestion
@pytest.mark.skipif(not (has_tensorflow_gpu and has_cupy_gpu)), reason="needs GPU")
```

Or something like that.

We should avoid `gpu_is_available` for specific tests, since it will also return `True` in the future for non-CUDA GPUs (e.g. MPS).

thinc/tests/model/test_model.py
```diff
@@ -404,15 +405,17 @@ def get_model_id(id_list, index):
...
     pytest.importorskip("ml_datasets")
     import ml_datasets
 
-    ops = "cpu"
-    if has_cupy_gpu:
-        ops = "cupy"
+    if has_tensorflow:
+        # Ensure that CuPy has enough memory as TF just
+        # loves to bogart all of the GPU's memory on init.
+        set_gpu_allocator("tensorflow")
```
danieldk, line 416: I think this resets the allocator for other tests as well? Can we somehow scope this?

thinc/tests/model/test_model.py
shadeMe: I think I removed `use_ops` at some point to try out another function and then re-added it to the end. Will fix.

thinc/tests/model/test_model.py
shadeMe: Good question! I guess I chose this location to change the allocator as this was the test that ran out of memory on the BuildKite agent. And the reason that happens is because the tests in `test_mnist`, which load and use a TF model, get executed before `test_model`, allowing TF to allocate almost all remaining GPU memory. From that point onwards, any allocations by `cupy` or `torch` run the risk of triggering an OOM error.

I agree with you that this is just papering over the cracks and that the most sensible way to fix this would be to implement a resource manager for TF ops. Unfortunately, there appears to be no [way to programmatically coerce TF into releasing its memory pool](https://github.com/tensorflow/tensorflow/issues/36465) 😭 - The only time it happens is when the host process gets terminated and the GPU driver (presumably) releases its GPU resources.

The only other alternative we have is to selectively execute TF tests in a child process. But then we will break the `pytest` machinery that captures the test result and its `stdout/err` output. And there seems to be [no portable and fool-proof ](https://stackoverflow.com/questions/48234032/run-py-test-test-in-different-process)way of executing specific `pytest` tests in child processes.

danieldk: Can you limit Tensorflow memory use with https://www.tensorflow.org/api_docs/python/tf/config/experimental/set_memory_growth ?

shadeMe: We could do it in the `compat` module, which is where we import `tensorflow` for the first time. But there's always a chance that the runtime is already initialized by then (by some other 3rd party code), leading to a `RuntimeError` exception. We can catch and handle it, but then we're back to square one.

shadeMe: Hmm, sounds feasible. I'll see how well that works.

danieldk (approved):

danieldk:

Checking if it works 😄

danieldk:

PR #695, opened 2022-06-10
Merge pytorch-device branch into master

danieldk:
This combines the previously-reviewed PRs #674 and #685.

I think the primary question is whether we want to merge this for 8.1.0 or wait for 8.1.1.


svlandeg:
Why would we wait for 8.1.1?

PR #696, opened 2022-06-13
Speed up HashEmbed layer by avoiding large temporary arrays

danieldk:
The HashEmbed layer sums up keyed embeddings. For instance, a key matrix of the shape (50000, 4) will result in 50,000 embeddings, each computed by summing 4 embeddings. The HashEmbed layer computed the embeddings as follows:

```
vectors[keys].sum(axis=1)
```

where `vectors` is an embedding matrix. However, this way of computing embeddings results in very large allocations. Suppose that `vectors` is (4000, 64). Even though the final embedding matrix is (50000, 64), the first expression will construct a temporary array of shape (50000, 4, 64). The overhead is visible in the profile of tok2vec using Thinc `master`:

before

This change avoids this by introducing a `gather_add` op as a counterpart to `scatter_add`. In this particular example, the `NumpyOps` implementation only allocates the final (50000, 64) array, computing the embeddings in-place using the BLAS saxpy function.

In benchmarks with an M1 Max on de_core_news_lg, this improved processing speed from 40511 WPS to 45591 (12.5% faster). The profile of Thinc using this change shows that far less time is spent in `HashEmbed`:

after

### DIFF
--- a/thinc/backends/cpu_kernels.hh
+++ b/thinc/backends/cpu_kernels.hh
@@ -9,7 +9,7 @@
 #include <type_traits>
 
 // Ideally we'd use an alias declaration for a generic definition of
-// *axpy. But Cython does support alias declarations yet:
+// *axpy. But Cython doesn't support alias declarations yet:
 //
 // https://github.com/cython/cython/issues/3272
 //
--- a/thinc/backends/cpu_kernels.hh
+++ b/thinc/backends/cpu_kernels.hh
@@ -8,9 +8,22 @@
 #include <string>
 #include <type_traits>
 
+// Ideally we'd use an alias declaration for a generic definition of
+// *axpy. But Cython does support alias declarations yet:
+//
+// https://github.com/cython/cython/issues/3272
+//
+// template <typename T>
+// using axpy = void (*)(int N, float alpha, const float* X, int incX,
+//                       float *Y, int incY);
+//
+// So, instead we'll do this the pre-C++11 way:
 
-typedef void (*saxpy_ptr)(int N, float alpha, const float* X, int incX,
-                          float *Y, int incY);
+template <typename T>
+struct axpy {
+    typedef void (*ptr)(int N, T alpha, const T* X, int incX, T *Y, int incY);
+};
+
 
 // All elementwise functions, such as most activations, work in-place.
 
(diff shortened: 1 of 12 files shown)


thinc/backends/cpu_kernels.hh
```diff
@@ -395,4 +398,20 @@ void backprop_seq2col(A* d_seqs, const A* d_cols, const L* lengths, L B, L I, L
...
+template <typename I, typename L>
+void cpu_gather_add(saxpy_ptr saxpy, float* out_bo, const float* table_to, const I* indices_bk, L T, L O, L B, L K)
+{
+    for (L b = 0; b < B; ++b) {
+        for (L k = 0; k < K; ++k) {
+            I idx = indices_bk[b * K + k];
+            if (idx > T) {
+                throw std::out_of_range("Embedding index out-of-bounds");
+            }
+            (*saxpy)(O, 1.0, table_to + idx * O, 1, out_bo + b * O, 1);
```
shadeMe, line 411: We should be able to call the function pointer without dereferencing it.

danieldk: Ah yes, fixed.

honnibal:
This must be good news for memory usage on the GPU too right?

danieldk:
> This must be good news for memory usage on the GPU too right?

I still have to write a CUDA kernel, but then it should definitely help.

thinc/backends/cpu_kernels.hh
```diff
@@ -8,6 +8,22 @@
 #include <string>
 #include <type_traits>
 
+// Ideally we'd use an alias declaration for a generic definition of
+// *axpy. But Cython does support alias declarations yet:
```
svlandeg, line 12:
```suggestion
// *axpy. But Cython doesn't support alias declarations yet:
```

svlandeg (commented):
Another 🤩 PR! 

Shall we also document `Ops.gather_add` in `api-backends`?

danieldk:
> Shall we also document `Ops.gather_add` in `api-backends`?

Good idea! I'll add docs as well.

danieldk:
Added docs. Plus some additional checks.

thinc/backends/ops.py
```diff
@@ -1254,6 +1254,9 @@ def position_encode(
         numpy_ops = NumpyOps()
         return self.asarray2f(numpy_ops.position_encode(N, D, period, out))
 
+    def gather_add(self, table: Floats2d, indices: Ints2d) -> FloatsXd:
```
svlandeg, line 1257: Isn't the return value `Floats2d`?
(if it is - we don't need the `cast` in `hashembed.py`)

danieldk: Yeah, fixed!

svlandeg (approved):
This looks good to merge!

shadeMe (approved):

PR #697, opened 2022-06-14
`NumpyOps`: Do not use global for `CBlas`

shadeMe:
To avoid downstream users from changing the pointers and modifying the behaviour of the struct globally.


shadeMe:

danieldk (approved):

PR #700, opened 2022-06-15
Make `CBlas` methods standalone functions to avoid using vtables

danieldk:
When testing #696, we found that adding new `CBlas` methods results in an ABI compatibility. This would mean that every time we add a CBlas method, we also have to rebuild spaCy.

The ABI incompatibility occurs because Cython generates a vtable for `cdef` methods, even when the class or its methods are final. This vtable is used by the caller to look up the (address of) the methods. When methods are added, the vtable of the caller is out-of-sync when the calling code is not recompiled.

This change works around this issue by making the methods of `CBlas` standalone functions.

Many thanks to @shadeMe for discussing the solution space.

### DIFF
--- a/thinc/backends/cblas.pxd
+++ b/thinc/backends/cblas.pxd
@@ -18,7 +18,13 @@ cdef struct BlasFuncs
 
 cdef class CBlas:
     cdef shared_ptr[BlasFuncs] ptr
-    cdef saxpy_ptr saxpy(self) nogil
-    cdef sgemm_ptr sgemm(self) nogil
-    cdef void set_saxpy(self, saxpy_ptr saxpy) nogil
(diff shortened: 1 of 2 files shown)


shadeMe (approved):
LGTM! Now, we just need to update [spaCy's usage](https://github.com/explosion/spaCy/search?q=cblas%28%29) of the `CBlas` wrapper.

danieldk:
> LGTM! Now, we just need to update [spaCy's usage](https://github.com/explosion/spaCy/search?q=cblas%28%29) of the `CBlas` wrapper.

Yeah, and thinc-apple-ops.

danieldk:
But in both cases, I'll do a new Thinc dev release first, so that we can bump the release in the PR.

svlandeg (approved):

PR #702, opened 2022-06-15
Unroll `argmax` in `maxout` for small sizes of `P`

danieldk:
`maxout` uses the `argmax` function to determine the index of the maximum value of each `P` inputs. `argmax` uses a generic array loop, which impedes speculative execution and could also prevent unrolling of the outer `maxout` loop.

This change unrolls `argmax` with small values of `P` using a variadic template. This leads to a small performance improvement.

Before this change, `maxout` takes ~3.9% of the runtime:

Screen Shot 2022-06-15 at 16 58 57

After this change, `maxout` takes ~2.7% of the runtime:

Screen Shot 2022-06-15 at 16 56 33

(The unrolled `argmax` function is no longer a separate unit in the profile.)

### DIFF
--- a/thinc/backends/cpu_kernels.hh
+++ b/thinc/backends/cpu_kernels.hh
@@ -25,7 +25,7 @@ argmax_result<T, L> argmax(T const *arr, L len)
         "Array should be floating point");
     static_assert(std::is_integral<L>::value, "Array length should be integral");
 
-    argmax_result<T, L> r { .max = arr[0], .max_idx = 0 };
+    argmax_result<T, L> r { arr[0],  0 };
 
     for (L i = 1; i < len; ++i) {
         if (arr[i] > r.max) {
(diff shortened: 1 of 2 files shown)


thinc/backends/cpu_kernels.hh
```diff
@@ -11,23 +11,58 @@
...
+template <typename T, typename L>
+argmax_result<T, L> argmax(T const *arr, L len)
 {
-    static_assert(std::is_floating_point<A>::value,
+    static_assert(std::is_floating_point<T>::value,
         "Array should be floating point");
     static_assert(std::is_integral<L>::value, "Array length should be integral");
 
-    L max = 0;
+    argmax_result<T, L> r { .max = arr[0], .max_idx = 0 };
```
shadeMe, line 28: This type of initializer requires C++20, IIRC. Surprised that it compiles with C++11.

shadeMe (approved):

PR #703, opened 2022-06-17
`get_array_module` with non-array input returns `None`

kadarakos:
The `get_array_module` utility had the behavior of returning `numpy` as the array module if the input to the function is not a `cupy` array. I think this can lead to some hard to catch errors so this PR suggests to return `None` in case the input is neither a `cupy` nor a `numpy` array.


thinc/util.py
```diff
@@ -47,10 +47,12 @@ def get_torch_default_device() -> "torch.device":
...
 
 def get_array_module(arr):  # pragma: no cover
-    if is_cupy_array(arr):
+    if is_numpy_array(arr):
+        return numpy
+    elif is_cupy_array(arr):
         return cupy
     else:
-        return numpy
+        return None
```
shadeMe, line 55: I'd say that it would be more appropriate to raise a `ValueError` here, given (my reading of) the function's semantics (`arr` is implicitly expected to be an array type, so any other input is inherently invalid in the context of the function).

thinc/tests/test_util.py
```diff
@@ -39,11 +46,22 @@ def test_get_width_fail(obj):
...
+    with pytest.raises(ValueError, match=error):
+        get_array_module(0)
+    zeros = xp.zeros((1, 2))
+    xp_ = get_array_module(zeros)
+    assert hasattr(xp_, "ndarray")
+    if xp == numpy:
+        assert is_numpy_array(zeros)
+        assert not is_numpy_array((1, 2))
+    else:
+        from thinc.util import is_cupy_array
```
shadeMe, line 62: I think we can move this import to the top of the module.

shadeMe (approved):
LGTM! Just one, minor change.

thinc/util.py
```diff
@@ -47,10 +47,15 @@ def get_torch_default_device() -> "torch.device":
...
-    if is_cupy_array(arr):
+    if is_numpy_array(arr):
+        return numpy
+    elif is_cupy_array(arr):
         return cupy
     else:
-        return numpy
+        raise ValueError(
+            "Input was neither a numpy or cupy array"
+            f", but was of type {type(arr)}."
```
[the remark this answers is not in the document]
kadarakos: Currently `util.py` is technically public right?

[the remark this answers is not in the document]
svlandeg: yes, and technically `get_array_module` is not private so users could call it directly

thinc/tests/test_util.py
```diff
@@ -39,11 +47,23 @@ def test_get_width_fail(obj):
...
+def test_array_module_cpu_gpu_helpers(xp):
+    error = ("Only numpy and cupy arrays are supported"
+             ", but found <class 'int'> instead. If "
+             "get_array_module module wasn't called "
+             "directly, this might mean a bug in Thinc.")
+    with pytest.raises(ValueError, match=error):
+        get_array_module(0)
+    zeros = xp.zeros((1, 2))
+    xp_ = get_array_module(zeros)
+    assert hasattr(xp_, "ndarray")
```
danieldk, line 60: Don't we want something like

```suggestion
    assert xp_ == xp
```

here? (untested)

svlandeg (changes requested):
In principle this looks good to merge, though I'm slightly worried that this may introduce breaking bwd-compatibility.

I'd like us to hold off on merging before we can properly test this PR from within the spaCy test suite, which should give a good measurement of whether this is breaking or not: https://github.com/explosion/spaCy/pull/11005

The explosion bot needs some tweaks to work with the PR from this branch, which Madeesh will look into in the next week, so let's pause until that's finished & tested.

thinc/util.py
```diff
@@ -47,10 +47,17 @@ def get_torch_default_device() -> "torch.device":
...
-    if is_cupy_array(arr):
+    if is_numpy_array(arr):
+        return numpy
+    elif is_cupy_array(arr):
         return cupy
     else:
-        return numpy
+        raise ValueError(
+            "Only numpy and cupy arrays are supported"
+            f", but found {type(arr)} instead. If "
```
danieldk, line 57: I wonder if we want the `If [...] Thinc.`. This is true for a lot of exceptions in Thinc and I don't think we'd want these verbose errors.

It's up to the caller to decide whether to add more context to errors.

Also, if this fails somewhere, I am not sure how a downstream user knows whether it was or wasn't called directly.

danieldk: Ah, I missed that. Fine with me as well!

svlandeg: It would definitely be nice to have the errors in Thinc structured more properly like we do in spaCy (in a different PR though). For this particular one, I don't mind so much and am fine either way, so you get the decisive vote, Ákos!

thinc/util.py
```diff
@@ -47,10 +47,17 @@ def get_torch_default_device() -> "torch.device":
...
+            "get_array_module module wasn't called "
+            "directly, this might mean a bug in Thinc."
```
adrianeboyd, line 59:
```suggestion
            "directly, this might indicate a bug in Thinc."
```

adrianeboyd: "be"?

thinc/tests/test_util.py
```diff
@@ -39,11 +47,23 @@ def test_get_width_fail(obj):
...
-    xp = get_array_module(0)
-    assert hasattr(xp, "ndarray")
-    assert is_numpy_array(numpy.zeros((1, 2)))
-    assert not is_numpy_array((1, 2))
+@pytest.mark.parametrize("xp", ALL_XP)
+def test_array_module_cpu_gpu_helpers(xp):
+    error = ("Only numpy and cupy arrays are supported"
+             ", but found <class 'int'> instead. If "
+             "get_array_module module wasn't called "
+             "directly, this might mean a bug in Thinc.")
```
adrianeboyd, line 55:
```suggestion
             "directly, this might indicate a bug in Thinc.")
```

Again, I think we shouldn't have complete error messages copied in the tests or it's too hard to maintain.

PR #707, opened 2022-06-24
Label smooth threshold fix

kadarakos:
The maximum label-smoothing parameter allowed in `to_categorical` was `0.5`. 
However, this value was not correct. In reality the thing we need to make sure is that the value at the index where the `1` is in the one-hot vector remains the maximum after smoothing: We are looking for the largest `alpha` such that  `argmax(x) = argmax(smooth(x, alpha))`. 

In the new version the input validation dynamically adapts what is the maximum value that `label_smoothing` could take, plus it checks whether the `label_smoothing` parameter is less than `0`.

### DIFF
--- a/thinc/tests/test_util.py
+++ b/thinc/tests/test_util.py
@@ -115,16 +115,13 @@ def test_to_categorical(label_smoothing):
     ):
         to_categorical(numpy.asarray([0, 0, 0]), label_smoothing=0.01),
 
-    error_msg = ("For 5 number of classes "
-                 "label_smoothing parameter has to be less than "
-                 "0.8, but found 0.8.")
     with pytest.raises(
-        ValueError, match=error_msg
+        ValueError, match=r"label_smoothing parameter"
     ):
         to_categorical(numpy.asarray([0, 1, 2, 3, 4]), label_smoothing=0.8)
 
     with pytest.raises(
-        ValueError, match=error_msg
+        ValueError, match=r"label_smoothing parameter"
     ):
         to_categorical(numpy.asarray([0, 1, 2, 3, 4]), label_smoothing=0.88)
 


danieldk (approved):
Nice improvement! 🎉

thinc/util.py
```diff
@@ -212,16 +212,15 @@ def to_categorical(
...
-            "equal to 0.0 and less than 0.5, "
-            f"but {label_smoothing} was provided."
-        )
 
     if n_classes is None:
         n_classes = int(numpy.max(Y) + 1)  # type: ignore
 
+    if label_smoothing < 0.0:
+        raise ValueError(
+            "Label-smoothing parameter has to be greater or equal to 0"
```
svlandeg, line 221:
```suggestion
            "Label-smoothing parameter has to be greater than or equal to 0"
```

thinc/util.py
```diff
@@ -234,6 +233,13 @@ def to_categorical(
             )
         nongold_prob = label_smoothing / (n_classes - 1)
 
+    if (1 - label_smoothing) < nongold_prob:
+        raise ValueError(
+            f"For {n_classes} number of classes "
+            "label_smoothing parameter has to be less than "
+            f"{1 - nongold_prob}, but found {label_smoothing}."
+        )
+
```
svlandeg, line 242: Is this correct? The current error message implies that `label_smoothing` should be less than `1 - nongold_prob`, but `nongold_prob` is actually defined in terms of the `label_smoothing` parameter (L234). If you would run this with `label_smoothing` 0.88, wouldn't it say 
> For 5 number of classes label_smoothing parameter has to be less than 0.78, but found 0.88.

?

and if you run it with `label_smoothing` 0.8
> For 5 number of classes label_smoothing parameter has to be less than 0.8, but found 0.8.

Instead, shouldn't we check & report on the value of 
```
(n_classes - 1)/n_classes
```
so that it would consistently report
> For 5 number of classes label_smoothing parameter has to be less than 0.8, but found 0.88.

and

> For 5 number of classes label_smoothing parameter has to be less than 0.8, but found 0.8.

or am I misunderstanding this?

svlandeg: No worries. Let's add that second test case to the test suite so we can check the error message for 0.88 as well?

PR #715, opened 2022-07-04
Use config functionality from catalogue

rmitsch:
## Goals

Discard redundant config functionality, use `catalogue` instead.
Note: will fail until new `catalogue` version with config functionality is available.

## Description

Removes redundant config functionality in Thinc, including the corresponding tests.

### Types of change

Refactoring.

## Checklist

- [ ] I ran the tests, and all new and existing tests passed.


thinc/api.py
```diff
@@ -1,4 +1,5 @@
-from .config import Config, registry, ConfigValidationError
+from catalogue import Config, ConfigValidationError
+from thinc.config import registry
```
adrianeboyd, line 2:
```suggestion
from .config import registry
```

adrianeboyd:
Just a note to make sure we don't forget: this PR also needs to update `requirements.txt` and `setup.cfg`.

PR #720, opened 2022-07-07
Ops: replace FloatsType by constrained typevar

danieldk:
Ops used the `FloatsType`, which had `FloatsXd` as its bound. MyPy could not infer that code such as the following is correct,
    
```
def dish(self, X: FloatsType, inplace: bool = False) -> FloatsType:
    tmp = X * X
    # ...
```
    
because the inferred type is the union (or a subtype). If we instead constrain the type variable as follows:
    
```
FloatsType = TypeVar("FloatsType", Floats1d, Floats2d, Floats3d, Floats4d)
```
    
the type parameter will be instantiated with a single concrete type, solving such issues.

### DIFF
--- a/thinc/backends/ops.py
+++ b/thinc/backends/ops.py
@@ -1,6 +1,6 @@
 import math
 
-from typing import Optional, List, Tuple, Sequence, Union, cast, TypeVar
+from typing import Optional, List, Tuple, Sequence, Type, Union, cast, TypeVar
 from typing import Iterator, overload
 import numpy
 import itertools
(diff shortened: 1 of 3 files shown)


svlandeg:
@explosion-bot please test_slow_gpu --run-on spaCy --run-on-branch master --thinc-branch current_pr

svlandeg (approved):
This makes a lot of sense, and it's great to see all of the removed `cast`'s and `ignore`'s! 

I ran some checks on spaCy locally and through our `explosion-bot` friend, but I don't expect any issues and this should be fine to merge as soon as the spaCy test suite finishes green :-)

PR #721, opened 2022-07-07
Allow escaped section names

percevalw:
Hi ! Currently, Thinc config's system cannot export nor can it read section names that contain dots.
This PR makes the Config object able to:
- loads escaped sections
- export sections that contain dots by escaping them

For instance
```toml
[components."org.parser".v1]
param = 2
```
is now parsed as / exported from
```json
{
  "components": {
    "org.parser": {
      "v1": {
        "param": 2
      }
    }
  }
}
```

This fixes an issue with spacy (https://github.com/explosion/spaCy/issues/10954) that produces configs with dots in component section names when the factory's name contains dots.

### DIFF
--- a/thinc/config.py
+++ b/thinc/config.py
@@ -2,7 +2,7 @@ from typing import Union, Dict, Any, Optional, List, Tuple, Callable, Type, Mapp
 from typing import Iterable, Sequence, cast
 from types import GeneratorType
 from dataclasses import dataclass
-from configparser import ConfigParser, ExtendedInterpolation, MAX_INTERPOLATION_DEPTH
+from configparser import ConfigParser, ExtendedInterpolation, MAX_INTERPOLATION_DEPTH, Error
 from configparser import InterpolationMissingOptionError, InterpolationSyntaxError
 from configparser import NoSectionError, NoOptionError, InterpolationDepthError
 from configparser import ParsingError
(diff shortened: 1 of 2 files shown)


adrianeboyd:
Thanks for the PR, unfortunately I think we're going to decide instead to restrict the use of dots in factory and section names instead, similar to the restrictions for python identifiers.

percevalw:
Thanks for the feedback ! However, I think the library would benefit from some form of namespaces to (optionally) make the source of custom components explicit, and being able to use dots would be consistent with how other factories behave (models, readers, scorers, etc). 

Rather than restricting their use in factory names, would it then be conceivable to apply a `default_component_name = fn(factory_name)` transform here https://github.com/explosion/spaCy/blob/master/spacy/language.py#L641, in order to retain the ease of use of add_pipe while allowing the above-mentioned benefits ?

adrianeboyd:
The factory registry is a bit different from the other registries because the versioning is tied to the package versions rather than the names. (So it's consistently `nlp.add_pipe("parser")` across multiple versions of spacy and not `nlp.add_pipe("parser_v3_1_1")`.)

This still leaves namespacing as a concern, but there's nothing preventing multiple packages from trying to provide the same factory name and also nothing enforcing any kind of namespacing in any of the registries. My feeling is that transforming names is only going to lead to more potential for collisions between packages?

PR #726, opened 2022-07-13
More general remap_ids

kadarakos:
The `remap_ids` layer in `spaCy` is never used as far as I can tell, but I am implementing a `MultiEmbed` version of `MultiHashEmbed`. This uses the more conventional `Embed` layer instead of the `HashEmbed`. 

Before running `Embed` it runs `remap_ids` to convert `token.i` into positions in the embedding table: `chain(remap_ids(), Embed()`. 

To be able to use it as a replacement for `MultiHashEmbed` a couple modifications had to be made. `HashEmbed` takes a `column` attribute, because in `spaCy` the `FeatureExtractor` returns an `Ints2d` which is number of tokens x number of features. 
To be able to run the same idea with `Embed` the `remap_ids` now also implements `column`: `chain(remap_ids(column=column), Embed(column=0))`.

Furthermore in the common case where the `mapping_table` is a `Dict[int, int]` there was a bit of a strange error: whereas with `numpy` this works `{3: 'q'}[np.array([3])[0]]` with cupy we get:

```
----> 1 {3: 'q'}[cp.array([3])[0]]

TypeError: unhashable type: 'cupy._core.core.ndarray'
```
In the new version when we detect a `cupy` integer array we cast the elements to `int`. 

Finally, the `backprop` used to return a `[]`, but when the input is an `xp array` then this can lead to confusing errors e.g.: if the `remap_ids` is within a `with_array` block as is the case with `Embed`. Now it returns an empty array instead.


thinc/layers/remap_ids.py
```diff
@@ -31,11 +41,23 @@ def forward(
...
     dtype = model.attrs["dtype"]
+    column = model.attrs["column"]
+    if column is not None:
+        inputs = cast(Array2d, inputs)
+        inputs = inputs[:, column]
+    # elements of cupy arrays are 0-dimensional arrays
+    # not the integers stored in the original mapper.
+    if is_cupy_array(inputs):  # type: ignore
+        if inputs.dtype.kind == 'i':  # type: ignore
+            inputs = map(int, inputs)  # type: ignore
```
danieldk, line 52: Doesn't this retrieve the elements from the GPU one by one? Convert everything all at once as a NumPy array.

ljvmiranda921 (dismissed):

thinc/layers/remap_ids.py
```diff
@@ -31,11 +41,22 @@ def forward(
...
     default = model.attrs["default"]
     dtype = model.attrs["dtype"]
+    column = model.attrs["column"]
+    if column is not None:
+        inputs = cast(Ints2d, inputs)
+        inputs = inputs[:, column]
+    # elements of cupy arrays are 0-dimensional arrays
+    # not the integers stored in the original mapper.
+    if is_cupy_array(inputs):  # type: ignore
+        inputs = to_numpy(inputs)  # type: ignore
```
danieldk, line 51:
```suggestion
    inputs = to_numpy(inputs)  # type: ignore
```

`to_numpy` already checks if the array is a CuPy array, so now we'd be checking it twice in case it is. So, I think it's better to just call `to_numpy_ unconditionally.

I also wonder if we can avoid the `type: ignore` here, or if not make it a more specific ignore.

danieldk (commented):
One more small nit. I think it'd also be good if @adrianeboyd can look at this PR, since she is much more familiar with the embedding handling.

adrianeboyd:
Why isn't `column` implemented in the model in the same way as in `HashEmbed`?

kadarakos:
> Why isn't `column` implemented in the model in the same way as in `HashEmbed`?

How is it different?

adrianeboyd:
Both `HashEmbed` and the PR description look more like this:

https://github.com/explosion/thinc/blob/40c129f6ee51d4060f1af6b29f903ce51fd89736/thinc/layers/hashembed.py#L48-L53

It's not that I think the output here are incorrect, but I'm trying to understand why it's implemented differently.

kadarakos:
> Both `HashEmbed` and the PR description look more like this:
> 
> https://github.com/explosion/thinc/blob/40c129f6ee51d4060f1af6b29f903ce51fd89736/thinc/layers/hashembed.py#L48-L53
> 
> It's not that I think the output here is incorrect, but I'm trying to understand why it's implemented differently.



> Both `HashEmbed` and the PR description look more like this:
> 
> https://github.com/explosion/thinc/blob/40c129f6ee51d4060f1af6b29f903ce51fd89736/thinc/layers/hashembed.py#L48-L53
> 
> It's not that I think the output here is incorrect, but I'm trying to understand why it's implemented differently.

The

kadarakos:
> Both `HashEmbed` and the PR description look more like this:
> 
> https://github.com/explosion/thinc/blob/40c129f6ee51d4060f1af6b29f903ce51fd89736/thinc/layers/hashembed.py#L48-L53
> 
> It's not that I think the output here is incorrect, but I'm trying to understand why it's implemented differently.

In the `HashEmbed` the `init` creates a new model that's as the comment says
```
 This is equivalent to array[:, column].
```
and I think this is convenient, because the backprop now works properly due to the `chain` combinator combining column selection with the original embedding model. In `remap_ids` we don't need `backprop` so we can just do the simpler `array[:, column].`

thinc/tests/layers/test_layers_api.py
```diff
@@ -8,7 +8,7 @@
 from thinc.compat import has_torch
 import numpy
 import pytest
-
+import cupy
```
adrianeboyd, line 11: You can't unconditionally import cupy anywhere in thinc even in the tests. Maybe this was a mistake, what were you trying to do?

kadarakos: mistake

adrianeboyd:
Ah, I see what you mean. I was mainly thinking that implementing this the same way would avoid a lot of the casting and numpy/cupy problems.

In general I'm not sure this is going to work (easily) for `Hashable` due to the types? It looks like `util.to_numpy` converts everything to strings for your test case, and I think there may be cases where it would convert ints to floats for numeric input?

I wonder if this would be easier if it becomes `remap_ids.v2` and is restricted to `IntsXd` instead?

kadarakos:
> Ah, I see what you mean. I was mainly thinking that implementing this the same way would avoid a lot of the casting and numpy/cupy problems.
> 
> In general I'm not sure this is going to work (easily) for `Hashable` due to the types? It looks like `util.to_numpy` converts everything to strings for your test case, and I think there may be cases where it would convert ints to floats for numeric input?
> 
> I wonder if this would be easier if it becomes `remap_ids.v2` and is restricted to `IntsXd` instead?

Hmm I guess you are right. But I'm thinking that having `Ints1d` and `Ints2d` is basically what we would need from the point of view of `spaCy`, because we use the integer attributes there and not the strings themselves. But in general I think people would like to use `remap_ids` with `str2int` maps more likely (provided anyone would use this outside of `spaCy`). 

What do you think about implementing `remap_ids` in such a way that it only supports `Union[Sequence[str], Sequence[int], Ints1d, Ints2d]` and that's it?

adrianeboyd:
I think that sounds reasonable.

kadarakos:
> I think that sounds reasonable.

Added the new version.

thinc/layers/remap_ids.py
```diff
@@ -1,41 +1,99 @@
...
-    values = [table.get(x, default) for x in inputs]
-    arr = model.ops.asarray2i(values, dtype=dtype)
+    column = model.attrs["column"]
+    if is_xp_array(inputs):
+        xp_inputs = cast(IntVecOrMat, inputs)
+        xp_input = True
+        _check_1d_or_2d_str_or_int(xp_inputs)
+        if xp_inputs.ndim == 2:
+            _check_column(xp_inputs, column)
+            idx = to_numpy(xp_inputs[:, column])
```
adrianeboyd, line 83: This seems like way too much slow validation for `forward` for things that types are intended to handle, or for things that can just fail if the settings are invalid.

I don't understand why you're checking for `"U"` here at all, since you've just cast to this `Ints1/2d` and a numpy array of strings isn't `Sequence["str"]`, either, I don't think?

thinc/layers/remap_ids.py
```diff
@@ -1,41 +1,99 @@
...
 from ..config import registry
-from ..types import Ints2d, DTypes
+from ..types import Ints1d, Ints2d, ArrayXd
+from ..util import is_xp_array, to_numpy
 
 
-InT = Sequence[Any]
+InT = Union[Sequence[str], Sequence[int], Ints1d, Ints2d]
 OutT = Ints2d
+IntVecOrMat = Union[Ints1d, Ints2d]
```
adrianeboyd, line 12: To follow our usual naming conventions, something more like:

```suggestion
Ints1dOr2d = Union[Ints1d, Ints2d]
```

thinc/layers/remap_ids.py
kadarakos: Which check is unnecessary? I guess the `U` or `i` is unnecessary, but if `ndim > 2` is probably an unintended mistake by the user same with the `column` being unset or it being too large.

adrianeboyd: The input is already typed as:

```
InT = Union[Sequence[str], Sequence[int], Ints1d, Ints2d]
```

I think the `column` can just fail if it's incorrect.

thinc/layers/remap_ids.py
```diff
@@ -1,41 +1,99 @@
...
     default = model.attrs["default"]
-    dtype = model.attrs["dtype"]
-    values = [table.get(x, default) for x in inputs]
-    arr = model.ops.asarray2i(values, dtype=dtype)
+    column = model.attrs["column"]
+    if is_xp_array(inputs):
+        xp_inputs = cast(IntVecOrMat, inputs)
+        xp_input = True
+        _check_1d_or_2d_str_or_int(xp_inputs)
+        if xp_inputs.ndim == 2:
```
adrianeboyd, line 81:
```suggestion
        if column is not None:
```

thinc/layers/remap_ids.py
```diff
@@ -1,41 +1,82 @@
...
+InT = Union[Sequence[str], Sequence[int], Ints1d, Ints2d]
 OutT = Ints2d
+Ints1dOr2d = Union[Ints1d, Ints2d]
+
+
+def _check_1d_or_2d(arr: Ints1dOr2d) -> None:
+    if arr.ndim > 2:
+        raise ValueError(
+            "Inputs array can be only one or two dimensional."
+        )
```
adrianeboyd, line 19: I still really don't think this is necessary?

thinc/layers/remap_ids.py
```diff
@@ -1,41 +1,82 @@
...
 
 from ..model import Model
 from ..config import registry
-from ..types import Ints2d, DTypes
+from ..types import Ints1d, Ints2d, ArrayXd
+from ..util import is_xp_array, to_numpy
 
 
-InT = Sequence[Any]
+InT = Union[Sequence[str], Sequence[int], Ints1d, Ints2d]
```
adrianeboyd, line 10: And with the changes now I think this could actually go back to `Sequence[Hashable]`?

thinc/layers/remap_ids.py
```diff
@@ -1,41 +1,76 @@
...
-    values = [table.get(x, default) for x in inputs]
-    arr = model.ops.asarray2i(values, dtype=dtype)
+    column = model.attrs["column"]
+    if is_xp_array(inputs):
+        xp_inputs = cast(Ints1dOr2d, inputs)
+        xp_input = True
+        if column is not None:
+            xp_inputs = cast(Ints2d, xp_inputs)
+            idx = to_numpy(xp_inputs[:, column])
+        else:
```
adrianeboyd, line 61: And then you can avoid the custom type, I think?

```suggestion
        xp_input = True
        if column is not None:
            xp_inputs = cast(Ints2d, xp_inputs)
            idx = to_numpy(xp_inputs[:, column])
        else:
            xp_inputs = cast(Ints1d, xp_inputs)
```

thinc/layers/remap_ids.py
```diff
@@ -1,41 +1,66 @@
...
 from ..config import registry
-from ..types import Ints2d, DTypes
+from ..types import Ints1d, Ints2d
+from ..util import is_xp_array, to_numpy
 
 
-InT = Sequence[Any]
+InT = Union[Sequence[Hashable], Ints1d, Ints2d]
 OutT = Ints2d
+Ints1dOr2d = Union[Ints1d, Ints2d]
```
adrianeboyd, line 12:
```suggestion
```

thinc/layers/remap_ids.py
```diff
@@ -1,41 +1,66 @@
...
     default = model.attrs["default"]
-    dtype = model.attrs["dtype"]
-    values = [table.get(x, default) for x in inputs]
-    arr = model.ops.asarray2i(values, dtype=dtype)
+    column = model.attrs["column"]
+    if is_xp_array(inputs):
+        xp_input = True
+        if column is not None:
+            xp_inputs = cast(Ints2d, inputs)
+            idx = to_numpy(xp_inputs[:, column])
```
adrianeboyd, line 50: And one last small thing, I think that if we can avoid having the similar variable names `xp_input` and `xp_inputs`, it will be less confusing for someone looking at this in the future:

```suggestion
            idx = to_numpy(cast(Ints2d, inputs)[:, column])
```

kadarakos:
> This version looks good! It does change the parameters and types enough that I think it should be `remap_ids.v2`, though. I think we can just have both in parallel here?

I'm not sure what to have for `.v1`. I think that changing 

```
    def backprop(dY: OutT) -> InT:
        return []
```

to 

```
    def backprop(dY: OutT) -> InT:
        if xp_input:
            return model.ops.xp.empty(dY.shape)
        else:
            return []
```
is a bugfix that prevents `remap_ids` to break when its in the `with_array` context.

Same with `idx = to_numpy(inputs)`, because `values = [table.get(x, default) for x in idx]` doesn't actually work with `cupy` integer arrays.

adrianeboyd:
I agree that the existing version is wonky, but the main issue is the parameters, since this removes `dtype` and adds `column`. If you hadn't removed the existing test it would break on `dtype`, I think?

kadarakos:
> I agree that the existing version is wonky, but the main issue is the parameters, since this removes `dtype` and adds `column`. If you hadn't removed the existing test it would break on `dtype`, I think?

That's true, but I'm also not sure why the `dtype` was added. The original typing was:

```
InT = Sequence[Any]
OutT = Ints2d
```

so like we could give `dtype` float, but then this would return something not use able:

```
    arr = model.ops.asarray2i(values, dtype=dtype)
    output = model.ops.reshape2i(arr, -1, 1)
```

Not sure why we need to pass a `dtype` to `asarray2i` 

```
   def asarray2i(
        self, data: Union[Ints2d, Sequence[int]], *, dtype: Optional[DTypes] = "int32"
    ) -> Ints2d:
        return cast(Ints2d, self.asarray(data, dtype=dtype))
```

kadarakos:
So yeah I guess I find it a bit hard to understand what should I include in `.v1`

adrianeboyd:
I don't think you need to spend much effort trying to clean up `.v1`. It can be buggy/broken and we wouldn't touch it again.

For now I would propose keeping a copy of `.v1` unmodified and just adding the new one as `.v2`.

kadarakos:
Ok I've added a legacy version in `remap_ids.py` is this what you've meant?

thinc/layers/remap_ids.py
```diff
@@ -1,32 +1,37 @@
...
-from ..types import Ints2d, DTypes
+from ..types import Ints1d, Ints2d, DTypes
+from ..util import is_xp_array, to_numpy
 
 
-InT = Sequence[Any]
+InT = Union[Sequence[Hashable], Ints1d, Ints2d]
 OutT = Ints2d
 
+InT_legacy = Sequence[Any]
```
adrianeboyd, line 13: `_v1` instead of `_legacy`?

svlandeg: (same comment for some of the other names like `forward_legacy`)

thinc/tests/layers/test_layers_api.py
```diff
@@ -128,7 +128,8 @@ def assert_data_match(Y, out_data):
     # ("CauchySimilarity.v1", {}, (array2d, array2d), array1d),
     ("ParametricAttention.v1", {}, ragged, ragged),
     ("SparseLinear.v1", {}, (numpy.asarray([1, 2, 3], dtype="uint64"), array1d, numpy.asarray([1, 1], dtype="i")), array2d),
-    ("remap_ids.v1", {"dtype": "f"}, ["a", 1, 5.0], array2dint)
+    ("remap_ids.v1", {"mapping_table": {}, "dtype": "int64"}, ["a", 1, 5.0], array2dint),
```
svlandeg, line 131: I'd be a fan of keeping the existing test as is, then add new/additional ones as well.

adrianeboyd:
Also please update the corresponding docs: https://thinc.ai/docs/api-layers#remap_ids

website/docs/api-layers.md
```diff
@@ -1267,21 +1267,25 @@ https://github.com/explosion/thinc/blob/master/thinc/layers/padded2list.py
...
 </inline-list>
 
-Remap string or integer inputs using a mapping table, usually as a preprocess
-before embeddings. The mapping table can be passed in on input, or updated after
-the layer has been created. The mapping table is stored in the `"mapping_table"`
-attribute.
+Remap a sequence of strings, integers or other Hashable inputs using a
+mapping table, usually as a preprocessing step before embeddings. The
+input can also be a two dimensional integer array in which case the
+`column: int` attribute tells the `remap_ids` layer which column of the
```
adrianeboyd, line 1278:
```suggestion
`column` attribute tells the `remap_ids` layer which column of the
```

website/docs/api-layers.md
```diff
@@ -1267,21 +1267,25 @@ https://github.com/explosion/thinc/blob/master/thinc/layers/padded2list.py
...
+array to map. For 2D inputs the `column` has to be set on initialization.
+The mapping table can be set at initialization, but can also be passed in on input, 
+When provided on initializtion the mapping table is stored in the `"mapping_table"`
+attribute.
```
adrianeboyd, line 1282: Aside from the spelling error ("initializtion"), I don't understand the point that you're trying to make here? It's always stored in this attribute, right?

website/docs/api-layers.md
```diff
@@ -1267,21 +1267,25 @@ https://github.com/explosion/thinc/blob/master/thinc/layers/padded2list.py
...
-Remap string or integer inputs using a mapping table, usually as a preprocess
-before embeddings. The mapping table can be passed in on input, or updated after
-the layer has been created. The mapping table is stored in the `"mapping_table"`
-attribute.
+Remap a sequence of strings, integers or other Hashable inputs using a
+mapping table, usually as a preprocessing step before embeddings. The
+input can also be a two dimensional integer array in which case the
+`column: int` attribute tells the `remap_ids` layer which column of the
+array to map. For 2D inputs the `column` has to be set on initialization.
+The mapping table can be set at initialization, but can also be passed in on input,
```
adrianeboyd, line 1280: I don't think it can be passed in on input (or I don't know what you mean by "passed in on input")?

Something more like "can also be set or modified at a later point by setting the `mapping_table` attribute."

website/docs/api-layers.md
```diff
@@ -1267,21 +1267,25 @@ https://github.com/explosion/thinc/blob/master/thinc/layers/padded2list.py
...
+- **Input:** <tt>Union[Sequence[Hashable], Ints1d, Ints2d]</tt>
 - **Output:** <ndarray>Ints2d</ndarray>
 
 </inline-list>
 
-Remap string or integer inputs using a mapping table, usually as a preprocess
-before embeddings. The mapping table can be passed in on input, or updated after
-the layer has been created. The mapping table is stored in the `"mapping_table"`
-attribute.
+Remap a sequence of strings, integers or other Hashable inputs using a
```
adrianeboyd, line 1275:
```suggestion
Remap a sequence of strings, integers or other hashable inputs using a
```

website/docs/api-layers.md
```diff
@@ -1267,21 +1267,25 @@ https://github.com/explosion/thinc/blob/master/thinc/layers/padded2list.py
...
+mapping table, usually as a preprocessing step before embeddings. The
+input can also be a two dimensional integer array in which case the
+`column: int` attribute tells the `remap_ids` layer which column of the
+array to map. For 2D inputs the `column` has to be set on initialization.
```
adrianeboyd, line 1279: Technically you can modify `column` later just like `mapping_table`, right?

website/docs/api-layers.md
kadarakos: Oh right that's true. But it can also be passed later, like its not going to throw an error basically if you want to switch it later on. Do you have some suggestion about how to say that?

adrianeboyd: I think it would be clearer to mention that `column` and `mapping_table` can be modified after initialization. Something about how it's set up to be flexible in terms of how the mapping is configured so that the values matter at the point where you call `forward` and not in `init` (but I wouldn't phrase it exactly like that).

website/docs/api-layers.md
kadarakos: Ok I tried to rework the explanation that reflects that technically both `column` and `mapping_table` can be modified any time before calling `forward`.

website/docs/api-layers.md
```diff
@@ -1267,21 +1267,28 @@ https://github.com/explosion/thinc/blob/master/thinc/layers/padded2list.py
...
 
-Remap string or integer inputs using a mapping table, usually as a preprocess
-before embeddings. The mapping table can be passed in on input, or updated after
-the layer has been created. The mapping table is stored in the `"mapping_table"`
-attribute.
+Remap a sequence of strings, integers or other hashable inputs using a
+mapping table, usually as a preprocessing step before embeddings. The
+input can also be a two dimensional integer array in which case the
+`column: int` attribute tells the `remap_ids` layer which column of the
+array to map with the `mapping_table: Dict[Any, int]`.
```
adrianeboyd, line 1279: The types are in the table, let's not add them here.
```suggestion
array to map with the `mapping_table`.
```

Or even just plain "column" and "mapping table" (no quotes).

website/docs/api-layers.md
```diff
@@ -1267,21 +1267,28 @@ https://github.com/explosion/thinc/blob/master/thinc/layers/padded2list.py
...
+Both attributes can be passed during initialization, but since during
+`forward` they are retrieved from `model.attrs`, they can be set
+any time before calling `forward`. This means that they can also be
+changed between calls. Before calling `forward` the `mapping_table`
```
adrianeboyd, line 1283:
```suggestion
modified between calls. Before calling `forward` the `mapping_table` 
```

adrianeboyd:
As a final step can you format the docs?

website/docs/api-layers.md
```diff
@@ -1267,21 +1267,26 @@ https://github.com/explosion/thinc/blob/master/thinc/layers/padded2list.py
...
 | --------------- | ------------------------------------- | ------------------------------------------------------------------------------------------------------------ |
 | `mapping_table` | <tt>Dict[Any, int]</tt>               | The mapping table to use. Can also be set after initialization by writing to `model.attrs["mapping_table"]`. |
 | `default`       | <tt>int</tt>                          | The default value if the input does not have an entry in the mapping table.                                  |
-| `dtype`         | <tt>DTypes</tt>                       | The data type of the array.                                                                                  |
+| `column`        | <tt>int</tt>                          | The column to apply the mapper to in case of 2D input.                                                       |
 | **RETURNS**     | <tt>Model[Sequence[Any], Ints2d]</tt> | The layer to compute the transformation.                                                                     |
```
adrianeboyd, line 1290: Ah, the model type here didn't get updated.

adrianeboyd:
And reformat again!

adrianeboyd:
Let me do a last-minute renaming to have new `_v2` instead of renamed `_v1`.

PR #729, opened 2022-07-14
Multi gpu

### LINKED ISSUE #713: Model.to_gpu is not usable
I am attempting to assign individual layers to separate GPUs in order to conserve memory.  However, the Model.to_gpu function takes an all or nothing approach which prevents this from working.

While diagnosing the origin of memory access error during training, (`cupy_backends.cuda.api.driver.CUDADriverError: CUDA_ERROR_ILLEGAL_ADDRESS: an illegal memory access was encountered`), I noticed that [`CupyOps.device_id`](https://github.com/explosion/thinc/blob/master/thinc/backends/cupy_ops.py#L23) is never used
[or set](https://github.com/explosion/thinc/blob/c7b0d6759645babe94315a36c84d56ec877252f2/thinc/model.py#L513).

Ideally, all the CupyOps would run inside a `cp.cuda.Device(device_id)` context, but that is not the case.  Instead, the `xp` attribute is (ab)used in many places.  That will try and run everything through GPU 0, so errors won't appear until something was moved to another GPU.

Two other difficulties are the initialization step, which doesn't declare memory in the right places,
and the `finish_update` step, where the optimizer does arithmetic on parameters outside of a context.

frobnitzem:
This PR adds support for the `CupyOps.device_id` by creating an `Ops.context()` interface that wraps a [cupy.cuda.Device](https://docs.cupy.dev/en/stable/reference/generated/cupy.cuda.Device.html).

It was tested with a multi-GPU chain model here:
https://code.ornl.gov/99R/autoencoder.

The above test works with the current thinc after one change to CupyOps.asarray -- forcing `return self.xp.asarray(data, dtype)` for cupy inputs.

After merging this PR, the above example simplifies to: https://code.ornl.gov/99R/autoencoder/-/tree/multi_gpu_thinc.

The example outputs for a 3-GPU run are:
```
# multi_gpu branches, num_gpu = 3, size = 7000
$ jsrun -n1 -g3 -c21 -bpacked:21 python3 autoencoder.py
Initialized model with input dimension nI=7000 and output dimension nO=7000
0: iteration time = 3.619584083557129 mem = 697 M / gpu 0
1: iteration time = 3.61960506439209 mem = 969 M / gpu 1
2: iteration time = 3.6196165084838867 mem = 647 M / gpu 2
0: iteration time = 0.3242201805114746 mem = 0 M / gpu 0
1: iteration time = 0.32422947883605957 mem = 0 M / gpu 1
2: iteration time = 0.3242321014404297 mem = 0 M / gpu 2
0: iteration time = 0.31447744369506836 mem = 0 M / gpu 0
1: iteration time = 0.31448888778686523 mem = 0 M / gpu 1
2: iteration time = 0.3144950866699219 mem = 0 M / gpu 2
0: iteration time = 0.31055688858032227 mem = 0 M / gpu 0
1: iteration time = 0.31056880950927734 mem = 0 M / gpu 1
2: iteration time = 0.310575008392334 mem = 0 M / gpu 2
0: main time = 9.183293342590332 mem = 961 M / gpu 0
1: main time = 9.183251857757568 mem = 1211 M / gpu 1
2: main time = 9.183244705200195 mem = 808 M / gpu 2
```

This PR uses encapsulation of cupy array math in these contexts to remove warnings (encountered during testing) about non-cupy.cuda.Device()-context bound accesses:
```
thinc/thinc/backends/cupy_ops.py:256: PerformanceWarning: The device where the array resides (1) is different from the current device (0). Peer access has been activated automatically.
  gradient *= cupy.minimum(threshold, grad_norm) / grad_norm
thinc/thinc/backends/cupy_ops.py:343: PerformanceWarning: The device where the array resides (1) is different from the current device (0). Peer access has been activated automatically.
  adam_kernel(
...
```

closes #713

### DIFF
--- a/examples/benchmarks/lstm_tagger.py
+++ b/examples/benchmarks/lstm_tagger.py
@@ -112,7 +112,7 @@ def set_backend(name, gpu_id):
         if gpu_id == -1:
             set_current_ops(NumpyOps(use_blis=True))
         else:
-            set_current_ops(CupyOps())
+            set_current_ops(CupyOps(device_id=gpu_id))
         if name == "pytorch":
             import torch
 
--- a/thinc/backends/__init__.py
+++ b/thinc/backends/__init__.py
@@ -112,7 +112,7 @@ def get_ops(name: str, **kwargs) -> Ops:
 def get_array_ops(arr):
     """Return CupyOps for a cupy array, NumpyOps otherwise."""
     if is_cupy_array(arr):
-        return CupyOps()
+        return CupyOps(device_id=arr.device)
     else:
         return NumpyOps()
 
--- a/thinc/backends/cupy_ops.py
+++ b/thinc/backends/cupy_ops.py
@@ -1,6 +1,9 @@
+from contextlib import contextmanager
+from typing import Optional
+
 import numpy
 from .. import registry
-from .ops import Ops
+from .ops import Ops, contextmethod
 from .numpy_ops import NumpyOps
 from . import _custom_kernels
 from ..types import DeviceTypes
@@ -17,10 +20,27 @@ class CupyOps(Ops):
     _xp2 = cupyx
 
     def __init__(
-        self, device_type: DeviceTypes = "gpu", device_id: int = 0, **kwargs
+        self,
+        device_type: DeviceTypes = "gpu",
+        device_id: Optional[int] = None,
+        **kwargs
     ) -> None:
         self.device_type = device_type
+        if device_id is None:
+            device_id = self.xp.cuda.runtime.getDevice()
         self.device_id = device_id
+
+    @contextmanager
+    def context(self):
+        """Create a context placing Cupy operations on a specific GPU.
+        used as:
+
+        >>> ops = CupyOps(device_id=1)
+        >>> with ops.context():
+        >>>     x = ops.xp.zeros((4,10)) + ops.xp.arange(10)
+        """
+        with self.xp.cuda.Device(self.device_id) as ctxt:
+            yield ctxt
 
     def to_numpy(self, data, *, byte_order=None):
         if not isinstance(data, numpy.ndarray):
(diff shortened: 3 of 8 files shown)


honnibal:
Thanks for looking at this!

While I understand that multi-GPU models are necessary for some workloads, it's not trivial to make sure that we come out ahead of a single GPU in terms of performance. And in the meantime changes like this can add overhead to both the code and execution that impacts the single-GPU case, which is what almost everyone will be using. 

We therefore want to think carefully and do our own experiments before we take on this topic. We'd likely target data-parallel multi-GPU workloads rather than model parallelism, as that's the most common requirement people will have for the main usage of Thinc, which is models for spaCy.

Unfortunately that means this isn't really a good target for user contribution. It's a change that will have significant impact for the library, so we can only get to this when we're really ready to work on it.

frobnitzem:
I expect the performance impact to come primarily from python function call overhead when encountering context `__enter__` and `__exit__` calls.  Those happen once for each ops primitive.

For CPU, the context itself does nothing.  For GPU, both methods end up calling [cudaSetDevice](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g159587909ffa0791bbe4b40187a4c6bb).  It is fairly common to see codes that call this on every blas operation (e.g. page 2 of [SLATE port to AMD and Intel platforms](https://www.icl.utk.edu/files/publications/2021/icl-utk-1479-2021.pdf)).  From NVIDIA's docs, it is described as:

> This function will do no synchronization with the previous or new device, and should be considered a very low overhead call.

The [HIP docs](https://rocmdocs.amd.com/en/latest/ROCm_API_References/HIP_API/Device-management.html#hipsetdevice) go further,

> This function does no synchronization with the previous or new device, and has very little runtime overhead. Applications can use hipSetDevice to quickly switch the default device before making a HIP runtime call which uses the default device.
> The default device is stored in thread-local-storage for each thread. Thread-pool implementations may inherit the default device of the previous thread. A good practice is to always call hipSetDevice at the start of HIP coding sequency to establish a known standard device.

Note also that the change to `CupyOps.asarray` has no overhead, since both [cupy.asarray](https://docs.cupy.dev/en/stable/reference/generated/cupy.asarray.html) and [numpy.asarray](https://numpy.org/doc/stable/reference/generated/numpy.asarray.html#numpy.asarray) state that no copy is performed if the array already matches the call's request.

svlandeg:
Considering @honnibal's earlier reply, I'll go ahead and close this particular PR, but we'll add the idea to our backlog of things to explore further. Thanks again!

PR #738, opened 2022-08-05
Exclude `blis v0.9.0` due to buggy Haswell kernels

shadeMe:


adrianeboyd:
It's not due to the buggy Haswell kernels (although they are buggy) but due to the changes from https://github.com/explosion/cython-blis/pull/70, but now I see that I was confused by the release notes (that I wrote) and this was in v0.9.0, so we don't need this PR, sorry!

PR #742, opened 2022-08-10
Remove vestigial/mostly unused `backends.linalg` module

shadeMe:
The `backends.linalg` module was [introduced back in 2016](https://github.com/explosion/thinc/commits/b41e68d912f083674558dc0bc16592b295cb3fa5/thinc/linalg.pyx?browsing_rename_history=true&new_path=thinc%2Fbackends%2Flinalg.pyx&original_branch=master), presumably as a BLAS-specific implementation of linear algebra functions that could be used in the context of thinc ops. In the current version of thinc, however, this module is almost entirely unused (and untested), where calls to its functions have been replaced by their more performant equivalents in `numpy`/`cupy`. 

This PR replaces the last remaining references to the module's functions with `numpy` and `BLAS` calls and removes it from the project.

Outside of thinc, this module was only used in the transition-based paser code in spaCy. [This spaCy PR](https://github.com/explosion/spaCy/pull/11292) replaces those references in a similar fashion.


thinc/backends/cblas.pxd
```diff
@@ -33,6 +34,8 @@ cdef class CBlas:
 cdef daxpy_ptr daxpy(CBlas cblas) nogil
 cdef saxpy_ptr saxpy(CBlas cblas) nogil
 cdef sgemm_ptr sgemm(CBlas cblas) nogil
+cdef sscalv_ptr sscalv(CBlas cblas) nogil
```
danieldk, line 37: I think this should be `sscal` right? If I am not mistaken `sscalv` is a BLIS symbol/name.

https://www.intel.com/content/www/us/en/develop/documentation/onemkl-developer-reference-c/top/blas-and-sparse-blas-routines/blas-routines/blas-level-1-routines-and-functions/cblas-scal.html

While at it, we should probably also add `dscal`.

thinc/backends/numpy_ops.pyx
```diff
@@ -523,9 +523,10 @@ def check_seq2col_lengths(ops, lengths, B):
 
 
 def cpu_clip_gradient(weight_t[::1] gradient, weight_t threshold):
-    grad_norm = Vec.norm(&gradient[0], gradient.shape[0])
+    arr_view = numpy.asarray(<np.float32_t[:gradient.shape[0]]> &gradient[0])
+    grad_norm = numpy.linalg.norm(arr_view)
     if grad_norm >= threshold:
-        Vec.mul_i(&gradient[0], threshold / grad_norm, gradient.shape[0])
+        arr_view *= threshold / grad_norm
```
danieldk, line 529: Probably nicer to use `sscal` here?

thinc/backends/numpy_ops.pyx
```diff
@@ -523,9 +523,10 @@ def check_seq2col_lengths(ops, lengths, B):
 
 
 def cpu_clip_gradient(weight_t[::1] gradient, weight_t threshold):
-    grad_norm = Vec.norm(&gradient[0], gradient.shape[0])
+    arr_view = numpy.asarray(<np.float32_t[:gradient.shape[0]]> &gradient[0])
+    grad_norm = numpy.linalg.norm(arr_view)
```
danieldk, line 527: Can we use BLAS `snrm2` here to avoid rewrapping the array using numpy? Maybe that'd also allow making it `cdef nogil`?

thinc/backends/cblas.pxd
shadeMe: Ah, yes. Good catch! Will fix/add.

thinc/backends/numpy_ops.pyx
shadeMe: Actually, that entire function is unused/vestigial and can/should be removed in `v9.0`. The default `Ops` class handles the gradient clipping in its [`clip_gradient` method](https://github.com/explosion/thinc/blob/eda4c75771667aee114e640b9c1e4021a514a275/thinc/backends/ops.py#L1158). `CupyOps` overrides this, but `NumpyOps` doesn't.

thinc/backends/numpy_ops.pyx
shadeMe: Same as above/below.

shadeMe:

shadeMe:

danieldk (approved):
LGTM. Could you add a ticket that we should update `thinc-apple-ops` to add the `?scal` functions when we release v9?

shadeMe:
Seems like the [test failure](https://dev.azure.com/explosion-ai/Public/_build/results?buildId=20143&view=logs&j=e422489b-dc1d-5d7a-c1a5-f174ea0b4cd2&t=e9d3726d-ebaa-576c-7797-a136b0335dd4&l=14) on Windows 3.10 is spurious?

PR #746, opened 2022-08-24
`PyTorchGradScaler`: Cache `_found_inf` on the CPU

shadeMe:
This prevents unnecessary overhead from launching kernels on the GPU in hot backward passes, resulting in sizable savings in training time when training `torch` models.

While training the experimental `CoreferenceClusterer` component with a `Transformer`, this resulted in a 25% reduction in execution time of the `Transformer`'s `finish_update`/backprop pipeline phase.

### DIFF
--- a/thinc/shims/pytorch_grad_scaler.py
+++ b/thinc/shims/pytorch_grad_scaler.py
@@ -51,14 +51,12 @@ class PyTorchGradScaler:
         self._backoff_factor = backoff_factor
         self._growth_interval = growth_interval
 
-        self._found_inf = torch.full((1,), 0.0)
-        # Caches `self._found_inf` to minimize DTH transfers in hotloops.
-        self._found_inf_bool = False
         self._growth_tracker = torch.full((1,), 0, dtype=torch.int)
         self._scale = torch.full((1,), init_scale)
+        # Pin to the CPU to minimize DTH transfers in hot loops.
+        self._found_inf = torch.full((1,), 0.0, device="cpu")
 
     def to_(self, device):
-        self._found_inf = self._found_inf.to(device)
         self._growth_tracker = self._growth_tracker.to(device)
         self._scale = self._scale.to(device)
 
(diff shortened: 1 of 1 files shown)


shadeMe:

danieldk:
> This prevents unnecessary overhead from launching kernels on the GPU in hot backward passes, resulting in sizable savings in training time when training `torch` models.

Nice find! But it's interesting that this would be so expensive compared to the parameter updates themselves. Are there lots of small parameters?

Besides that, I think we could just always store `find_inf` on the CPU? Then we don't need an extra member variable. I think tracking the same state in two members is redundant and a bit error-prone.

shadeMe:
> Nice find! But it's interesting that this would be so expensive compared to the parameter updates themselves. Are there lots of small parameters in that model?

There were around 200 `torch.Parameter`s from just the transformer (`roberta-base`) encoder and another 10 odd from the co-ref module. The time in `finish_update` was roughly spent in a 60-40 split between the updates and the call to `found_inf`.

I think the main reason it even turned up in the profile was due to lack of batching in the pipeline. So, we were essentially performing updates on a per-example basis, which ended up exposing this pathological behaviour.

> Besides that, I think we could just always store find_inf on the CPU? Then we don't need an extra member variable. I think tracking the same state in two members is redundant and a bit error-prone.

I think the device conversions came from when we'd always allocate everything as CUDA tensors (before Thinc 8.1).

Yeah, that was my first thought too but the device conversions gave me pause. But it does look like we always have to copy the per-device tensors to `self.found_inf_`'s device anyway, so we can indeed pin it to the CPU. I'll change that.

danieldk:
> I think the main reason it even turned up in the profile was due to lack of batching in the pipeline. So, we were essentially performing updates on a per-example basis, which ended up exposing this pathological behaviour.

Ah, yes, that will show things that are otherwise cheap. Still surprised though that reading a single value from GPU is so visible besides updating e.g. 768x768 matrices from the transformer using the optimizer. But I guess those parameter updates are also dominated by kernel starts, etc. due to their small sizes without batching.

shadeMe:

danieldk (approved):
Looks good! The PyTorch GPU sweep is also happy.

PR #750, opened 2022-08-31
Relax type definition for AxisIndex

willfrey:
Change `int` to `SupportsIndex`.

### DIFF
--- a/thinc/layers/array_getitem.py
+++ b/thinc/layers/array_getitem.py
@@ -1,7 +1,12 @@
-from typing import Union, Sequence, SupportsIndex, Tuple, TypeVar
+import sys
+from typing import Union, Sequence, Tuple, TypeVar
 from ..types import ArrayXd, FloatsXd, IntsXd
 from ..model import Model
 
+if sys.version_info < (3, 8):
+    from typing_extensions import SupportsIndex
+else:
+    from typing import SupportsIndex
 
 AxisIndex = Union[SupportsIndex, slice, Sequence[SupportsIndex]]
 Index = Union[AxisIndex, Tuple[AxisIndex, ...]]
--- a/thinc/layers/array_getitem.py
+++ b/thinc/layers/array_getitem.py
@@ -1,9 +1,9 @@
-from typing import Union, Sequence, Tuple, TypeVar
+from typing import Union, Sequence, SupportsIndex, Tuple, TypeVar
 from ..types import ArrayXd, FloatsXd, IntsXd
 from ..model import Model
 
 
-AxisIndex = Union[int, slice, Sequence[int]]
+AxisIndex = Union[SupportsIndex, slice, Sequence[SupportsIndex]]
 Index = Union[AxisIndex, Tuple[AxisIndex, ...]]
 ArrayTXd = TypeVar("ArrayTXd", bound=ArrayXd)
 
--- a/thinc/layers/array_getitem.py
+++ b/thinc/layers/array_getitem.py
@@ -1,12 +1,7 @@
-import sys
 from typing import Union, Sequence, Tuple, TypeVar
-from ..types import ArrayXd, FloatsXd, IntsXd
+from ..types import ArrayXd, FloatsXd, IntsXd, SupportsIndex
 from ..model import Model
 
-if sys.version_info < (3, 8):
-    from typing_extensions import SupportsIndex
-else:
-    from typing import SupportsIndex
 
 AxisIndex = Union[SupportsIndex, slice, Sequence[SupportsIndex]]
 Index = Union[AxisIndex, Tuple[AxisIndex, ...]]
--- a/thinc/types.py
+++ b/thinc/types.py
@@ -17,7 +17,7 @@ if sys.version_info < (3, 8):
     from typing_extensions import SupportsIndex as SupportsIndex
 else:
     from typing import Protocol, Literal  # noqa: F401
-    from typing_extensions import SupportsIndex as SupportsIndex
+    from typing import SupportsIndex as SupportsIndex
 
 
 # fmt: off
--- a/thinc/types.py
+++ b/thinc/types.py
@@ -14,8 +14,10 @@ else:
 # Use typing_extensions for Python versions < 3.8
 if sys.version_info < (3, 8):
     from typing_extensions import Protocol, Literal
+    from typing_extensions import SupportsIndex as SupportsIndex
 else:
     from typing import Protocol, Literal  # noqa: F401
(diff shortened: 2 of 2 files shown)


willfrey:
I hope these small pull requests are okay! I realized that the initial PR body is a bit terse.

svlandeg:
Hi! Small PRs are fine, but preferably the original PR post contains a bit more information why the change is beneficial. Can you explain a bit more why you'd like to see this change?

willfrey:
@svlandeg Of course!

This is beneficial because `SupportsIndex` is a weaker precondition than requiring an explicit `int` as an index for a sequence, which makes the annotation more flexible and report fewer false positives, like in the following example.

```python3
from collections.abc import Sequence
from typing import TypeAlias

import numpy

Index: TypeAlias = int | slice | Sequence[int]


def func(index: Index | tuple[Index, ...], /) -> None:
    ...

func(numpy.int64(0))  # Error: Argument 1 to "func" has incompatible type "signedinteger[_64Bit]"; expected "Union[Union[int, slice, Sequence[int]], Tuple[Union[int, slice, Sequence[int]], ...]]"  [arg-type]
```

For example, it is perfectly valid to use numpy integers to index a sequence but numpy integers do not explicitly subclass `builtins.int`.

```python3
>>> import numpy
>>> index = numpy.int64(0)
>>> [1, 2, 3][index]
1
>>> isinstance(index, int)
False
```

And the reason that you want `SupportsIndex` instead of just `SupportsInt` is because types like `builtins.str` and `builtins.float` support conversion to `int` but they shouldn't be used as an index into a sequence when used with item access via `__getitem__`. The `__index__` special method was proposed in [PEP 357](https://peps.python.org/pep-0357/).

Use of `SupportsIndex` can be found throughout the numpy codebase. The smallest file in there that I could find to use an example is [numpy/_typing/_shape.py](https://github.com/numpy/numpy/blob/main/numpy/_typing/_shape.py).

thinc/layers/array_getitem.py
```diff
@@ -1,9 +1,14 @@
+import sys
 from typing import Union, Sequence, Tuple, TypeVar
 from ..types import ArrayXd, FloatsXd, IntsXd
 from ..model import Model
 
+if sys.version_info < (3, 8):
+    from typing_extensions import SupportsIndex
+else:
+    from typing import SupportsIndex
```
svlandeg, line 9: We can move this bit to `thinc.compat` and then import `SupportsIndex` from there.

thinc/layers/array_getitem.py
```diff
@@ -1,9 +1,14 @@
+import sys
 from typing import Union, Sequence, Tuple, TypeVar
 from ..types import ArrayXd, FloatsXd, IntsXd
 from ..model import Model
```
svlandeg, line 4:
```suggestion
from typing import Union, Sequence, Tuple, TypeVar
from ..types import ArrayXd, FloatsXd, IntsXd
from ..model import Model

import sys
```
(you couldn't know this, but this order is more in line with other files in the codebase. We typically have the `from typing` imports on the first line(s))

svlandeg (commented):
Thanks for the additional explanations, that's really helpful for reviewing 🙏 

I just had a few minor style comments - let us know if you need help implementing those.

thinc/layers/array_getitem.py
svlandeg: Oh sorry, actually this should go into `thinc.types` instead!

thinc/layers/array_getitem.py
willfrey: Removing the `import sys` statement because it's not needed since we're handling exposing `SupportsIndex` in `thinc.types.

thinc/layers/array_getitem.py
willfrey: Done!

Sicne the intention is to make `SupportsIndex` available as an public symbol in `thinc.types`, I've opted to do what PEP 484 wants you to do for stub files, which is to re-export the imported symbol as the same name. Pyright/Pylance _really_ doesn't like it if you don't do this (or explicitly declare a `__all__` full of literal strings) although mypy tends to be okay. In either case, it doesn't hurt anything.

svlandeg (approved):
Thanks again! I'll leave this open just a tad longer in case any of my colleagues want to chime in, but otherwise I think this is good to merge.

willfrey:
My pleasure! Thanks for all the thoughtful feedback. I’m happy to open more pull requests to improve the typing annotations. I’ll just be sure to explain why I’m doing what I’m doing in the original pull requests body 😅

willfrey:
I'll defer to whatever you all decide, of course, but if you'll allow me to bike-shed a little... 😅 

My personal thinking is that type checkers like mypy should only be concerned with issues of type correctness since they make no guarantees at runtime. To ensure users are only passing `builtins.int` and `builtins.slice` as the primitives of their `index`, you'd have to do something like the following (which would incur its own performance cost)

```python3
from collections.abc import Sequence
from typing import Any, TypeAlias, TypeGuard


AxisIndex: TypeAlias = int | slice | Sequence[int]
Index: TypeAlias = AxisIndex | tuple[AxisIndex, ...]


def is_axis_index(obj: object, /) -> TypeGuard[AxisIndex]:
    return (
        isinstance(obj, int | slice)
        or isinstance(obj, Sequence)
        and all(isinstance(i, int) for i in obj)
    )


def is_index(obj: object, /) -> TypeGuard[Index]:
    return (
        is_axis_index(obj)
        or isinstance(obj, tuple)
        and all(is_axis_index(i) for i in obj)
    )


def validate_index(value: Any, /) -> None:
    if not is_index(value):
        raise TypeError(f"{value!r} is not valid as an index")
```

and then use `validate_index` as a guard to guarantee correctness.

I guess I just really abhor false negatives from type checkers because that can promote other bad habits like overuse of `# type: ignore` directives, using `typing.cast` as a cudgel, or using other type constructors to convert the object myself to whatever the interface thinks it wants.

I have never worked on a mypy plugin, so this may not be feasible, but perhaps a compromise would be to have the thinc mypy plugin be configurable such that it can optionally raise errors about performance issues based on type checking?

svlandeg:
Hi @willfrey, thanks again for your PR and for the interesting discussion. We've talked about this some more internally and have finally decided that we'll prioritize performance over more flexible type annotations. Unfortunately that means we'll be closing this PR without merging it. I hope that doesn't discourage you from future contributions/discussions though!

willfrey:
@svlandeg No worries. I already have another issue that I opened that I’d be happy to open a PR for. I’ll wait for a response over there to see if you’d all be interested in it!

Issue #752, opened 2022-09-01
`SparseLinear` does not look up weights correctly


danieldk:
Reported in https://github.com/explosion/spaCy/discussions/11379

The weight matrix is allocated as follows:

```
model.set_param("W", model.ops.alloc((nO * length,), dtype="f"))
```

But indexed like this:

```
scores[clas] += weights[idx1 + clas] * value
scores[clas] += weights[idx2 + clas] * value
```

Where `idx1`/`idx2` is in `[0,length)` and `clas` is in `[0, nO)`. So the largest index ever used is `(length-1) + (clas-1)`, not `nO * length - 1`. Depending on how we look at `W`, we should either do something like

```
scores[clas] += weights[(idx1 * nO) + clas] * value
```

or

```
scores[clas] += weights[idx1 + (clas * length)] * value
```

PR #754, opened 2022-09-02
Add `SparseLinear_v2`, fixing indexing issues

danieldk:
Introduce `SparseLinear_v2` to fix indexing issues `SparseLinear` does not correctly index the gradient/weight matrix (#752). This change fixes the indexing, so that the full matrix is used.

To retain compatibility with existing models that use `SparseLinear`, which works relatively well if there are not too many hash collisions, the fixed version is renamed to `SparseLinear_v2`.

---

While at it, fix another indexing-related bug:

The output of MurMur hashes were mapped to array indices as follows:

```
idx = hash & (nr_weight-1)
```

This works well when `nr_weight` is a power of two. For instance,
if we have 16 buckets:

```
idx = hash & 15
idx = hash & 0b1111
```

However, when the user uses a bucket count that is not a power of two, this breaks down. For instance, if we have 15 buckets:

```
idx = hash & 14
idx = hash & 0b1110
```

This would mask out all odd indices. We fix this by using the modulus instead. To preserve compatibility with existing models, this change is only added to `SparseLinear_v2`.


adrianeboyd:
Do you really need to do the older version of the hash mapping at all?

danieldk:
> Do you really need to do the older version of the hash mapping at all?

I don't know. If someone used

```
SparseLinear(length=not_a_power_of_2)
```

their models would break by changing from bit masking to modulo.

adrianeboyd:
But with this as `SparseLinear_v2` I'm not sure where that would happen?

danieldk:
> But with this as `SparseLinear_v2` I'm not sure where that would happen?

I am not sure I follow? There may be existing uses for `SparseLinear`/`SparseLinear_v1` out there with lengths that are not a power of 2? So, we can't really remove `SparseLinear` without breaking the API. But we can also not retroactively fix `SparseLinear`, because the indices would change and existing models would output garbage.

thinc/layers/sparselinear.pyx
```diff
@@ -110,41 +127,57 @@ class _finish_linear_update:
 
 cdef void set_scoresC(float* scores,
         const uint64_t* keys, const float* values, const int32_t* lengths,
-        int batch_size, int nr_out,
-        const float* weights, int nr_weight) nogil:
+        int batch_size, int nr_out, const float* weights, int nr_weight,
+        bint invalid_indexing) nogil:
```
shadeMe, line 131: I think it'd be a good idea to add a comment clarifying what the `invalid_indexing` parameter is used for (so that one wouldn't have to go through the Git history to understand why it was introduced).

adrianeboyd:
Ah, I thought more was redefined than actually is. I think it would be clearer to call it something like `v1_indexing`.

thinc/layers/sparselinear.pyx
danieldk: Done.

danieldk:
> Ah, I thought more was redefined than actually is. I think it would be clearer to call it something like `v1_indexing`.

Changed the name to `v1_indexing`.

svlandeg:
You'll need to merge in the latest from `master` to avoid the CI failure with the macOS 10.15 environment.

danieldk:
> You'll need to merge in the latest from `master` to avoid the CI failure with the macOS 10.15 environment.

Merged.

shadeMe (approved):

shadeMe:

thinc/tests/layers/test_layers_api.py
```diff
@@ -128,6 +128,7 @@ def assert_data_match(Y, out_data):
     # ("CauchySimilarity.v1", {}, (array2d, array2d), array1d),
     ("ParametricAttention.v1", {}, ragged, ragged),
     ("SparseLinear.v1", {}, (numpy.asarray([1, 2, 3], dtype="uint64"), array1d, numpy.asarray([1, 1], dtype="i")), array2d),
+    ("SparseLinear.v2", {}, (numpy.asarray([1, 2, 3], dtype="uint64"), array1d, numpy.asarray([1, 1], dtype="i")), array2d),
```
svlandeg, line 131: Can we think of a specific test that would have caught this bug, though?

danieldk: I was thinking about this. Say that we have a sparselinear of a certain size and a certain variety of inputs, we can compute the expected number of non-zero values. Of course, it won't be exact since the hash is a randomization and there are collisions. But if the actual number of non-zeros is within a certain bandwidth (I guess the number of non-zeros are normally distributed), then the test would pass. We could do this for the rows of very class.

Hope that makes sense?

website/docs/api-layers.md
```diff
@@ -802,6 +802,42 @@ length, describing the concatenated batch of input features and their values.
...
+| ----------- | --------------------------------------------------------- | -------------------------------------------------------- |
+| `nO`        | <tt>Optional[int]</tt>                                    | The size of the output vectors.                          |
+| `length`    | <tt>int</tt>                                              | The size of the weights vector, to be tuned empirically. |
+| **RETURNS** | <tt>Model[Tuple[ArrayXd, ArrayXd, ArrayXd], ArrayXd]</tt> | The created layer.                                       |
+
+```python
+https://github.com/explosion/thinc/blob/master/thinc/layers/sparselinear.pyx
+```
+
+### SparseLinear_v2 {#sparselinear_v2 tag="function"}
```
svlandeg, line 823: add `new` annotation

adrianeboyd:
Would the slow GPU tests pass after the changes Madeesh made for tensorflow?

shadeMe:
The fix for the failing tests is in [this commit](https://github.com/explosion/thinc/commit/fc323e177a6edd70de60b56b581addf2aa3e9371). So, performing a new rebase on `master` ought to sort out the GPU test failures.

danieldk:

website/docs/api-layers.md
danieldk: Added.

thinc/tests/layers/test_layers_api.py
danieldk: Added a test that does this.

danieldk:

PR #686, opened 2022-05-30
xp2{tensorflow,torch}: convert NumPy arrays using dlpack

danieldk:
Newer versions of NumPy can expose arrays as dlpack capsules. Use this functionality (when supported) to speed up NumPy -> Torch/Tensorflow array conversion.

PR #757, opened 2022-09-07
`util`: Explicitly call `__dlpack__` built-in method in `xp2tensorflow`

shadeMe:
`tf.experimental.dlpack.from_dlpack` expects a `PyCapsule` object.

### DIFF
--- a/thinc/util.py
+++ b/thinc/util.py
@@ -404,7 +404,8 @@ def xp2tensorflow(
         dlpack_tensor = xp_tensor.toDlpack()  # type: ignore
         tf_tensor = tf.experimental.dlpack.from_dlpack(dlpack_tensor)
     elif hasattr(xp_tensor, "__dlpack__"):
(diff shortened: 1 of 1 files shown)


svlandeg (commented):
Thanks!

Was this simply not working before? Can we add a test to capture that?

shadeMe:
>Was this simply not working before? Can we add a test to capture that?

That particular code path was never actually called until recently, as the `__dlpack__` was [introduced in a newer version of `numpy`](https://github.com/explosion/thinc/pull/686) than the one we currently support. The bug above was in fact uncovered by an existing test in the GPU test suite.

shadeMe:

shadeMe:
> I think all the buildkite tests are going to fail until we fix the numpy-related problems with the scripts.

Yup, but this PR should at least fix some of those failures, which is what I wanted to check.

svlandeg:
So this is good to merge, yes?

shadeMe:
> So this is good to merge, yes?

Yeah, I'd say so!

PR #759, opened 2022-09-09
Extend tests to include newly-added data types

rmitsch:
## Description

Extend tests to include data types added via relaxed signatures in https://github.com/explosion/thinc/pull/599.

### Types of change

Enhancement.

## Checklist

- [ ] My changes don't require a change to the documentation, or if they do, I've added all required information.

In response to https://github.com/explosion/thinc/pull/599#discussion_r864553335.

### DIFF
--- a/thinc/tests/backends/test_ops.py
+++ b/thinc/tests/backends/test_ops.py
@@ -15,9 +15,8 @@ from thinc.api import LSTM
 from thinc.types import Floats2d
 import inspect
 
-from .. import strategies
-from ..strategies import arrays_BI, ndarrays_of_shape
-
+from thinc.tests import strategies
+from thinc.tests.strategies import ndarrays_of_shape
 
 MAX_EXAMPLES = 10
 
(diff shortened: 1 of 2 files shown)


danieldk:
> @danieldk Does this kind of modification to the tests reflect what you had in mind? If so, I'll go on adjusting the others that are relevant w.r.t. [#599 (comment)](https://github.com/explosion/thinc/pull/599#discussion_r864553335).

Looks good indeed!

I don't think this is easily possible, but maybe something we want to look into longer term -- ideally, we'd generate examples based on the type signatures. So, if the type signature would change, the generated data would change with it.

Though, I think that's probably a very large undertaking, so going though the types and trying to modify the tests accordingly seems like the best short-term solution.

rmitsch:
> I don't think this is easily possible, but maybe something we want to look into longer term -- ideally, we'd generate examples based on the type signatures. 

That's interesting - is there already a concept on how to do that? I can imagine having a function deriving input types from another function like
```python
@pytest.mark.parametrize("cpu_ops", CPU_OPS)
@settings(max_examples=MAX_EXAMPLES * 2, deadline=None)
@given(X=strategies.generate_examples_for(ops.flatten))
def test_flatten_unflatten_roundtrip(cpu_ops: NumpyOps, X: numpy.ndarray):
    ...
```
where `strategies.generate_examples_for()` uses `inspect.signature(ops.flatten)` to fetch the signature and then translates Thinc types to `numpy` types using `_Array.dtype()`, and finally picks a `hypothesis` strategy based on the `numpy` type. 
That sketch doesn't feel like a _that_ much work though, so I'm probably overlooking something :smile: 

```

**Comment by danieldk** [comment]
> That's interesting - is there already a concept on how to do that?

No, I don't think we have anything like that yet.

> This sketch doesn't feel like a _that_ much work though, so I'm probably overlooking something 😄

Try it 👍. As long as it's opt-in per test, we could enable them gradually. For reviewing I think it's the nicest to do this as a proof of concept for maybe two or three methods first. Avoids a lot of rewrites as a result of review comments. Then once we are happy with the design, we could apply it to more methods.

**Comment by danieldk** [comment]
> Considering this I'd recommend implementing the type-dynamic testing once thinc drops support for 3.6. What do you think?

Sounds completely sensible to me, we should avoid relying on internals of external dependencies.

**Comment by richardpaulhudson** [comment]
An alternative to waiting for support for 3.6 to be dropped would be to decorate the tests with

```
@unittest.skipIf(sys.version_info.minor < 7, 'Test unsupported for old version')
```

**Comment by richardpaulhudson** [comment]
An alternative to waiting for 3.6 support to be dropped before implementing the tests would be to decorate them with

```
@pytest.mark.skipif(sys.version_info < (3, 7), reason="test tooling requires python3.7 or higher")
```

PR #763, opened 2022-09-09
Standardize `blis` calls in `NumpyOps`

shadeMe:
Initial work to eventually make the dependency on `blis` optional.


thinc/backends/numpy_ops.pyx
```diff
@@ -616,7 +611,7 @@ cdef void cpu_update_averages(weight_t* ema,
         ema[i] -= one_minus_decay * (ema[i] - weights[i])
 
 
-def lstm_forward_training(
+def lstm_forward_training(CBlas cblas,
```
svlandeg, line 614: Why not use `self.cblas()` like `gather_add` does? It's a little more indirect but doesn't require changing the function args. Or is there a reason you want to be able to pass a specific (different) `Cblas`?

shadeMe: This function isn't a method of `Ops`, so that's not an option here. As for the above methods, we could very well call `self.cblas()` instead of instantiating the `CBlas` struct directly. We save a Python call with the latter, but the overhead will be negligible in most, if not all, cases. I can change that.

svlandeg (approved):

thinc/backends/numpy_ops.pyx
```diff
@@ -95,7 +90,7 @@ class NumpyOps(Ops):
         y = self.as_contig(y)
         if out is not None:
             out = self.as_contig(out)
-        return blis.py.gemm(x, y, out=out, trans1=trans1, trans2=trans2, beta=0.)
+        return blis_py.gemm(x, y, out=out, trans1=trans1, trans2=trans2, beta=0.)
```
danieldk, line 93: I think we can switch on the dtype of the array here and use `self.cblas().sgemm`/`self.cblas().dgemm`?

danieldk (approved):

shadeMe:

PR #766, opened 2022-09-13
Re-export `floats_getitem` and `ints_getitem` in `thinc.layers` and `thinc.api`

### LINKED ISSUE #749: `floats_getitem` and `ints_getitem` missing from `thinc.api`
These two layers don't get imported and re-exported into `thinc/layers/__init__.py`. Happy to open a PR to fix this unless you have a reason that you don't want to expose them in `thinc.api`.

willfrey:
These were missing in `thinc.api`, so this exposes them.

Closes #749.


willfrey:
Looks like this is failing because numpy is using positional-only arguments in a stub file, which is a language feature added in Python 3.8. Strange that it's failing on Python 3.10 on Windows, though?

kadarakos:
Hey @willfrey,

Just saw your Issue #749 and sorry for not coming back earlier. Thank you for bringing up the `_getitem` layers! What we were thinking is that we would not expose these layers while `thinc` is in `8.x` version. Then for `v9` it might be better to just have `array_getitem` that is generic across `IntsXd` and `FloatsXd` and then have that layer exposed in the future instead.

willfrey:
Gotcha! I’ll close this PR then.

Issue #765, opened 2022-09-12
Improving type definitions for forward/backward in models


willfrey:
Right now forward and backward/backprop in models are just plain `Callable` but this could be changed to be in agreement with the [concept and design docs](https://thinc.ai/docs/concept) and how models appear to be called throughout the codebase.

It'd be a little more verbose than this to accommodate Python versions before 3.8, but the idea would be something like the following

```python3
# thinc/types.py

from __future__ import annotations

from typing import TYPE_CHECKING, Protocol, TypeAlias
from collections.abc import Callable

# Avoid circular imports because we need to use these in thinc/model.py
if TYPE_CHECKING:
    from .model import InT, OutT, Model


Backward: TypeAlias = Callable[[OutT], InT]


class Forward(Protocol):
    def __call__(
        self, model: Model, X: InT, /, is_train: bool
    ) -> tuple[OutT, Backward[OutT, InT]]:
        ...
```

```python3
# thinc/model.py

# NOTE: I'm using shorthand for brevity to convey the idea at the sake of completeness

from .types import Backward, Forward


class Model(Generic[InT, OutT]):
    _func: Forward[InT, OutT]
    
    def __init__(self, ..., forward: Forward[InT, OutT], ...): ...
    def __call__(self, X: InT, is_train: bool) -> tuple[OutT, Backward[OutT, InT]]: ...
```

This would have at least two benefits:

1. Improved type checking between a layer's factory function and the its forward function definition.
2. Improved type inference inside of layers that make use of other layers since the backward/backprop callback function will be correctly annotated instead of being `(...) -> Any`.

I'm happy to submit a PR with these changes but I wanted to check here before I did since while it's not a ton of work, it's a non-trivial amount for you all to review.

Thanks!

PR #767, opened 2022-09-13
Update `Model.__call__` callback signature

willfrey:
Since `Model` is generic, this should help with type inference when you're actually using the backprop/backward callback in layers/architectures/etc. Right now the type for every callback is `(...) -> Any`, meaning it can take anything and returns anything.

Hope this is okay 😄

### DIFF
--- a/thinc/model.py
+++ b/thinc/model.py
@@ -285,7 +285,7 @@ class Model(Generic[InT, OutT]):
         else:
             raise ValueError("Cannot add reference to node not in tree.")
 
-    def __call__(self, X: InT, is_train: bool) -> Tuple[OutT, Callable]:
+    def __call__(self, X: InT, is_train: bool) -> Tuple[OutT, Callable[[OutT], InT]]:
         """Call the model's `forward` function, returning the output and a
         callback to compute the gradients via backpropagation."""
         return self._func(self, X, is_train=is_train)


willfrey:
Related to #765. This is probably the best, smallest first step. You do get a lot of benefit from adding that forward callback protocol though, since mypy will automatically infer the generic type variables for the model based on the types defined for the actual `def forward(...)` or equivalent passed `Model`'s forward parameter.

thinc/model.py
```diff
@@ -285,7 +285,7 @@ def set_ref(self, name: str, value: Optional["Model"]) -> None:
         else:
             raise ValueError("Cannot add reference to node not in tree.")
 
-    def __call__(self, X: InT, is_train: bool) -> Tuple[OutT, Callable]:
+    def __call__(self, X: InT, is_train: bool) -> Tuple[OutT, Callable[[OutT], InT]]:
```
danieldk, line 288: We can't universally commit to this signature. For instance, tok2vec models have a signature like

```
Model[List[Doc], List[Floats2d]]
```

But you cannot return gradients of the type `IntT`, which is `List[Doc]` in that case.

willfrey: I see! Thanks for pointing that out. I'll close this PR :)

PR #776, opened 2022-09-27
update to latest mypy and exclude Python 3.6

svlandeg:
The latest version of `mypy` (0.981) now works with Python 3.10, so we're updating and at the same time dropping support for `mypy` usage with Python 3.6.

The new version of `mypy` identified a few problematic issues in the repo that are fixed in this PR as well. They all revolved around the fact that a `TypeVar` doesn't make sense to be used as a returning type when it's not also used in the arguments of the same function.

Follow-up on https://github.com/explosion/thinc/pull/768


adrianeboyd:
Let's release confection v0.0.2 including `py.typed` and then make sure everything works here before merging this PR.

svlandeg:
@explosion-bot please test_gpu --run-on spacy --run-on-branch master --thinc-branch current_pr

svlandeg:
@explosion-bot please test_gpu --run-on spaCy --run-on-branch master --thinc-branch current_pr

PR #779, opened 2022-10-03
Support both Python 3.6 and Pydantic 1.10

svlandeg:
Adding support for Pydantic 1.10.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,8 +9,7 @@ catalogue>=2.0.4,<2.1.0
 confection>=0.0.1,<1.0.0
 ml_datasets>=0.2.0,<0.3.0
 # Third-party dependencies
-pydantic>=1.7.4,!=1.8,!=1.8.1,<1.10.0; python_version < "3.7"
-pydantic>=1.7.4,!=1.8,!=1.8.1,<1.11.0; python_version >= "3.7"
+pydantic>=1.7.4,!=1.8,!=1.8.1,<1.11.0
 numpy>=1.15.0
 # Backports of modern Python features
 dataclasses>=0.6,<1.0; python_version < "3.7"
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -9,7 +9,8 @@ catalogue>=2.0.4,<2.1.0
 confection>=0.0.1,<1.0.0
 ml_datasets>=0.2.0,<0.3.0
 # Third-party dependencies
-pydantic>=1.7.4,!=1.8,!=1.8.1,<1.10.0
+pydantic>=1.7.4,!=1.8,!=1.8.1,<1.10.0; python_version < "3.7"
```
polm, line 12: I was confused by the specifically excluded versions here, but it looks like Pydantic has a `1.8` but no `1.8.0`. Just commenting for anyone else unfamiliar with this.

requirements.txt
```diff
@@ -9,7 +9,8 @@ catalogue>=2.0.4,<2.1.0
...
+pydantic>=1.7.4,!=1.8,!=1.8.1,<1.11.0; python_version >= "3.7"
```
adrianeboyd, line 13: You don't need the split requirements, pip just won't install newer versions if they're marked as python 3.7+.

```suggestion
pydantic>=1.7.4,!=1.8,!=1.8.1,<1.11.0
```

setup.cfg
```diff
@@ -49,7 +49,8 @@ install_requires =
     # Third-party dependencies
     setuptools
     numpy>=1.15.0
-    pydantic>=1.7.4,!=1.8,!=1.8.1,<1.10.0
+    pydantic>=1.7.4,!=1.8,!=1.8.1,<1.10.0; python_version < "3.7"
+    pydantic>=1.7.4,!=1.8,!=1.8.1,<1.11.0; python_version >= "3.7"
```
adrianeboyd, line 53:
```suggestion
    pydantic>=1.7.4,!=1.8,!=1.8.1,<1.11.0
```

svlandeg:
@adrianeboyd : honestly, I'm not sure where my brain went to have the split requirements. Haha. Anyway, thanks. Merging 🎉

Issue #777, opened 2022-09-27
Type checkers don't like non-literal `__all__` in `thinc.api` for exposing the public interface


willfrey:
If you forbid implicit reexports in Mypy (or turn Pyright/Pylance on at all), the type checkers will complain about `thinc.api` not exposing anything publicly. This is because the imports are of the form `from name import target` and `__all__` can only be inferred at runtime.

Would there be any objection to doing either of the following?

1. Explicitly re-export names in `thinc.api` through imports of the form `from namespace import target as target`.
2. Change `__all__` to be made up of string literals for each public member in `thinc.api`.

The first option eliminates the need for a `__all__` entirely, if you're concerned about having to repeat each public interface member three times.

If you're okay with either (or both), I'm happy to make the changes and open a PR.

Thank you!

svlandeg:
Hi! Thanks for the report, we'll have a look how best to address this issue!

PR #780, opened 2022-10-06
Make __all__ static to support type checking

rmitsch:
## Description

Resolves https://github.com/explosion/thinc/issues/777:
- Converts `__all__` into static declarations where necessary to pass a `mypy` check with `--no-implicit-reexport`.
- Adds a test checking whether `__all__` in modules supposed to reexport all imports (`thinc/__init__.py`,  `thinc/api.py`, `thinc/shims/__init__.py`, `thinc/layers/__init__.py`) matches their imports.

### Types of change

Chore.

## Checklist


thinc/__init__.py
```diff
@@ -3,3 +3,9 @@
 
 from .about import __version__
 from .config import registry
+
+__all__ = [
+    "registry",
+    "__version__",
+    "numpy"
```
adrianeboyd, line 10: I don't think we should reexport `numpy` anywhere.

rmitsch: I agree. I already have one exception for `thinc/__init__.py`, which is to consider `__version__` as to be exported despite being prefixed with `_` - I could another one? Or we don't treat `thinc/__init__.py` as one of the modules that reexport everything they import.

thinc/api.py
```diff
@@ -46,4 +46,62 @@
 from .layers import reduce_first, reduce_last, reduce_max, reduce_mean, reduce_sum
 
 
-__all__ = list(locals().keys())
+__all__ = [
```
adrianeboyd, line 49: Does this need a `# fmt: off`?

thinc/__init__.py
adrianeboyd: Maybe try to figure out whether the import is actually necessary now?

thinc/api.py
rmitsch: The formatting is as intended. Do we prefer a different one?

adrianeboyd: `black` will reformat this, though.

thinc/__init__.py
rmitsch: Internal discussion indicates that this is likely still necessary. I added an exception for `numpy`, but I'm open to alternative suggestions.

thinc/__init__.py
```diff
@@ -3,3 +3,11 @@
 
 from .about import __version__
 from .config import registry
+
+
+# fmt: off
+__all__ = [
+    "registry",
+    "__version__"
```
adrianeboyd, line 11: For future editing/diffs, add trailing commas to all of the lists, or in the cases where you don't really need the `fmt` tags, you can just format them with black.

thinc/compat.py
```diff
@@ -88,3 +88,13 @@
...
 has_gpu = has_cupy_gpu or has_torch_mps_gpu
+
+__all__ = [
+    "cupy",
+    "cupyx",
+    "torch",
+    "tensorflow",
+    "mxnet",
+    "h5py",
+    "os_signpost"
```
adrianeboyd, line 99: Did you reformat with black? It should have caught this?

```suggestion
    "os_signpost",
```

rmitsch: No - did so in https://github.com/explosion/thinc/pull/780/commits/312ded305fe53cb19fec24c27ffb6b3edadebd25. Thanks for the catch.

thinc/tests/strategies.py
```diff
@@ -122,3 +122,22 @@ def arrays_OPFI_BI_lengths(max_B=5, max_P=3, max_F=5, max_I=8):
...
+    "ndarrays_of_shape",
+    "ndarrays",
+    "arrays_BI",
+    "arrays_BOP",
+    "arrays_BOP_BO",
+    "arrays_BI_BO",
+    "arrays_OI_O_BI",
+    "arrays_OPFI_BI_lengths",
+    "floats",
+]
```
adrianeboyd, line 143: I don't think we want to reexport anything in the test suite, especially not function names that kind of shadow core library methods. Is there are a way to avoid this check in the test suite instead? It could be a setting in `setup.cfg` if it needs to be?

adrianeboyd:
Could you also run `mypy --no-implicit-reexport` on spacy and check whether there are any additional thinc-related errors?

thinc/tests/strategies.py
rmitsch: Strangely, 
```ini
[mypy-thinc.tests.*]
implicit_reexport = False
```
does _not_ work in disabling the implicit-reexport check (it also doesn't work for enabling them). Two alternative workarounds:
- Adding `# type: ignore[attr-defined]` to problematic imports.
- Defining a minimal `__all__` in `thinc.tests.strategies` - the current list could be reduced to `["floats"]` and still pass `mypy thinc --no-implicit-reexport`.

adrianeboyd: You want:

```
[mypy-thinc.tests.*]
implicit_reexport = True
```

rmitsch:
@explosion-bot please test_cpu --run-on spaCy --run-on-branch master --thinc-branch current_pr

rmitsch:
@explosion-bot please test_gpu --run-on spaCy --run-on-branch master --thinc-branch current_pr

rmitsch:
> Could you also run mypy --no-implicit-reexport on spacy and check whether there are any additional thinc-related errors?

Hm, installed my local state of `thinc` with `pip install -e thinc` in the venv I use for spaCy (after uninstalling the distributed version). A `python -m mypy spacy --no-implicit-reexport | grep thinc` still shows issues having been fixed in `thinc` already, such as 
```spacy/compat.py:38: error: Module "thinc.api" does not explicitly export attribute "Optimizer"; implicit reexport disabled```
- `Optimizer` is in `__all__`. Dunno why `mypy` doesn't pick up on that?

**Comment by adrianeboyd** [comment]
I tried it and only ended up with a few things in `thinc.config` that actually aren't exported. Maybe try a standard (non-editable) install in a new venv just to check?

**Comment by rmitsch** [comment]
Yeah, seems to have been an issue with the env. Running mypy on a clean env yields this for me:
```
spacy/schemas.py:9: error: Module "thinc.config" does not explicitly export attribute "Promise"; implicit reexport disabled
spacy/cli/debug_config.py:5: error: Module "thinc.config" does not explicitly export attribute "VARIABLE_RE"; implicit reexport disabled
```
Both are included in https://github.com/explosion/thinc/pull/780/commits/14b3e4af219e3f7eca1b9a64e43c78ed412c808b. Does this match with what you got?

adrianeboyd:
Yeah, that's what I got too.

PR #791, opened 2022-10-18
Updates for python 3.11, set version to v7.4.6

adrianeboyd:


adrianeboyd:
Python 3.11 will require https://github.com/explosion/srsly/pull/76

PR #796, opened 2022-10-24
`PyTorchShim`: Add serde callbacks to facilitate lazy loading models

shadeMe:
When attempting to lazily initialize/load PyTorch models, special care needs to be taken when deserializing trained models from disk. Since the `init` method of `Model`s are not called during deserialization, shape inference  cannot be performed before restoring model parameters. Therefore, the necessary information required to initialize the model must additionally be serialized to disk for later use during deserialization.


thinc/shims/pytorch.py
```diff
@@ -40,6 +47,8 @@ def __init__(
         mixed_precision: bool = False,
         grad_scaler: Optional[PyTorchGradScaler] = None,
         device: Optional["torch.device"] = None,
+        serialize_model_config: Optional[Callable[[Any], Dict[str, Any]]] = None,
+        deserialize_model_from_config: Optional[Callable[[Dict[str, Any]], Any]] = None,
```
danieldk, line 51: I gave this some more thought and realized that it may be better to give the serialization functions full control of the serialization of the Torch model? So the signatures would be:

```suggestion
        serialize_model_config: Optional[Callable[[Any], bytes]] = None,
        deserialize_model_from_config: Optional[Callable[[bytes], Any]] = None,
```

In this way, the caller could also apply more advanced transformations, such as removing certain parameters, etc.

thinc/shims/pytorch.py
```diff
@@ -40,6 +48,8 @@ def __init__(
         mixed_precision: bool = False,
         grad_scaler: Optional[PyTorchGradScaler] = None,
         device: Optional["torch.device"] = None,
+        serialize_model: Optional[Callable[[Any], bytes]] = None,
```
danieldk, line 51: Can't we pass `default_(de)?serialize` models as default values for these arguments?

shadeMe: I just wanted to keep the parameters in line with [how it's done in `PyTorchWrapper`](https://github.dev/explosion/thinc/blob/e4a11573fdc6ee3c6c0f03641d4dd2407ac4f9bd/thinc/layers/pytorchwrapper.py#L76). This also lets us define the default methods anywhere in the module (as opposed to above the class for visibility when using them as default arguments).

danieldk (approved):
Looks good to me, but since this is a fairly large API extension, someone else should also look at it.

rmitsch (commented):
Looks reasonable from what I can tell with my limited exposure to `thinc`. Only minor comments.

thinc/shims/pytorch.py
rmitsch: I agree that default args in both places would be nicer. If in doubt, I'd favor consistency.

danieldk: So, after some discussion, we decided to leave this as-is.

thinc/layers/pytorchwrapper.py
```diff
@@ -132,6 +132,84 @@ def PyTorchWrapper_v2(
...
+        returns a "bytes" representation of the same. The representation should
+        contain all the necessary information to fully deserialize the model.
+
+        When set to "None", the default serializer serializes the model's parameters.
+    deserialize_model:
+        Callback that receives the default PyTorch model (passed to the constructor), the
+        serialized "bytes" representation and a PyTorch device. It should return a
+        fully deserialized model on the target device as its result.
+
+        When set to "None", the default dserializer deserializes the model's parameters.
```
adrianeboyd, line 189:
```suggestion
        When set to "None", the default deserializer deserializes the model's parameters.
```

thinc/layers/pytorchwrapper.py
```diff
@@ -132,6 +132,84 @@ def PyTorchWrapper_v2(
...
+    device:
+        The PyTorch device to run the model on. When this argument is
+        set to "None", the default device for the currently active Thinc
+        ops is used.
+    serialize_model:
+        Callback that receives the wrapped PyTorch model as its argument and
+        returns a "bytes" representation of the same. The representation should
+        contain all the necessary information to fully deserialize the model.
+
+        When set to "None", the default serializer serializes the model's parameters.
```
adrianeboyd, line 183:
```suggestion
        When set to "None", the default serializer serializes the model's parameters.
```

thinc/layers/pytorchwrapper.py
```diff
@@ -132,6 +132,84 @@ def PyTorchWrapper_v2(
...
+    deserialize_model:
+        Callback that receives the default PyTorch model (passed to the constructor), the
+        serialized "bytes" representation and a PyTorch device. It should return a
+        fully deserialized model on the target device as its result.
+
+        When set to "None", the default deserializer deserializes the model's parameters.
```
adrianeboyd, line 189:
```suggestion
        When set to "None", the default deserializer deserializes the model's parameters.
```

thinc/layers/pytorchwrapper.py
```diff
@@ -11,7 +11,7 @@
 
 @registry.layers("PyTorchRNNWrapper.v1")
 def PyTorchRNNWrapper(
-    pytorch_model,
+    pytorch_model: "torch.nn.Module",
```
adrianeboyd, line 14: Different types in the new layer is fine, but I might err on the side of not modifying the types in the existing layers?

thinc/tests/layers/test_pytorch_wrapper.py
```diff
@@ -1,9 +1,14 @@
...
 from thinc.api import xp2torch, torch2xp, ArgsKwargs, use_ops
 from thinc.api import chain, get_current_ops, Relu
 from thinc.api import CupyOps, MPSOps, NumpyOps
 from thinc.backends import context_pools
+from thinc.layers.pytorchwrapper import PyTorchWrapper_v3
 from thinc.shims.pytorch_grad_scaler import PyTorchGradScaler
+from thinc.shims.pytorch import (
+    default_deserialize_torch_model,
+    default_serialize_torch_model,
+)
```
adrianeboyd, line 11: formatting

thinc/layers/pytorchwrapper.py
shadeMe: Fixed.

PR #799, opened 2022-10-27
Add missing packaging requirement

adrianeboyd:
* Test `import thinc` before dev requirements are installed.
* Add requirement for `packaging`.


adrianeboyd:
This should initially fail, then I will actually add the `packaging` requirement.

PR #804, opened 2022-11-24
Give schedules access to the key, step, and last eval score

danieldk:
<s>This should target `v9`, but since there is currently an incompatibility between v9 and spaCy, it's still on `master` for testing.</s>

Before this change schedules were generators that generate a value for each training step. This, however has the limitation that scheduler cannot use other information that is available in the optimizer such as the parameter key. This information is useful for e.g. discriminative learning rates, where certain parameters are on a different schedule than others.

To accommodate passing additional information, this change converts schedules to instances of the `Schedule` object. The `Schedule` class is used with composition rather than inheritance, similar to `Model`. The schedules are passed the training step, the parameter key, and the last evaluation score (when available).

Traditional scalar and generated schedules are converted to callables by the optimizer for compatibility.


thinc/schedules.py
```diff
@@ -73,50 +96,54 @@ def slanted_triangular(
     ratio: int = 32,
     decay: float = 1.0,
```
shadeMe, line 97: `decay` is unused.

thinc/schedules.py
```diff
@@ -73,50 +96,54 @@ def slanted_triangular(
...
+) -> ScheduleCallable:
     """Yield an infinite series of values according to Howard and Ruder's
     "slanted triangular learning rate" schedule.
     """
     cut = int(num_steps * cut_frac)
-    while True:
-        t += 1
+
+    def callback(*, step: int, **kwargs) -> float:
+        t = step + 1
```
shadeMe, line 106: As with `decaying`, this should be something like `t_new = t + step + 1` to preserve the old functionality.

website/docs/usage-config.md
```diff
@@ -201,10 +201,11 @@ from typing import Iterable
 import thinc
 
 @thinc.registry.schedules("my_cool_decaying_schedule.v1")
-def decaying(base_rate: float, decay: float, *, t: int = 0) -> Iterable[float]:
-    while True:
-        yield base_rate * (1.0 / (1.0 + decay * t))
-        t += 1
+def decaying(base_rate: float, decay: float, *, t: int = 0) -> SchedulerCallable:
```
shadeMe, line 204: Same change as the one in `schedules.py`.

website/docs/usage-training.md
```diff
@@ -225,37 +224,36 @@ normalize = true
...
+You can use schedules directly, by calling the schedule with the `step` keyword
+argument and using it to update hyperparameters in your training loop. Since
+schedules are particularly common for optimization settings, the
+[`Optimizer`](/docs/api-optimizer) object accepts floats, lists, iterators, and
+schedule callbacks for most of its parameters. When you call
 [`Optimizer.step_schedules`](/docs/api-optimizer#step_schedules), the optimizer
-will draw the next value from the generators and use them to change the given
-attributes. For instance, here's how to create an instance of the `Adam`
-optimizer with a custom learning rate schedule:
+will increase its step count and pass this step count to the schedules. For
```
shadeMe, line 236:
```suggestion
will increase its step count and pass it to the schedules. For
```

website/docs/usage-training.md
```diff
@@ -225,37 +224,36 @@ normalize = true
...
+instance, here's how to create an instance of the `Adam` optimizer with a custom
```
shadeMe, line 237:
```suggestion
instance, this is how one creates an instance of the `Adam` optimizer with a custom
```
This sounds slightly better, I find.

thinc/optimizers.py
```diff
@@ -354,4 +386,34 @@ def _adam(self, ops, weights, gradient, lr_scale, key, nr_upd):
         )
 
 
+def _generator_schedule(attr_name: str, schedule: Generator) -> ScheduleCallable:
+    last_step = 0
+    try:
+        value = next(schedule)
+    except (StopIteration, TypeError) as e:  # schedule exhausted, use last value
```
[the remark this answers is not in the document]
danieldk: Moved down where it actually uses the last value.

thinc/schedules.py
```diff
@@ -1,30 +1,50 @@
...
+    def callback(*args, **kwags) -> float:
+        return rate
+
+    return callback
 
 
 @registry.schedules("decaying.v1")
-def decaying(base_rate: float, decay: float, *, t: int = 0) -> Iterable[float]:
+def decaying(base_rate: float, decay: float, *, t: int = 0) -> ScheduleCallable:
     """Yield an infinite series of linearly decaying values,
```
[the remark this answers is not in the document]
danieldk: Fixed here and for `slanted_triangular`. Also added tests for `t != 0.0`.

thinc/schedules.py
```diff
@@ -35,15 +55,17 @@ def decaying(base_rate: float, decay: float, *, t: int = 0) -> Iterable[float]:
...
+
+    def callback(*, step: int, **kwargs) -> float:
+        return base_rate * (1.0 / (1.0 + decay * step))
+
+    return callback
 
 
 @registry.schedules("compounding.v1")
 def compounding(
     start: float, stop: float, compound: float, *, t: float = 0.0
```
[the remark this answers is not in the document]
danieldk: Changed so that it gets used and is tested.

thinc/schedules.py
danieldk: Removed.

website/docs/usage-config.md
danieldk: Fixed.

website/docs/usage-training.md
danieldk: Fixed.

website/docs/usage-training.md
danieldk: Fixed.

thinc/optimizers.py
```diff
@@ -221,28 +238,40 @@ def __call__(
...
             return weights, gradient
+
         ops = get_array_ops(weights)
         self.nr_update[key] += 1
         nr_upd = self.nr_update[key]
-        if self.L2 != 0 and not self.L2_is_weight_decay:
-            gradient += self.L2 * weights
-        if self.grad_clip:
-            gradient = ops.clip_gradient(gradient, self.grad_clip)
+        if self.L2(**self._schedule_args(key)) != 0 and not self.L2_is_weight_decay:
```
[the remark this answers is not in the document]
danieldk: Done.

thinc/optimizers.py
```diff
@@ -111,15 +112,17 @@ class Optimizer(object):
     schedules: Dict[str, Generator]
     nr_update: Dict[KeyT, int]
     last_seen: Dict[KeyT, int]
-    grad_clip: float
-    learn_rate: float
-    b1: float
-    b2: float
-    eps: float
-    L2: float
+    grad_clip: ScheduleCallable
```
kadarakos, line 115: Why not `ScheduleT` ?

thinc/optimizers.py
```diff
@@ -168,13 +173,14 @@ def __init__(
         L2_is_weight_decay (bool): Whether to interpret the L2 parameter as a
             weight decay term, in the style of the AdamW optimizer.
         """
+        self._step = 0
+        self._last_score = None
```
kadarakos, line 177: Maybe a type-hint that this will be a `Tuple[int, Comparable]` in the future. Or not sure what to put in place of `Comparable` , maybe its always just a `float` dunno. https://github.com/python/typing/issues/59

thinc/optimizers.py
```diff
@@ -189,24 +195,35 @@ def __init__(
...
+
+    @property
+    def last_score(self) -> Optional[Tuple[int, float]]:
+        return self._last_score
+
+    @last_score.setter
+    def last_score(self, score: float):
+        self._last_score = (self._step, score)
+
+    def _schedule_args(self, key: KeyT) -> Dict[str, Any]:
```
kadarakos, line 221: It's not really an `Any` in the return type `Dict[str, Any]`, but I dunno if there is a better way of doing it. It could be some sort of data class maybe like a 

```python
@dataclass
class SchedulerArgs:
   key: Any
   last_score: Comparable
   step: int
```

but maybe its enough like this.

thinc/optimizers.py
```diff
@@ -221,28 +238,39 @@ def __call__(
...
         ops = get_array_ops(weights)
         self.nr_update[key] += 1
         nr_upd = self.nr_update[key]
-        if self.L2 != 0 and not self.L2_is_weight_decay:
-            gradient += self.L2 * weights
-        if self.grad_clip:
-            gradient = ops.clip_gradient(gradient, self.grad_clip)
+        schedule_args = self._schedule_args(key)
+
+        if self.L2(**schedule_args) != 0 and not self.L2_is_weight_decay:
```
kadarakos, line 247: Could these `(**schedule_args)` parts be handled in `step_schedules`?

thinc/optimizers.py
danieldk: The type hint is already above in the definition of member variables.

thinc/optimizers.py
danieldk: I typed it as `Any`, because we may want to add other arguments in the future. We have to make this a dict to do the kwarg expansion (I think?).

thinc/optimizers.py
danieldk: I wanted to move as much state as possible out of the optimizer. I like that the optimizer only has the number of steps as the changing state, because it allows someone to change the scheduler externally without making it out of sync with the optimizer.

Besides that, we cannot really cache these states in `step_schedules` because they are now parameter-key dependent, so we have to compute them per parameter key.

thinc/schedules.py
```diff
@@ -1,32 +1,52 @@
...
-from typing import Iterable
+from typing import Dict, Iterable, Optional, Tuple
+from typing_extensions import Protocol, runtime_checkable
 import numpy
 
 from .config import registry
 
 
+@runtime_checkable
+class ScheduleCallable(Protocol):
```
svlandeg, line 10: Alternatively, this definition could go into `thinc.types`

svlandeg (commented):
This will be quite breaking, right? Still, it's a good idea to make this more general, and the code looks good to me 👍

thinc/schedules.py
danieldk: That's a better place, moved.

thinc/optimizers.py
```diff
@@ -189,24 +195,35 @@ def __init__(
 
     def _set_attr_or_schedule(self, name, value):
         if isinstance(value, (float, bool, int)):
+            setattr(self, name, constant(value))
+        elif isinstance(value, list):
+            value = iter(value)
+            setattr(self, name, _generator_schedule(name, value))
+        elif isinstance(value, GeneratorType):
+            setattr(self, name, _generator_schedule(name, value))
```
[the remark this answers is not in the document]
danieldk: When we pass in a generator (like the old schedules). E.g.:

```
def constant(lr):
  while True:
    yield lr
```

By accepting this type, we preserve compatibility with any generators that third-party code might pass to the optimizer for a schedule.

danieldk:
I worked the PR to use instances of the new `Scheduler` class. As discussed internally, the callable and the old generator approaches have the issue that they cannot be pickled. This is a problem, because the schedules can become part of the `Language` object, which may have an associated `Optimizer`. The object-based schedulers can be pickled (also tested with spaCy).

Also rebased the PR to `v9`.

thinc/schedules.py
```diff
@@ -1,38 +1,81 @@
...
+    """Class for implementing Thinc schedules."""
+
+    name: str
+    _schedule: Callable
+    _attrs: Dict[str, Any]
+
+    __slots__ = ["name", "_schedule", "_attrs"]
+
+    def __init__(
+        self, name: str, schedule: Callable, *, attrs: Dict[str, Any] = {}
```
svlandeg, line 20: Is it a worry that `attrs` has a mutable default value here? We can also set it to `None` and rewrite to `{}` within the function.

website/docs/api-schedules.md
```diff
@@ -5,11 +5,94 @@ next: /docs/api-loss
...
+| Argument       | Type                    | Description                                              |
+| -------------- | ----------------------- | -------------------------------------------------------- |
+| `name`         | <tt>str</tt>            | The name of the schedule type.                           |
+| `schedule`     | <tt>Callable</tt>       | Function to compute the schedule value for a given step. |
+| _keyword-only_ |                         |                                                          |
+| `attrs`        | <tt>Dict[str, Any]</tt> | Dictionary of non-parameter attributes.                  |
+
+### Schedule.\_\_call\_\_ {#call tag="method"}
+
+Call the the schedule function, returning the value for the given step. The
```
svlandeg, line 78:
```suggestion
Call the schedule function, returning the value for the given step. The
```

svlandeg (approved):
This is all looking great, and I'm happy the pickling issue got caught early. Nice work!

I only had extreme nitpicks - otherwise feel free to merge.

thinc/schedules.py
danieldk: I did the same as `Model` does, make a copy of the default value in the constructor. I think that should cover it?

PR #805, opened 2022-11-25
Test with numpy==1.24.0.rc1 and newest torch

adrianeboyd:


adrianeboyd:
I reran the tests 4 times with no failures, so this initially seems fine.

PR #808, opened 2022-12-02
Make all layer defaults serializable

adrianeboyd:


adrianeboyd:
Someone should check extra carefully that all the initializer defaults remain unchanged.

svlandeg (commented):
Thanks for catching and fixing this! I went through it in detail and that all looks perfect. 

The only thing left to do is to update `api-layers.md` with the type changes and add "when set to `None`" to the descriptions of the defaults.

danieldk (approved):
Checked that all the initializers line up.

PR #809, opened 2022-12-02
Bring back support for missing labels to legacy cross entropy

danieldk:
The legacy cross-entropy losses were made more strict in v9. However, they are too strict and don't accept missing values anymore.


thinc/legacy/loss.py
```diff
@@ -65,10 +67,10 @@ def convert_truths(self, truths: TruthsT, guesses: Floats2d) -> Tuple[Floats2d,
                         )
                         raise ValueError(msg)
                     for i, value in enumerate(truths):
-                        if not isinstance(value, str):
+                        if not (isinstance(value, str) or value == None):
```
kadarakos, line 70:
```suggestion
                        if not (isinstance(value, str) or value is None):
```
Python religion.

danieldk: We discussed this a bit more and the check should be against `missing_value`.

kadarakos (approved):

PR #810, opened 2022-12-05
Sync `master` into `v9`

danieldk:
This should also fix CI build failures we are seeing in #809 .


svlandeg:
Don't squash merge!

PR #812, opened 2022-12-05
Update CupyOps.asarray to always copy cupy arrays to the current device

frobnitzem:
[cupy.asarray](https://docs.cupy.dev/en/stable/reference/generated/cupy.asarray.html) only performs a copy if necessary.  However, calling this function here is necessary in case the data had a different format or resided on a different GPU.

### DIFF
--- a/thinc/backends/cupy_ops.py
+++ b/thinc/backends/cupy_ops.py
@@ -79,7 +79,7 @@ class CupyOps(Ops):
     def asarray(self, data, dtype=None):
         # We'll try to perform a zero-copy conversion if possible.
         if is_cupy_array(data):
-            array = self.xp.asarray(data, dtype)
+            array = self.xp.asarray(data, dtype=dtype)
         elif is_torch_cuda_array(data):
             array = torch2xp(data)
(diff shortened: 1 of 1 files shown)


adrianeboyd:
Thanks for the PR! The CI didn't run for some reason, so I'm going to close and reopen the PR since it was probably just a fluke.

thinc/backends/cupy_ops.py
```diff
@@ -79,7 +79,7 @@ def gemm(self, x, y, out=None, trans1=False, trans2=False):
     def asarray(self, data, dtype=None):
         # We'll try to perform a zero-copy conversion if possible.
         if is_cupy_array(data):
-            array = data
+            array = self.xp.asarray(data, dtype)
```
adrianeboyd, line 82:
```suggestion
            array = self.xp.asarray(data, dtype=dtype)
```

adrianeboyd:
We don't have any CI infrastructure with more than one GPU, but we'll run the basic tests just to double-check.

danieldk (approved):
Looks good, thanks for the fix!

PR #815, opened 2022-12-07
Premap: Mapping layer from ints to ints

kadarakos:
The `remap_ids` layer is at the moment unused in `spaCy`, but we might want to use a small fixed sized `Embed` for `PREFIX` or `SHAPE` in the future instead of the `HashEmbed`. 

The `remap_ids` implementation, however, is quite slow because it uses `dict.get()`. This allows for flexibility which is nice, but in `spaCy` we only need an `int --> int` map. 

The `premap_ids` layer is similar to `remap_ids` except it only allows for `Ints1d` and `Ints2d` input and a `Mapping[int, int]` as the `mapping_table`. The speedup comes from a small `cython` function that can work with a `MemoryView` for the keys and the result and a `PreshMap` for the map. The `remap_ids_v2` is already running always on the `CPU` even when taking `cupy` arrays as input due to having to convert the `cupy` ints to (which just works with `numpy`).

With the rough speed test in the `examples/benchmarks/mappers` the `premap_ids` layer is 12x-14x faster than `remap_ids_v2` on the machine I tried it on. When used together with `Embed` like `chain(remap_ids_v2, Embed)` vs. `chain(premap, Embed)` the speedup is only 2x. Even though the speed gain is not necessarily super huge in the grand scheme of things I think it makes sense to exploit that we can assume the types of the input,


examples/benchmarks/mappers.py
```diff
@@ -0,0 +1,134 @@
...
+N_columns = 4
+N_dim = 300
+mapper = {}
+numbers = list(range(N_symbols))
+random.shuffle(numbers)
+for v, k in enumerate(numbers):
+    mapper[k] = v
+
+
+def speed_test_no_column():
```
danieldk, line 22: Maybe we should start using a test harness?

thinc/layers/premap_ids.pyx
```diff
@@ -0,0 +1,85 @@
...
+    for i in range(maxi):
+        v = mapping[keys[i]]
+        if v is None:
+            result_view[i] = default
+        else:
+            result_view[i] = v
+    return result
+
+
+@cython.binding(True)
```
adrianeboyd, line 35: Elsewhere we usually add `binding=True` in the header line rather than using individual decorators.

thinc/tests/layers/test_mappers.py
```diff
@@ -0,0 +1,59 @@
+from thinc.layers import premap_ids, remap_ids, remap_ids_v2
+import numpy as np
```
adrianeboyd, line 2: Our code style is to use `cimport numpy as np` and `import numpy` rather than `import numpy as np`:

```suggestion
import numpy
from thinc.layers import premap_ids, remap_ids, remap_ids_v2
```

thinc/tests/layers/test_mappers.py
```diff
@@ -0,0 +1,59 @@
...
+    keys = get_keys()
+    return {int(k): int(v) for v, k in enumerate(keys)}
+
+
+def test_premap():
+    mapper = get_mapper()
+    keys = get_keys()
+    premap = premap_ids(mapper, default=99)
+    values, _ = premap(keys, False)
+    assert all(values.squeeze() == np.asarray(range(len(keys))))
```
adrianeboyd, line 19: Rather than `all(True)` (with a potential for implicit conversions and python comparisons to mask bugs), for these kinds of comparisons in tests I'd prefer to use `numpy.testing.assert_equal` to directly compare the numpy arrays.

thinc/layers/premap_ids.pyx
```diff
@@ -0,0 +1,85 @@
...
+@registry.layers("premap_ids.v1")
+def premap_ids(
+    mapping_table: Mapping[int, int],
+    default: int = 0,
+    *,
+    column: Optional[int] = None
+):
+    """Remap integer inputs to integers a mapping table, usually as a
+    preprocess before embeddings."""
+    if isinstance(mapping_table, PreshMap):
```
adrianeboyd, line 45: Is it possible for `mapping_table` to have the type `PreshMap`?

thinc/tests/layers/test_mappers.py
```diff
@@ -0,0 +1,59 @@
+from thinc.layers import premap_ids, remap_ids, remap_ids_v2
+import numpy as np
+
+
+def get_keys():
```
adrianeboyd, line 5: Turn `keys` and `mapper` into pytest fixtures.

thinc/layers/premap_ids.pyx
```diff
@@ -0,0 +1,85 @@
+# cython: infer_types=True
+cimport cython
+import numpy as np
```
adrianeboyd, line 3:
```suggestion
import numpy
```

adrianeboyd (commented):
Mostly just style-related comments for the tests...

thinc/layers/premap_ids.pyx
kadarakos: good idea, i just hate these headers because they can silently fail without syntax error or anything.

thinc/layers/premap_ids.pyx
kadarakos: hmm okay. What I say in the type declaration is that it should be any `Mapping[int, int]` because all we need is for the `mapping_table` to implement `items()` and for that yield `Tuple[int, int]`. So I was thinking that users could give us any map like that including `PreshMap`. But I did realize now `PreshMap` is actually not a `Mapping`.

This might be a stupid question, but should be make `PreshMap` a `Mapping`? If not then I can make `mapping_table` type into `Union[Mapping[int, int], PreshMap]`.

adrianeboyd: You could consider making `PreshMap` a `Mapping`, but I think it would be pretty low priority and might be more work than you'd expect. (`PreshMap` is kind of fiddly and I don't think `__len__` is correct right now, so things could go wrong fast.)

adrianeboyd: I think for now I'd just keep it as `Mapping`. (I'd actually never initialized a `PreshMap` in python before testing this just now.)

thinc/tests/layers/test_mappers.py
```diff
@@ -0,0 +1,57 @@
...
+    assert (values2 == values3).all()
+
+
+def test_column(keys, mapper):
+    idx = numpy.zeros((len(keys), 4), dtype="int")
+    idx[:, 3] = keys
+    remap_v2 = remap_ids_v2(mapper, column=3)
+    premap = premap_ids(mapper, column=3)
+    assert all(remap_v2(idx, False)[0].squeeze() == numpy.asarray(range(len(keys))))
+    assert all(premap(idx, False)[0].squeeze() == numpy.asarray(range(len(keys))))
```
adrianeboyd, line 57: Also `assert_equal` in these cases?

thinc/layers/premap_ids.pyx
```diff
@@ -0,0 +1,83 @@
+# cython: binding=True, infer_types=True
+cimport cython
+import numpy
+from preshed.maps cimport PreshMap
+from preshed.maps import PreshMap
```
adrianeboyd, line 5:
```suggestion
```

examples/benchmarks/mappers.py
kadarakos: not this PR right? can I resolve this and add a ticket?

thinc/tests/layers/test_layers_api.py
```diff
@@ -130,7 +130,8 @@ def assert_data_match(Y, out_data):
     ("SparseLinear.v1", {}, (numpy.asarray([1, 2, 3], dtype="uint64"), array1d, numpy.asarray([1, 1], dtype="i")), array2d),
     ("SparseLinear.v2", {}, (numpy.asarray([1, 2, 3], dtype="uint64"), array1d, numpy.asarray([1, 1], dtype="i")), array2d),
     ("remap_ids.v1", {"dtype": "f"}, ["a", 1, 5.0], array2dint),
-    ("remap_ids.v2", {"mapping_table": {}, "column": 1}, numpy.array([[1, 2, 3], [4, 5, 6]]).T, array2dint)
+    ("remap_ids.v2", {"mapping_table": {}, "column": 1}, numpy.array([[1, 2, 3], [4, 5, 6]]).T, array2dint),
+    ("premap_ids.v1", {"mapping_table": {}, "column": 1}, numpy.array([[1, 2, 3], [4, 5, 6]]).T, array2dint)
```
adrianeboyd, line 134:
```suggestion
    ("premap_ids.v1", {"mapping_table": {}, "column": 1}, numpy.array([[1, 2, 3], [4, 5, 6]]).T, array2dint),
```

examples/benchmarks/mappers.py
danieldk: Yeah, let's do that in a follow-up PR. Though the timings here could use some cleanups. E.g., some of the constructed `start` instances are not used in the tests below. Two other tips:

* It's a very delicate issue, but it's usually better to use `perf_counter` over `process_time`, see e.g.: https://asv.readthedocs.io/en/stable/writing_benchmarks.html#benchmark-types `perf_counter` is also used by e.g. `timeit`.
* I like to time using a context manager. It's cleaner than all the manual timings + computations, see: https://github.com/explosion/spaCy/blob/4c60afb946f35e2675a5e21880ca3a09633d0bfa/spacy/cli/benchmark_speed.py#L60

kadarakos:
Ugh, i think I shouldn't have merged the master into this?

adrianeboyd:
If you need to force push to revert to a few commits ago that would be fine because there aren't any intervening reviewing comments.

adrianeboyd:
You lost my trailing comma at the very least?

thinc/util.py
```diff
@@ -583,6 +582,17 @@ def use_nvtx_range(message: str, id_color: int = -1):
...
+    """Register the running time of a context."""
+
+    def __enter__(self):
+        self.start = time.perf_counter()
+        return self
+
+    def __exit__(self, type, value, traceback):
+        self.elapsed = time.perf_counter() - self.start
+
+
```
adrianeboyd, line 595: Is this something that we need within `thinc` itself?

thinc/util.py
```diff
@@ -605,6 +615,7 @@ def check_consistency(self, arr: ArrayXd):
             )
 
 
+
```
adrianeboyd, line 618: It looks like this file needs to be reformatted.

thinc/util.py
kadarakos: couldn't it be useful sometimes? Would you prefer it in the speed-test file instead?

adrianeboyd: Putting it in `thinc.util` means we're kind of committing to maintain this in the future as part of `thinc`, so I think it would be better to just have it in the script for now. Possibly in `thinc.tests.util` in the future if it's actually used in the test suite, but for now I think it's better in the script.

adrianeboyd:
Can you remove all the changes to `thinc.util`?

adrianeboyd:
Looks good!

PR #820, opened 2022-12-13
Make resizable layer work with textcat and transformers

polm:
This PR is an attempted fix of the issue outlined in https://github.com/explosion/spaCy/issues/11968. I don't completely understand how the resizable textcat is supposed to work yet, and am not sure if this is the right approach, but locally it got rid of errors and I was able to train a model with reasonable performance.

There are also no tests yet.

### DIFF
--- a/thinc/layers/resizable.py
+++ b/thinc/layers/resizable.py
@@ -54,10 +54,13 @@ def resize_linear_weighted(
     assert not layer.shims
 
     # return the original layer if it wasn't initialized or if nO didn't change
-    if layer.has_dim("nO") is None or layer.has_dim("nI") is None:
-        layer.set_dim("nO", new_nO, force=True)
+    if layer.has_dim("nO") is None:
+        layer.set_dim("nO", new_nO)
         return layer
     elif new_nO == layer.get_dim("nO"):
+        return layer
+    elif layer.has_dim("nI") is None:
+        layer.set_dim("nO", new_nO, force=True)
         return layer
 
     dims = {name: layer.maybe_get_dim(name) for name in layer.dim_names}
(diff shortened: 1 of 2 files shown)


svlandeg:
Can you provide a minimal working example that we can use as test case for spaCy?

polm:
Made a failing test: https://github.com/explosion/spacy-transformers/pull/357. For testing will a realistic pipeline, you can use the goemotions sample project with transformers and just change the architecture to v2.

polm:
I would like someone more familiar with the resizable layer to look at this, but I believe the current fix is correct. To clarify what's going on:

**Before this PR**, the layer is assumed to be initialized if `nO` is set. This causes an error in the TextCatCNN when used with transformers, because the number of classes (`nO`) is known (or at least temporarily assumed) when the textcat component is created, but the size of the transformer embedding (`nI`) is not known until later. 

**After this PR**, the resizable layer is considered uninitialized if either `nO` **or** `nI` is unset. Because `nO` may already be set in that case, the new `nO` is set using `force=True`.

thinc/layers/resizable.py
```diff
@@ -59,6 +59,9 @@ def resize_linear_weighted(
         return layer
     elif new_nO == layer.get_dim("nO"):
         return layer
+    elif layer.has_dim("nI") is None:
+        layer.set_dim("nO", new_nO, force=True)
+        return layer
```
svlandeg, line 64: Nice find, Paul. Seems correct to me!

svlandeg (approved):
This should be good to merge and will make the test in https://github.com/explosion/spacy-transformers/pull/357 green.

PR #821, opened 2022-12-13
`with_flatten`: make the layer input/output types symmetric

danieldk:
The idea of the `with_flatten` layer is that it flattens a nested sequence, passes it to the wrapped layer, and then unflattens the output of the wrapped layer.

However, the layer was asymmetric in that it passes a list to the wrapped layer, but expects back an XP array. This breaks composition with other Thinc layers, such as `with_array`.

This change makes `with_flatten` symmetric, in that the inputs/outputs of the `with_flatten` and the wrapped layer are symmetric.

It seems that this layer is not used in Thinc or spaCy, so maybe it never worked correctly? At any rate, I needed to flatten a nested list in distillation with `with_flatten(with_array(...))` in distillation and found that it doesn't actually work.

### DIFF
--- a/thinc/layers/with_flatten.py
+++ b/thinc/layers/with_flatten.py
@@ -2,31 +2,32 @@ from typing import Tuple, Callable, Sequence, Any, cast, TypeVar, Optional, List
 
 from ..model import Model
 from ..config import registry
-from ..types import ListXd
-
+from ..types import ArrayXd, ListXd
 
 ItemT = TypeVar("ItemT")
 InT = Sequence[Sequence[ItemT]]
-OutT = TypeVar("OutT", bound=ListXd)
+OutT = ListXd
+InnerInT = Sequence[ItemT]
+InnerOutT = ArrayXd
 
 
 @registry.layers("with_flatten.v1")
-def with_flatten(layer: Model[InT, InT]) -> Model[OutT, OutT]:
+def with_flatten(layer: Model[InnerInT[ItemT], InnerOutT]) -> Model[InT[ItemT], OutT]:
     return Model(f"with_flatten({layer.name})", forward, layers=[layer], init=init)
 
 
 def forward(
-    model: Model[OutT, OutT], Xnest: OutT, is_train: bool
+    model: Model[InT, OutT], Xnest: InT, is_train: bool
 ) -> Tuple[OutT, Callable]:
-    layer: Model[InT, InT] = model.layers[0]
-    Xflat: Sequence[Sequence[Any]] = _flatten(Xnest)
+    layer: Model[InnerInT, InnerOutT] = model.layers[0]
+    Xflat = _flatten(Xnest)
     Yflat, backprop_layer = layer(Xflat, is_train)
     # Get the split points. We want n-1 splits for n items.
     arr = layer.ops.asarray1i([len(x) for x in Xnest[:-1]])
     splits = arr.cumsum()
     Ynest = layer.ops.xp.split(Yflat, splits, axis=0)
 
-    def backprop(dYnest: OutT) -> OutT:
+    def backprop(dYnest: OutT) -> InT:
         dYflat = model.ops.flatten(dYnest)  # type: ignore[arg-type, var-annotated]
         # type ignore necessary for older versions of Mypy/Pydantic
         dXflat = backprop_layer(dYflat)
@@ -36,7 +37,7 @@ def forward(
     return Ynest, backprop
 
 
-def _flatten(nested: OutT) -> InT:
+def _flatten(nested: InT) -> InnerInT:
     flat: List = []
     for item in nested:
         flat.extend(item)
@@ -44,7 +45,7 @@ def _flatten(nested: OutT) -> InT:
 
 
 def init(
-    model: Model[OutT, OutT], X: Optional[OutT] = None, Y: Optional[OutT] = None
+    model: Model[InT, OutT], X: Optional[InT] = None, Y: Optional[OutT] = None
 ) -> None:
     model.layers[0].initialize(
         _flatten(X) if X is not None else None,
--- a/thinc/layers/with_flatten.py
+++ b/thinc/layers/with_flatten.py
@@ -2,64 +2,49 @@ from typing import Tuple, Callable, Sequence, Any, cast, TypeVar, Optional, List
 
 from ..model import Model
 from ..config import registry
+from ..types import ListXd
 
 
-InItemT = TypeVar("InItemT")
-OutItemT = TypeVar("OutItemT")
 ItemT = TypeVar("ItemT")
-
-NestedT = List[List[ItemT]]
-FlatT = List[ItemT]
+InT = Sequence[Sequence[ItemT]]
+OutT = TypeVar("OutT", bound=ListXd)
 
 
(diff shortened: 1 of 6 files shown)


shadeMe (approved):

thinc/layers/with_flatten.py
```diff
@@ -2,49 +2,64 @@
...
-from ..types import ListXd
 
 
+InItemT = TypeVar("InItemT")
+OutItemT = TypeVar("OutItemT")
 ItemT = TypeVar("ItemT")
-InT = Sequence[Sequence[ItemT]]
-OutT = TypeVar("OutT", bound=ListXd)
+
+NestedT = List[List[ItemT]]
```
shadeMe, line 11: Missed this one:
```suggestion
NestedT = Sequence[Sequence[ItemT]]
```

danieldk: Right, I used to have `NestedT` and `NestedSeqT` aliases, but I simplified it and forgot to update the docs.

danieldk: Ok, fixed in the docs.

thinc/layers/with_flatten.py
```diff
@@ -2,49 +2,64 @@
...
+FlatT = List[ItemT]
 
 
 @registry.layers("with_flatten.v1")
```
adrianeboyd, line 15: I doubt anyone was using this, but to be on the safe side I think this should be `with_flatten.v2`.

danieldk: I'll do that, but I think that we could then have the docs only describe v2 since the documentation was not accurate for v1 anyway?

danieldk: Added back, also tried to fix the typing of the old layer as much as possible, since they didn't reflect what was happening in the old layer.

adrianeboyd: I would be tempted to just leave it completely alone, since it doesn't sound like anyone would want to use it?

adrianeboyd: Hmm, I don't really know what do about the docs now, though, since `with_flatten_v2` looks kind of awful.

danieldk: Do you mean the name? We already have `v2`s of the sparse linear and softmax layers.

website/docs/api-layers.md
```diff
@@ -1409,15 +1409,44 @@ https://github.com/explosion/thinc/blob/master/thinc/layers/with_array.py
...
+| Argument    | Type                                            | Description        |
+| ----------- | ----------------------------------------------- | ------------------ |
+| `layer`     | <tt>Model[Sequence[Any], ArrayXd]</tt>          | The layer to wrap. |
+| **RETURNS** | <tt>Model[Sequence[Sequence[Any]], ListXd]</tt> | The wrapped layer. |
 
 ```python
 https://github.com/explosion/thinc/blob/master/thinc/layers/with_flatten.py
 ```
 
+### with_flatten_v2 {#with_flatten_v2 tag="function" new="8.1.6"}
```
adrianeboyd, line 1429: I mean `with_flatten_v2` here is weird as the reference. The registered `with_flatten.v2` is fine/normal, but do these descriptions in the docs use the `_v2` names?

thinc/layers/with_flatten.py
adrianeboyd: I guess there's no good way around it.

website/docs/api-layers.md
danieldk: Outside the scope of this PR, but wouldn’t we want downstream users to get these functions through the registry, because then we can move these functions more freely (like into a future ‘spacy-legacy’).

website/docs/api-layers.md
```diff
@@ -1409,15 +1409,44 @@ https://github.com/explosion/thinc/blob/master/thinc/layers/with_array.py
...
 
-| Argument    | Type                                                             | Description        |
-| ----------- | ---------------------------------------------------------------- | ------------------ |
-| `layer`     | <tt>Model[Sequence[Sequence[Any]], Sequence[Sequence[Any]]]</tt> | The layer to wrap. |
-| **RETURNS** | <tt>Model[ListXd, ListXd]</tt>                                   | The wrapped layer. |
+<infobox variant="warning">
+
+Even though `with_flatten` is a layer wrapper, it does not preserve symmetry
+between the input and output data types. This often makes it hard to compose
+with other layers. [`with_flatten_v2`](#with_flatten_v2) instead.
```
adrianeboyd, line 1416:
```suggestion
with other layers. Use [`with_flatten_v2`](#with_flatten_v2) instead.
```

website/docs/api-layers.md
adrianeboyd: We do want them to use the registry, so that's part of why I think it's weird that the API docs are presented like this. The registry names aren't even included anywhere.

danieldk: Yeah, I'll make a ticket for this.

thinc/layers/with_flatten_v2.py
```diff
@@ -0,0 +1,67 @@
...
+def forward(
+    model: Model[NestedT[InItemT], NestedT[OutItemT]],
+    Xnest: NestedT[InItemT],
+    is_train: bool,
+) -> Tuple[NestedT[OutItemT], Callable]:
+    layer: Model[FlatT[InItemT], FlatT[OutItemT]] = model.layers[0]
+    Xflat = _flatten(Xnest)
+    Yflat, backprop_layer = layer(Xflat, is_train)
+    # Get the split points. We want n-1 splits for n items.
+    lens = [len(x) for x in Xnest]
```
kadarakos, line 31: Could the `_flatten` return the `lens` to not do the for loop over `Xnest` twice?

kadarakos (commented):
Maybe its me, but since `with_flatten` is in a deep learning library my initial assumption was that it will call `xp.flatten` on an `ArrayXd` making it `Array1d` and it will reshape it back. Should it dynamically either flatten a list or a tensor? Should it be able to flatten any nesting level with lists just like any order tensor in `xp.flatten`? Or we just have it like this and add it into the documentation that this is flattening lists? Or is it just me who made the assumption? Don't want to derail the PR just mentioning that `flatten` for me invoked something different.

thinc/layers/with_flatten_v2.py
danieldk: Done!

danieldk:
> Maybe its me, but since `with_flatten` is in a deep learning library my initial assumption was that it will call `xp.flatten` on an `ArrayXd` making it `Array1d` and it will reshape it back. Should it dynamically either flatten a list or a tensor? Should it be able to flatten any nesting level with lists just like any order tensor in `xp.flatten`? Or we just have it like this and add it into the documentation that this is flattening lists of lists (one level only)? Or is it just me who made the assumption? Don't want to derail the PR just mentioning that `flatten` for me invoked something different.

These are good questions, and I agree that our naming it not really in line with other deep learning libraries, though it is kinda the normal [FP definition](https://v2.ocaml.org/api/List.html)

At any rate, this PR is just for fixing the existing implementation which just flattens one level of lists. If we want something more advanced, it should probably have a different name to avoid confusion (and a separate PR).

PR #828, opened 2022-12-22
Remove TensorFlow/MXNet code and import it from `thinc_addons` instead

shadeMe:
Leaving this in draft mode until we pick the name for the addons package.


shadeMe:
After internal discussions, we've decided to not go ahead with this refactoring.

PR #830, opened 2022-12-26
Smooth one hot fix

kadarakos:
The `smooth_one_hot` function in `thinc.util` is being called in the `CategoricalCrossentropy` to apply label smoothing. This PR fixes the check for valid label-smoothing parameter, which was fixed before for `to_categorical`, but not for `smooth_one_hot`. Also added a test for `smooth_one_hot` which was also missing.

### DIFF
--- a/thinc/util.py
+++ b/thinc/util.py
@@ -257,12 +257,17 @@ def smooth_one_hot(X: Floats2d, label_smoothing: float) -> Floats2d:
     """
     Apply label-smoothing to one-hot array.
     """
+    n_classes = X.shape[1]
+    max_smooth = (n_classes - 1) / n_classes
     if label_smoothing < 0.0:
         raise ValueError(
             "Label-smoothing parameter has to be greater than or equal to 0"
         )
-    n_classes = X.shape[1]
-    max_smooth = (n_classes - 1) / n_classes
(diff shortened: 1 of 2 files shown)


svlandeg:
Opening and closing to run the unit tests

thinc/util.py
```diff
@@ -257,14 +257,20 @@ def smooth_one_hot(X: Floats2d, label_smoothing: float) -> Floats2d:
...
     Apply label-smoothing to one-hot array.
     """
-    if not 0.0 <= label_smoothing < 0.5:
+    if label_smoothing < 0.0:
+        raise ValueError(
+            "Label-smoothing parameter has to be greater than or equal to 0"
+        )
+    n_classes = X.shape[1]
+    max_smooth = (n_classes - 1) / n_classes
+    if label_smoothing >= max_smooth:
```
danieldk, line 266: I think like `to_categorical`, we also need a check here to ensure that the number of classes is larger than one. Otherwise, `max_smooth == 0.0` and even with label_smoothing==0.0, we'll get the strange error message _For 1 classes label_smoothing parameter has to be less than 0, but found 0._

danieldk (approved):

PR #832, opened 2022-12-28
Add pull request and bug report templates

shadeMe:


shadeMe:
The Netlify errors are spurious as the new Markdown files are only pertinent to our GitHub workflow. So, we can ignore those failures.

.github/ISSUE_TEMPLATE/01_bugs.md
```diff
@@ -0,0 +1,19 @@
+---
+name: "\U0001F6A8 Submit a Bug Report"
+about: Use this template if you came across a bug or unexpected behaviour differing from the docs.
+---
+
+<!-- NOTE: For questions or install related issues, please open a Discussion instead. -->
```
svlandeg, line 6: It should be made clear that they need to open a Discussion thread on spaCy though, right?
I see that we actually do have a "Discussions" tab, but were we planning on actually using that?

shadeMe: Ah, good point. Honestly, I'm not sure what would be best going forward. Using the spaCy Discussions to handle Thinc-specific issues seems like overkill at the moment - We only get infrequent issues in this repo.

We could just remove that comment for the time being and reassess once/if we start seeing more traffic.

svlandeg: That's fine for me as well, like you said the volume is not that high on Thinc.

svlandeg (approved):
Thanks!

PR #834, opened 2023-01-03
Fix cython directive `language_level`

shadchin:
`-3` use as command line argument


polm:
Thanks for catching that, that does look like a simple mistake!

PR #835, opened 2023-01-09
Fixes SASS error

essenmitsosse:
## Description

This is a super annoying breaking change in SASS (it also happens in `node-sass`/`node-scss`), but there seems to be no way around it.

context: https://sass-lang.com/documentation/breaking-changes/slash-div

### Types of change

- Website build

## Checklist


svlandeg (approved):
Thanks!

PR #848, opened 2023-02-07
Extend mypy to v1.0.x

adrianeboyd:
## Description

Extend mypy to v1.0.x

### Types of change

?

## Checklist


adrianeboyd:
I don't know if we want to add a mypy extra or something, because the package doesn't currently express which versions of mypy it's compatible with.

PR #860, opened 2023-03-10
Implement `pad` as a CUDA kernel

danieldk:
## Description

`Ops.pad` was a fairly slow operation on GPU. It iterates over all sequences and copies each sequence into the padded array. This results in a lot of kernel launches. In the biaffine parser, padding the inputs was more costly than applying the biaffine layers.

This change optimizes the `pad` op using a custom CUDA kernel. The kernel get an array of pointers to the CuPy arrays that are provided as a list. The output array is then filled, parallelizing over the 'time steps'. This should provides the largest amount of parallelism, since we usually have n_steps * hidden_size to parallelize over.

Before:

before-padding-change

After:

padding-after-change

~**Warning:** please do not review yet! There are still some todo items and I still want to be able to rebase the branch.~

- [x] Extend to int32 and int64, so that we can use this optimization in curated transformers as well.
- [x] Test by training in an actual model.
- [x] Rebase to Thinc `master`.
- [x] Clean up some redundancies in the `Ops.pad` code.

## Checklist

### DIFF
--- a/thinc/backends/_custom_kernels.py
+++ b/thinc/backends/_custom_kernels.py
@@ -1,4 +1,7 @@
 from typing import Optional, Tuple
+from functools import reduce
+import numpy
+import operator
 import re
 from pathlib import Path
 from collections import defaultdict
@@ -44,6 +47,10 @@ KERNELS_LIST = [
     "maxout<float>",
     "mish<double>",
     "mish<float>",
+    "pad<double>",
+    "pad<float>",
+    "pad<int>",
+    "pad<long long>",
     "reduce_max<double>",
     "reduce_max<float>",
     "reduce_sum<double>",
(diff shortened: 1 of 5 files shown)


thinc/backends/_custom_kernels.py
```diff
@@ -146,6 +157,61 @@ def _alloc_like(array, zeros: bool = True):
         return cupy.empty_like(array)
 
 
+def pad(seqs, round_to=1, *, threads_per_block=128, num_blocks=128):
+    for seq in seqs:
+        _is_float_or_int_array(seq)
+
+    lengths = [len(seq) for seq in seqs]
+    length = max(lengths)
```
shadeMe, line 165: Can we rename these two to `seq_lens` and `max_seq_len` (or something similar) to avoid confusion?

thinc/backends/_custom_kernels.py
```diff
@@ -146,6 +157,61 @@ def _alloc_like(array, zeros: bool = True):
...
+    length = (length + (round_to - 1)) // round_to * round_to
```
shadeMe, line 166: Could we have a comment here explaining the need for the rounding?

thinc/backends/_custom_kernels.py
```diff
@@ -93,6 +100,10 @@ def compile_mmh(src):
 maxout_kernel_double = _get_kernel("maxout<double>")
 mish_kernel_float = _get_kernel("mish<float>")
 mish_kernel_double = _get_kernel("mish<double>")
+pad_kernel_float = _get_kernel("pad<float>")
```
shadeMe, line 103: We should get #870 merged into this PR.

thinc/backends/_custom_kernels.py
danieldk: Done.

thinc/backends/_custom_kernels.py
danieldk: Merged.

thinc/backends/_custom_kernels.py
danieldk: I have added a comment. Also simplified the computation with modular arithmetic and added some tests with rounding.

shadeMe (approved):

shadeMe:

PR #861, opened 2023-03-13
Avoid h2d - d2h roundtrip when using `unflatten`

danieldk:
## Description

`unflatten` converts its `lengths` argument to a NumPy array, because CuPy's `split` function requires lengths to be in CPU memory. However, in various places in Thinc, we copy the lengths array to GPU memory when CupyOps is used. This results in an unnecessary roundtrip of the lengths array (host to device -> device to host). One of these roundtrips (array `list2array`) showed up in profiles of the biaffine parser.

This change fixes some length array allocations to avoid the round trip.

### Types of change

Performance improvement

## Checklist

### DIFF
--- a/thinc/layers/with_ragged.py
+++ b/thinc/layers/with_ragged.py
@@ -1,8 +1,13 @@
 from typing import Tuple, Callable, Optional, TypeVar, cast, List, Union
 
+from ..backends import NumpyOps
 from ..types import Padded, Ragged, Array2d, ListXd, List2d, Ints1d
 from ..model import Model
 from ..config import registry
+
+
+NUMPY_OPS = NumpyOps()
+
 
 RaggedData = Tuple[Array2d, Ints1d]
 SeqT = TypeVar("SeqT", bound=Union[Padded, Ragged, ListXd, RaggedData])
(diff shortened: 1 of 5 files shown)


thinc/layers/with_ragged.py
```diff
@@ -90,8 +95,8 @@ def _padded_forward(
     Xs = padded2list(Xp)
     # Bit annoying here: padded is in a different order, so we need to make new
     # lengths.
-    lengths = layer.ops.asarray1i([len(x) for x in Xs])
-    Yr, get_dXr = layer(Ragged(flatten(Xs), lengths), is_train)
+    lengths = NUMPY_OPS.asarray1i([len(x) for x in Xs])
+    Yr, get_dXr = layer(Ragged(flatten(Xs), layer.ops.asarray1i(lengths)), is_train)
```
svlandeg, line 99: Maybe add a quick comment here to avoid confusion in future readers of the code?

svlandeg (approved):
Nice find!

thinc/layers/with_ragged.py
danieldk: Added a comment.

shadeMe (approved):

PR #862, opened 2023-03-13
make has_ predicated private

kadarakos:
Make the `has_` predicates private according to the corresponding `v9` ticket.

## Description

Renames the `has_*` predicates `_has_*` to show that they are supposed to be private. The exceptions are `has_cupy` because that was exported in `thinc.api` before and `util.py` exports `_has_torch` as `has_torch` in `__all__`, because it used to do that before? Not sure about the second one.

### Types of change

Small technical debt mitigation.

## Checklist

### DIFF
--- a/thinc/backends/_custom_kernels.py
+++ b/thinc/backends/_custom_kernels.py
@@ -2,7 +2,7 @@ from typing import Optional, Tuple
 import re
 from pathlib import Path
 from collections import defaultdict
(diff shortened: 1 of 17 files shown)


danieldk:
We discussed this out-of-band and agreed to close this issue, since it has turned out to be quite convenient to have these `has_` booleans to probe Thinc's world view.

PR #863, opened 2023-03-14
Improve exception when CuPy/PyTorch MPS is not installed

danieldk:
## Description

`require_gpu` does not distinguish between different failure scenarios, which makes it harder to debug why GPU support is not working. See e.g.

https://github.com/explosion/spaCy/discussions/12410

Rather than raising a generic `No GPU devices can be detected` when CuPy or PyTorch with MPS isn't installed, this change adds more specific errors.

### Types of change

Enhancement

## Checklist

### DIFF
--- a/thinc/compat.py
+++ b/thinc/compat.py
@@ -32,6 +32,7 @@ try:  # pragma: no cover
 
     has_torch = True
     has_torch_cuda_gpu = torch.cuda.device_count() != 0
+    has_torch_mps = hasattr(torch, "has_mps") and torch.backends.mps.is_built()
     has_torch_mps_gpu = (
         hasattr(torch, "has_mps")
         and torch.has_mps  # type: ignore[attr-defined]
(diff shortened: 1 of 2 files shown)


thinc/compat.py
```diff
@@ -32,6 +32,7 @@
 
     has_torch = True
     has_torch_cuda_gpu = torch.cuda.device_count() != 0
+    has_torch_mps = hasattr(torch, "has_mps") and torch.backends.mps.is_built()
```
shadeMe, line 35: Why do we need a distinction between `has_torch_mps` and `has_torch_mps_gpu`? Also, don't we need to use [`torch.backends.mps.is_available()`](https://pytorch.org/docs/stable/backends.html#module-torch.backends.mps) in conjunction with `is_built()`?

On a related note, since the `has_mps` attribute appears to be undocumented, don't we want to stick to the functions in `torch.backends.mps` instead?

danieldk: Fixed.

shadeMe (approved):

PR #870, opened 2023-04-03
Add `LazyKernel` for lazily-compiled custom `cupy` kernels

shadeMe:
## Description

Add a wrapper around `cupy.RawKernel` that lazily compiles them on first invocation.

This prevents CuPy from allocating memory unnecessarily during module init.

### Types of change

Refactor

## Checklist

### DIFF
--- a/thinc/backends/_custom_kernels.py
+++ b/thinc/backends/_custom_kernels.py
@@ -66,8 +66,8 @@ class LazyKernel:
     """Wraps around `cupy.RawModule` to verify and lazily compile a `cupy.RawKernel` on first call."""
 
     name: str
-    _kernel: Optional[cupy.RawKernel]
-    _compile_callback: Optional[Callable[[], cupy.RawKernel]]
+    _kernel: Optional["cupy.RawKernel"]
+    _compile_callback: Optional[Callable[[], "cupy.RawKernel"]]
     _compiled: bool
 
     __slots__ = ["name", "_kernel", "_compile_callback", "_compiled"]
@@ -76,7 +76,7 @@ class LazyKernel:
         self,
         name: str,
         *,
-        compile_callback: Optional[Callable[[], cupy.RawKernel]] = None,
+        compile_callback: Optional[Callable[[], "cupy.RawKernel"]] = None,
     ) -> None:
         self.name = name
         self._kernel = None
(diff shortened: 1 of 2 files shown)


shadeMe:

thinc/backends/_custom_kernels.py
```diff
@@ -62,74 +62,106 @@
...
+
+    def _compile_kernel(self):
+        if self._compiled:
+            return
+
+        if self._compile_callback is not None:
+            self._kernel = self._compile_callback()
+        elif KERNELS is not None:
+            self._kernel = KERNELS.get_function(self.name)
+        self._compiled = True
```
danieldk, line 106: We always have `KERNELS` I think? So both code paths should return something (or raise an exception). I am wondering if we can't do without `_compiled` and signify that `self._kernel is None` means that the kernel is not yet compiled. But I might definitely be overlooking something here.

danieldk (commented):
Nice improvement! I wonder if we lost lazy compilation somewhere along the way or I just misremember (I seem to remember that I could see kernels getting added to the compilation cache as they were used).

Two small comments/questions.

thinc/backends/_custom_kernels.py
shadeMe: Looking at the initialization code, it seems like `KERNELS` is `None` when `has_cupy_gpu` is `False`. But it is true that `CupyOps` won't ever be called in that case, so I think we can get rid of the flag.

thinc/backends/_custom_kernels.py
```diff
@@ -62,74 +62,106 @@
...
-backprop_dish_kernel_float = _get_kernel("backprop_dish<float>")
-backprop_gelu_kernel_double = _get_kernel("backprop_gelu<double>")
-backprop_gelu_kernel_float = _get_kernel("backprop_gelu<float>")
-backprop_hard_swish_kernel_double = _get_kernel("backprop_hard_swish<double>")
-backprop_hard_swish_kernel_float = _get_kernel("backprop_hard_swish<float>")
-backprop_hard_swish_mobilenet_kernel_double = _get_kernel(
+    return cupy.RawKernel((PWD / "_murmur3.cu").read_text(encoding="utf8"), "hash_data")
+
+
+clipped_linear_kernel_float = LazyKernel("clipped_linear<float>")
```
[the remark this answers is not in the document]
shadeMe: Done.

shadeMe:

danieldk (approved):

PR #875, opened 2023-04-28
Switch to preshed v4

adrianeboyd:
## Description

Switch to preshed v4.

### Types of change

Enhancement?

## Checklist


adrianeboyd:
The idea here was to release the next version of thinc v9.0.0.devN so that we can use this in spacy v4, however preshed v4.0.0 seems to fail immediately in the tokenizer cache, so we'll put this on hold until we figure out what's going on.

PR #882, opened 2023-06-27
Support zero-length batches and hidden sizes in reduce_{max,mean,sum}

danieldk:
## Description

Before this change we would fail with an assertion, but it is valid to do reductions over zero-length arrays.

(As long as the length of a sequence is not zero in the case of max and mean, but we check for that separately.)

### Types of change

Bugfix

## Checklist

### DIFF
--- a/thinc/backends/numpy_ops.pyx
+++ b/thinc/backends/numpy_ops.pyx
@@ -317,8 +317,11 @@ class NumpyOps(Ops):
         cdef int O = X.shape[1]
         cdef int T = X.shape[0]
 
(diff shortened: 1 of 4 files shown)


danieldk:
* Updated tests to cover all cases of zero-length sequences and dimensions.
* Fix handling of zero dimensions in `reduce_first` and `reduce_last`.
* Make it clearer in the documentation what inputs are accepted/rejected.

svlandeg (approved):
Thanks Daniël!

PR #890, opened 2023-07-26
Import mxnet and tensorflow only if explicitly enabled

adrianeboyd:
## Description

Import `mxnet` and `tensorflow` only if explicitly enabled

### Types of change

?

## Checklist

### DIFF
--- a/thinc/tests/enable_tensorflow.py
+++ b/thinc/tests/enable_tensorflow.py
@@ -1,0 +1,3 @@
+from thinc.compat import enable_tensorflow
+
+enable_tensorflow()
--- a/thinc/tests/enable_tensorflow.py
+++ b/thinc/tests/enable_tensorflow.py
@@ -1,3 +1,6 @@
 from thinc.compat import enable_tensorflow
 
-enable_tensorflow()
+try:
+    enable_tensorflow()
+except ImportError:
+    pass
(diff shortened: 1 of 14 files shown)


adrianeboyd:
This was the most minimal way I could think of to do this for thinc v8.

Instructions and examples including the `enable_*` still need to be added somewhere more explicitly to the docs.

I mainly think that nearly no one is using thinc with mxnet or tensorflow so this isn't going to be very breaking either way.

I don't know how/if you could add deprecation warnings to v8.1 to prepare users for the change. Again, I think it's hardly anyone. I doubt many people are using this ancient version of tensorflow and I guess I don't really know, but I don't think mxnet ever had a large user base.

thinc/tests/enable_tensorflow.py
```diff
@@ -0,0 +1,6 @@
+from thinc.compat import enable_tensorflow
+
+try:
+    enable_tensorflow()
+except ImportError:
+    pass
```
rmitsch, line 6: Why are we passing on `ImportError`?

adrianeboyd: To get the same behavior as before in the CI without more changes because we can only install mxnet and tensorflow for a subset of OS/python versions (and not the same subset). These files could be renamed to `maybe_enable_tensforflow` or similar.

danieldk (commented):
We should probably decide if we want to remove this functionality from Thinc 9 and if so, add warnings to the `enable_` functions.

rmitsch (approved):
LGTM. Maybe @danieldk wants to give it a final once-over?

danieldk (approved):

PR #897, opened 2023-08-10
Preserve values with dtype for NumpyOps/CupyOps.asarray

adrianeboyd:
## Description

Always specify `dtype` when creating new arrays so that large integer values are preserved and not at risk of going through an intermediate `float64` conversion.

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/thinc/backends/cupy_ops.py
+++ b/thinc/backends/cupy_ops.py
@@ -93,10 +93,8 @@ class CupyOps(Ops):
             array = tensorflow2xp(data)
         elif is_mxnet_gpu_array(data):
             array = mxnet2xp(data)
-        elif dtype is not None:
-            array = self.xp.array(data, dtype=dtype)
         else:
-            array = self.xp.array(data)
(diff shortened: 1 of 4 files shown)


thinc/backends/cupy_ops.py
```diff
@@ -93,6 +93,8 @@ def asarray(self, data, dtype=None):
             array = tensorflow2xp(data)
         elif is_mxnet_gpu_array(data):
             array = mxnet2xp(data)
+        elif dtype is not None:
+            array = self.xp.array(data, dtype=dtype)
```
danieldk, line 97: Do we need this as a separate case? numpy/cupy.array have `None` as the default value of `dtype`. So, we could use the `else` branch and pass `dtype` there.

adrianeboyd: Maybe not? I was mainly trying to revert it to something closer to the version before #656.

PR #903, opened 2023-09-11
Update develop from master following v8.2.0

adrianeboyd:
## Description

This is in preparation for moving v8.2 development to `master`.

### Types of change

Chore.

## Checklist


adrianeboyd:
It looks like the CI is not going run because there are no changes in the code. This is just to sync the branches so v8.1 moves to `v8.1.x` (already created) and v8.2 moves to `master`.

PR #904, opened 2023-09-11
Redesign cython profile defaults due to python 3.12

adrianeboyd:
## Description

Profiling support for python 3.12 will not be available for cython 0.29, so toggle internal defaults to be able to disable profiling for python 3.12 completely in `setup.py`. The cython `profile` compiler directive in `setup.py` is then overridden by file-specific or function-specific settings.

* Swap file-specific `profile` settings to `False`
* In setup, set `profile` default to:
  * `True` for python < 3.12
  * `False` for python >= 3.12

### Types of change

?

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ MOD_NAMES = [
     "thinc.backends.numpy_ops",
     "thinc.extra.search",
     "thinc.layers.sparselinear",
-    "thinc.layers.premap_ids"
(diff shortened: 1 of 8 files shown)


adrianeboyd:
We can't test this (easily) on python 3.12 until a final version of numpy 1.26.0 is released. I've tested this locally with numpy 1.26.0rc1.

PR #910, opened 2023-11-06
CI: Use stable python 3.12

adrianeboyd:
## Description

Use stable python 3.12 and require torch v2.1.1+ for macos extras tests.

### Types of change

Chore.

## Checklist


adrianeboyd:
This seems like the best option to get the CI working again and hopefully in a state where it will continue to work in the future without modification.

Currently no extra torch tests are run on macos.

PR #913, opened 2023-12-12
Add ParametricAttention.v2

danieldk:
## Description

This layer is an extension of the existing `ParametricAttention` layer, adding support for transformations (such as a non-linear layer) of the key representation. This brings the model closer to the paper that suggested it (Yang et al, 2016) and gave slightly better results in experiments.

### Types of change

Feature

## Checklist

### DIFF
--- a/thinc/layers/parametricattention_v2.py
+++ b/thinc/layers/parametricattention_v2.py
@@ -1,0 +1,107 @@
+from typing import Callable, Optional, Tuple, cast
+
+from ..config import registry
+from ..model import Model
+from ..types import Floats2d, Ragged
+from ..util import get_width
+
+InT = Ragged
+OutT = Ragged
+
+
+@registry.layers("ParametricAttention.v2")
+def ParametricAttention_v2(
+    *,
+    key_transform: Optional[Model[Floats2d, Floats2d]] = None,
+    nO: Optional[int] = None
+) -> Model[InT, OutT]:
+    if key_transform is None:
+        layers = []
+        refs = {}
+    else:
+        layers = [key_transform]
+        refs = {"key_transform": cast(Optional[Model], key_transform)}
+
+    layers = [key_transform] if key_transform is not None else []
+
+    """Weight inputs by similarity to a learned vector"""
+    return Model(
+        "para-attn",
+        forward,
+        init=init,
+        params={"Q": None},
+        dims={"nO": nO},
+        layers=layers,
+        refs=refs,
+    )
+
+
+def forward(model: Model[InT, OutT], Xr: InT, is_train: bool) -> Tuple[OutT, Callable]:
+    Q = model.get_param("Q")
+    key_transform = model.maybe_get_ref("key_transform")
+
+    attention, bp_attention = _get_attention(
+        model.ops, Q, key_transform, Xr.dataXd, Xr.lengths, is_train
+    )
+    output, bp_output = _apply_attention(model.ops, attention, Xr.dataXd, Xr.lengths)
+
+    def backprop(dYr: OutT) -> InT:
+        dX, d_attention = bp_output(dYr.dataXd)
+        dQ, dX2 = bp_attention(d_attention)
+        model.inc_grad("Q", dQ.ravel())
+        dX += dX2
+        return Ragged(dX, dYr.lengths)
+
+    return Ragged(output, Xr.lengths), backprop
+
+
+def init(
+    model: Model[InT, OutT], X: Optional[InT] = None, Y: Optional[OutT] = None
+) -> None:
+    key_transform = model.maybe_get_ref("key_transform")
+    width = get_width(X) if X is not None else None
+    if width:
+        model.set_dim("nO", width)
+        if key_transform is not None:
+            key_transform.set_dim("nO", width)
+
+    # Randomly initialize the parameter, as though it were an embedding.
+    Q = model.ops.alloc1f(model.get_dim("nO"))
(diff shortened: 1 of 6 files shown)


thinc/tests/layers/test_layers_api.py
```diff
@@ -8,6 +8,7 @@
 from thinc.api import Dropout, Model, NumpyOps, registry, with_padded
 from thinc.backends import NumpyOps
 from thinc.compat import has_torch
+from thinc.layers.relu import Relu
```
svlandeg, line 11: Why is this import needed?

thinc/tests/layers/test_layers_api.py
```diff
@@ -129,6 +130,8 @@ def assert_data_match(Y, out_data):
     ("MultiSoftmax.v1", {"nOs": (1, 3)}, array2d, array2d),
     # ("CauchySimilarity.v1", {}, (array2d, array2d), array1d),
     ("ParametricAttention.v1", {}, ragged, ragged),
+    ("ParametricAttention.v2", {}, ragged, ragged),
+    ("ParametricAttention.v2", {"key_transform": {"@layers": "Gelu.v1"}}, ragged, ragged),
```
svlandeg, line 134: We should also have a test to check that the `key_transform` layer is actually inserted in the model, as opposed to when it's `None`.

thinc/layers/parametricattention_v2.py
```diff
@@ -0,0 +1,107 @@
...
+        params={"Q": None},
+        dims={"nO": nO},
+        layers=layers,
+        refs=refs,
+    )
+
+
+def forward(model: Model[InT, OutT], Xr: InT, is_train: bool) -> Tuple[OutT, Callable]:
+    Q = model.get_param("Q")
+    key_transform = model.maybe_get_ref("key_transform")
```
svlandeg, line 41: With respect to the unit tests. I'm thinking that if we would somehow misspell this line, something like `model.maybe_get_ref("key_transformmm")` then our unit tests wouldn't catch this, nor would users - everything would still run as is as `key_transform` is allowed to be `None`. So let's think of ways we can test this more robustly.

thinc/layers/parametricattention_v2.py
```diff
@@ -0,0 +1,107 @@
...
+        dX += dX2
+        return Ragged(dX, dYr.lengths)
+
+    return Ragged(output, Xr.lengths), backprop
+
+
+def init(
+    model: Model[InT, OutT], X: Optional[InT] = None, Y: Optional[OutT] = None
+) -> None:
+    key_transform = model.maybe_get_ref("key_transform")
```
svlandeg, line 61: The other thing we should probably do, is to store this `"key_transform"` string in a variable at the top of this file.

svlandeg (commented):
Looks good! I only had a few concerns about robustness & tests.

thinc/layers/parametricattention_v2.py
```diff
@@ -0,0 +1,107 @@
...
+    nO: Optional[int] = None
+) -> Model[InT, OutT]:
+    if key_transform is None:
+        layers = []
+        refs = {}
+    else:
+        layers = [key_transform]
+        refs = {"key_transform": cast(Optional[Model], key_transform)}
+
+    layers = [key_transform] if key_transform is not None else []
```
danieldk, line 25: I was thinking, all this custom handling could be avoided if we had an identity layer. If `key_transform` is `None`, we could just use the identity layer. Then we don't need any conditionals after that (also simplifies the testing al lot).

svlandeg: Good point! Isn't the `noop` basically an identity layer?

danieldk: It is, thanks :), I somehow looked over it because the standard terminology is identity.

danieldk: I'll update the PR after my current training run is done.

thinc/tests/layers/test_layers_api.py
danieldk: Removed.

thinc/tests/layers/test_layers_api.py
danieldk: Added a test that checks if the transform is added to the model.

thinc/layers/parametricattention_v2.py
danieldk: This shouldn't occur anymore. Since the `noop` layer is now used, all the `maybe_get_ref`s are replaced by `get_ref`s.

thinc/layers/parametricattention_v2.py
danieldk: Done.

website/docs/api-layers.md
```diff
@@ -686,6 +686,44 @@ attention mechanism.
...
+<inline-list>
+
+- **Input:** <ndarray>Ragged</ndarray>
+- **Output:** <ndarray>Ragged</ndarray>
+- **Parameters:** <ndarray shape="nO,">Q</ndarray>
+
+</inline-list>
+
+A layer that uses the parametric attention scheme described by
+[Yang et al. (2016)](https://www.cs.cmu.edu/~./hovy/papers/16HLT-hierarchical-attention-networks.pdf).
```
adrianeboyd, line 700:
```suggestion
[Yang et al. (2016)](https://aclanthology.org/N16-1174/).
```

svlandeg (approved):
Definitely much cleaner implementation with the `noop` layer. Looks good to merge!

PR #916, opened 2023-12-21
Temporarily revert new loss implementations

danieldk:
## Description

Temporarily revert new loss implementations for the more limited scope v9.

### Types of change

Maintenance

## Checklist


svlandeg (commented):
This is difficult to review, but nothing jumps out at me. As internally discussed, let's keep this functionality for v10 so we don't lose this work.

danieldk:
The new losses are now in https://github.com/explosion/thinc/tree/v10 , getting ready for the subsequent major release.

PR #917, opened 2024-01-09
Merge master into v9

danieldk:
## Description

**Do not squash merge**

### Types of change

Maintenance

## Checklist


thinc/backends/__init__.py
```diff
@@ -149,22 +155,14 @@ def contextvars_eq_thread_ops() -> bool:
...
-
-
-def _create_thread_local(
-    attrs: Dict[str, Any], local_class: Type[threading.local] = threading.local
-):
-    obj = local_class()
-    for name, value in attrs.items():
-        setattr(obj, name, value)
-    return obj
+    if not hasattr(_LOCAL_STATE, "initialized") or not _LOCAL_STATE.initialized:
```
danieldk, line 161: The other change that needs more careful review is here. This code didn't type check anymore, and it's hard to get to type check without ignores because we can't nilly-willy assign instance variables to classes. Also, it is not really necessary, since (paradoxically) we can use a global variable for thread-local storage, the storage will differ per thread.

thinc/layers/strings2arrays.py
```diff
@@ -17,8 +18,10 @@ def strings2arrays() -> Model[InT, OutT]:
 
 
 def forward(model: Model[InT, OutT], Xs: InT, is_train: bool) -> Tuple[OutT, Callable]:
-    hashes = [[hash_unicode(word) for word in X] for X in Xs]
-    hash_arrays = [model.ops.asarray2i(h, dtype="uint64") for h in hashes]
+    # Cast 32-bit (signed) integer to 64-bit unsigned, since such casting
```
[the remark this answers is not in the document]
svlandeg: Ah never mind, I think I've got it.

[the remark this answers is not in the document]
svlandeg: Change looks good to me. Backport for `master` is here: https://github.com/explosion/thinc/pull/918

thinc/backends/__init__.py
svlandeg: Looks good to me, though I'm not that across these details.

svlandeg (approved):

.github/workflows/autoblack.yml
```diff
@@ -1,44 +0,0 @@
-# GitHub Action that uses Black to reformat all Python code and submits a PR
```
svlandeg, line 1: Don't we want to keep this?

PR #936, opened 2024-05-04
Switch on `importlib.resources`

shadchin:
## Description

What's wrong with`__file__`? The assumption that you have files and subdirectories available is not correct. This approach doesn't work if executing code which is packed in a zip or a wheel, and it may be entirely out of the user's control whether or not your package gets extracted to a filesystem at all.

### Types of change

Switch on `importlib.resources.files` (Added in version Python 3.9.)

## Checklist

### DIFF
--- a/thinc/backends/_custom_kernels.py
+++ b/thinc/backends/_custom_kernels.py
@@ -2,14 +2,14 @@ import operator
 import re
 from collections import defaultdict
 from functools import reduce
-from pathlib import Path
+from importlib.resources import files
 from typing import Callable, Optional, Tuple
 
 import numpy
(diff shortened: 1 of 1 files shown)


honnibal:
As a general rule we're not claiming to support zip-safe operation, and there may be other places in the code where we're not zip safe. So I'm not inclined to accept a piecemeal change; if we want zip safety it should all be done at once.

This isn't something I'd ask a contributor to do as I haven't thought through the trade-offs properly, and I may end up having to reject the PR. Nevertheless, thanks for your attention and the submission of what seemed like a small fix!

PR #944, opened 2024-07-03
replace old "mock" with newer "unittest.mock" from the standard library

a-detiste:
Hi, 

After it's merging inside the standard library as `unittest.mock`,
the old external `mock`library is left to slowly bitrot.

---

mock is now part of the Python standard library, available as [unittest.mock](https://docs.python.org/dev/library/unittest.mock.html) in Python 3.3 onwards.

https://github.com/testing-cabal/mock

## Checklist

- [x ] I confirm that I have the right to submit this contribution under the project's MIT license.


davispuh:
This is needed for Python 3.13 support.

PR #954, opened 2025-04-03
Disable type validation in order to update to Pydantic2

honnibal:
We need to update to Pydantic2 in order to update to Python3.13. The type validation approach we use here doesn't work in Pydantic 2, and I haven't yet figured out what we should do instead. As it's an optional feature, I've disabled this for now to unblock the release.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,9 +51,8 @@ install_requires =
     confection>=0.0.1,<1.0.0
     # Third-party dependencies
     setuptools
-    numpy>=1.19.0,<3.0.0; python_version < "3.9"
-    numpy>=1.19.0,<3.0.0; python_version >= "3.9"
-    pydantic>=1.7.4,!=1.8,!=1.8.1,<3.0.0
+    numpy>=2.0.0,<3.0.0
+    pydantic>=2.0.0,<3.0.0
     packaging>=20.0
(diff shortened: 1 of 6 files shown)


setup.cfg
```diff
@@ -51,9 +51,8 @@ install_requires =
     confection>=0.0.1,<1.0.0
     # Third-party dependencies
     setuptools
-    numpy>=1.19.0,<3.0.0; python_version < "3.9"
-    numpy>=1.19.0,<3.0.0; python_version >= "3.9"
-    pydantic>=1.7.4,!=1.8,!=1.8.1,<3.0.0
+    numpy>=2.0.0,<3.0.0
```
jamesobutler, line 54: @honnibal Why was numpy>=2.0.0 updated here? This code change ultimately was included in https://github.com/explosion/thinc/releases/tag/release-v8.3.6 however in https://github.com/explosion/thinc/releases/tag/release-v8.3.3 the point was made that the package runtime could be indeed compatible with Numpy 1 and 2 (while build process requires Numpy 2).

> Widen numpy pin, allowing versions across v1 and v2. Previously I had thought that if I build against numpy v2, I couldn't also have v1 as a runtime dependency. This is actually incorrect, so we can widen the numpy pin

Issue #966, opened 2026-02-10
Provide support for out of tree Pytorch backends


tehbone:
This is somewhat related to some of the discussion in PR #559. The prevailing guidance from Pytorch for new devices is to take advantage of the `privateuseone` device. The conventions established by that approach allow for an out of tree device to be defined and for a unique name to be registered, replacing the `privateuseone` name while in use.

I looked around in `thinc` to see if such support could be added similarly to how `thinc_mps_ops` and `thinc_bigendian_ops` were added, but as noted in the previous PR, there is some added complexity since currently `thinc` has to be aware of the backend and vise-versa, leading to potential cyclic import loops.

Could a backend interface be established such that one could register either an out of tree cpu or "gpu" backend, and have thinc pick it up? My first thought would be that this backend could register an entrypoint (`thinc.backends` maybe?) and then provide all the information necessary for not only the current Ops but whatever pytorch behavior is necessary.

I did start playing with this conceptually with https://github.com/tehbone/thinc_nnpa_ops - which will actually need a mix of the CPU and GPU functionality.

honnibal:
@tehbone I'm happy to accept guidance on what this should look like. Ultimately I think our ecosystem needs to get out of the way more.

tehbone:
I have a patch that I'm testing that I'll gladly open a PR for once I'm successful. Should be within a few days.

PR #968, opened 2026-03-27
Improves support for out of tree Ops.

tehbone:
Utilizes the entry_point registration to search for Ops that have been defined out of tree. Additionally refactors a bit of the GPU configuration and selection logic to not assume CUDA devices but also support the possibility that the device is non-CUDA.

## Description

Pytorch is convincing vendors to add support for their devices out of their tree, by way of the PrivateUseOne device. This has a side effect that properly configured devices using that strategy will not present themselves as PrivateUseOne, but by one of the vendor's naming convention. In our case, the device is the `nnpa` device. The support provided here would allow one to extend the Ops interface and have the ability to choose one of these devices.

The main usage update is follows: `require_gpu` and `prefer_gpu` now take a keyword parameter "name", which can refer to any of the registered ops, and the functions will behave as expected - the prefer form will use the ops if it is found and the newly-added `Ops.has_gpu_support` is `True`. The require form will error if it cannot be used. The default form where the name parameter is not specified will search the entrypoints for out of tree definitions before falling back on the old method of autodetection.

A minimal example of an out of tree op that has been tested with this is [here](https://github.com/tehbone/thinc-nnpa-ops): (note - it only works on s390x machines) 

### Types of change

Enhancement/new feature, addresses #966.

## Checklist

- [ ] My changes don't require a change to the documentation, or if they do, I've added all required information.

### DIFF
--- a/thinc/backends/cupy_ops.py
+++ b/thinc/backends/cupy_ops.py
@@ -1,7 +1,8 @@
 import numpy
+from typing import Any
 
 from .. import registry
-from ..compat import cublas, cupy, cupyx
+from ..compat import cublas, cupy, cupyx, has_cupy_gpu, has_torch_cuda_gpu, torch
 from ..types import DeviceTypes
 from ..util import (
     is_cupy_array,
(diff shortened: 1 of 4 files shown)


tehbone:
This is a stab at addressing #966. Please let me know if this approach would work.

tehbone:
Bump?

cc: @honnibal

PR #970, opened 2026-04-21
Fix loading little-endian models on big-endian hosts 

satyamg1620:
## Description
Loading spaCy pipelines (e.g. en_core_web_sm) on big-endian hosts like s390x currently fails with ValueError: Little-endian buffer not supported on big-endian compiler, because thinc's Cython typed memoryviews reject the little-endian arrays that ship in the serialized model. This PR adds a thinc.util.ensure_native_byteorder helper and routes both Ops.asarray and NumpyOps.asarray through it, so deserialized arrays are byteswapped to native order before reaching any Cython path. On little-endian platforms native arrays are returned by identity, so there is no behavioural or performance change.

### Types of change
Bug fix — enables thinc (and downstream spaCy) to load little-endian model artifacts on big-endian CPUs (like s390x ). 

## Checklist

- [X] I confirm that I have the right to submit this contribution under the project's MIT license.
- [X] I ran the tests, and all new and existing tests passed.
- [X] My changes don't require a change to the documentation, or if they do, I've added all required information.

### DIFF
--- a/thinc/backends/numpy_ops.pyx
+++ b/thinc/backends/numpy_ops.pyx
@@ -18,7 +18,7 @@ from preshed.maps cimport PreshMap
 
 from .. import registry
 from ..types import ArrayXd, DeviceTypes, DTypes, Shape
-from ..util import copy_array, get_array_module
+from ..util import copy_array, ensure_native_byteorder, get_array_module
 
 from .cblas cimport CBlas, daxpy, saxpy
 from .linalg cimport Vec, VecVec
@@ -76,7 +76,7 @@ class NumpyOps(Ops):
         if dtype is not None:
             array = array.astype(dtype=dtype, copy=False)
 
-        return array
+        return ensure_native_byteorder(array)
 
 
     def alloc(self, shape: Shape, *, dtype: Optional[DTypes] = "float32", zeros: bool = True) -> ArrayXd:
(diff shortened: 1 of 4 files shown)


satyamg1620:
@honnibal @ines Can you please review this PR

modassarrana89-new (approved):
LGTM

tehbone:
Shouldn't this also drop the lazy load of BigEndianOps in compat.py?

satyamg1620:
> Shouldn't this also drop the lazy load of BigEndianOps in compat.py?

I think the lazy load isn't in compat.py, it's in thinc/backends/__init__.py inside _import_extra_cpu_backends(), mirroring the AppleOps pattern.

tehbone:
> > Shouldn't this also drop the lazy load of BigEndianOps in compat.py?
> 
> I think the lazy load isn't in compat.py, it's in thinc/backends/**init**.py inside _import_extra_cpu_backends(), mirroring the AppleOps pattern.

Yes, my mistake. I was thinking of the wrong file. But it would be good to get rid of that if it is now just supported within the base numpy ops

ckhordiasma:
hi @tehbone, just to confirm - did you want @satyamg1620 to remove that before proceeding with merging this PR?

satyamg1620:
@honnibal @ines Can you please review this PR

